PR #118, opened 2026-04-05
feat: add Observio sample agent for agent health evaluation

goyamegh:
## Summary
- Add **Observio**, a reference ReAct agent (based on [osd-agents](https://github.com/opensearch-project/OpenSearch-Dashboards/tree/main/packages/osd-agents)) as a practice target for agent health improvements
- Rewrite README to frame it as a sample agent with clear "What to Improve" guidance
- Update agent health docs (README, GETTING_STARTED, CONNECTORS, CHANGELOG) to reference Observio
- Exclude `observio-sample-agent/` from parent tsconfig (it has its own build)

## Test plan
- [ ] Verify `observio-sample-agent/` builds independently with `cd observio-sample-agent && npm install && npm run build`
- [ ] Verify parent project builds: `npm run build:all`
- [ ] Verify all tests pass: `npm run test:all`
- [ ] Verify Observio starts: `cd observio-sample-agent && npm run start:ag-ui`
- [ ] Verify Agent Health can evaluate Observio via `agui-streaming` connector

### DIFF
--- a/tests/integration/server/routes/config.integration.test.ts
+++ b/tests/integration/server/routes/config.integration.test.ts
@@ -25,7 +25,7 @@ const TEST_TIMEOUT = 30000;
  * Number of agents in DEFAULT_CONFIG (lib/constants.ts).
  * If the defaults change, update this constant.
  */
-const DEFAULT_AGENT_COUNT = 4;
+const DEFAULT_AGENT_COUNT = 2;
 
 /**
  * Minimum number of models in DEFAULT_CONFIG.


goyamegh:
**Re: Comment #8 (E2E `.catch(() => null)` in comparison tests)**

The `.catch(() => null)` pattern is intentional — these E2E tests guard against cases where demo data may not have the expected buttons (e.g., no runs to compare). The surrounding `if` guards ensure the test only asserts when the prerequisite UI elements are present. Without this, the test would fail in CI environments where sample data isn't guaranteed.

**All other Copilot comments addressed in latest commit** (`76a5f8d`):
- ✅ Route `/` → `/run-agent`
- ✅ `streaming: false` → `streaming: true`
- ✅ `default-model.json` → `default_model.json`
- ✅ `join(filePath, '..')` → `dirname(filePath)`
- ✅ Removed hooks from `settings.json`
- ✅ Deleted `jest.config.js`
- ✅ Fixed port 3000 → 3001 in docs

PR #219, opened 2026-05-21
fix(ui): theming paper cuts in dark/light mode (Scope A)

goyamegh:
## Summary

Resolves the top 5 highest-impact CSS/theming paper cuts found in a UX design review of light + dark themes.

### Fixed
1. **RawEventsPanel** — replaced hardcoded `bg-gray-*`/`text-gray-*` with theme tokens (`bg-card`, `bg-muted`, `text-muted-foreground`, etc.) so dark mode no longer renders washed-out grays on a near-black background.
2. **LatencyHistogram & MetricsOverview** — added `dark:` color variants for the bucket bars and overview tiles so they remain legible in dark mode.
3. **MetricsTimeSeriesChart** — Recharts axes/grid now read from CSS variables (`hsl(var(--border))`, `hsl(var(--muted-foreground))`) instead of hex constants, picking up theme changes live.
4. **Trace flow views** — `TraceFlowView`, `AgentMapView`, `TraceFlowComparison`, and `SpanNode` switched off `slate-*` `!important` overrides in favor of tokenised classes.
5. **Tooltip** — shadcn `TooltipContent` now uses `bg-popover` / `text-popover-foreground` so the background and text contrast flip correctly between themes; removed hardcoded gray TooltipContent overrides in `AgentTracesPage` and `TraceFlyoutContent`.

### Tests
- Added `tests/unit/components/uxThemingRegression.test.ts` — 16 source-level guards that fail if any of the above regressions are reintroduced (verified via red-green by sabotaging `tooltip.tsx`).
- Repaired 3 pre-existing test failures that were blocking the pre-push gate (one TS syntax error from a bad merge, one stale assertion after the evaluation-spans addition, one Pi connector test using outdated flag names).

## Test plan

- [x] `npm run test:unit` — 3544 passed
- [x] Manual screenshot diff in light + dark themes (captured under `~/ux-review-shots/<feature>/{before,after}-{light,dark}.png`)
- [ ] Reviewer spot-check in dark mode: traces table, agent map, metrics overview, latency histogram, and any tooltip


goyamegh:
**Coverage fixed** (commit 1516cd9): `codecov/patch` and `codecov/project` both flipped red \u2192 green.
- Root cause: the Scope A theming diff's only regression coverage was source-string regexes (`uxThemingRegression.test.ts`), which earn zero code coverage \u2014 patch sat at 29.41% (target 59.25%), and project ticked down to 59.24% (-0.01%).
- Fix: added real render tests for `RawEventsPanel`, `TrajectoryView`, `codingAgents/CodingAgentsPage` (exported `SessionDetailPanel` for direct testing), and an extracted `getLogLevelColor()` pure function for `RunDetailsContent` (its JSX call site is unreachable \u2014 `isTraceMode` is hardcoded `true` \u2014 so it's unit-tested directly). 4 new/updated test files, 8 new test cases.
- Patch: 29.41% \u2192 all modified/coverable lines covered (codecov: "All modified and coverable lines are covered by tests."). Project: 59.24% \u2192 70.71% in the unit-flag-only snapshot (final blended number settles once the e2e flag uploads).
- Verified locally: `npm run build:all`, `npm run test:unit` (260 suites / 4647 passed), `npm run test:integration` (71 suites / 750 passed), `tsc --noEmit` clean.

TackAdam (approved):

PR #234, opened 2026-06-03
fix(sdk): traces fixture now reflects real OTel data (#230)

### LINKED ISSUE #230: [SDK] traces fixture silently returns zeros — trace-based assertions falsely pass
## Summary

The code-based SDK `traces` fixture silently returns zeros in `0.5.x`. Trace-based assertions therefore appear to pass while actually measuring nothing — a silent footgun for anyone who has instrumented OTel traces and expects the fixture to reflect them.

## Details

In the shipped build, `buildFixtures()` always returns an empty traces accessor regardless of the agent's `useTraces` setting:

```js
function buildFixtures(result) {
  return {
    result,
    judge,
    traces: emptyTracesAccessor(),  // <-- always empty
    expect,
  };
}
```

`emptyTracesAccessor()` returns `0`/`[]` for every accessor (`totalTokens`, `totalCost`, `spanDuration()`, `toolCalls`, `spans`).

This is consistent with the documented roadmap item in `docs/SDK.md`:

> - [ ] Real traces pre-loading from OTel exporter (currently always empty)

…but the docs example actively encourages writing assertions against the fixture:

```js
expect(traces.totalTokens).to.be.lessThan(10_000);
expect(traces.totalCost).to.be.lessThan(0.05);
expect(traces.spanDuration('search_logs')).to.be.lessThan(2_000);
```

## Why this is a problem

Because every accessor returns `0`, assertions like:

```js
expect(traces.totalTokens).to.be.lessThan(10_000);   // passes against 0
expect(traces.totalCost).to.be.lessThan(0.05);        // passes against 0
```

**pass silently** even when no trace data exists. A user who has instrumented OTel traces (and sees them in the Traces tab) reasonably expects the fixture to reflect them, but instead gets green matchers that measure nothing. The failure mode is invisible — there's no error, warning, or skipped status.

## Repro

1. Configure an agent with `useTraces: true` and emit OTel traces (verifiable in the Traces tab).
2. Write a `.eval.ts` test asserting on `traces.totalTokens` / `traces.totalCost`.
3. Run via `agent-health benchmark -f ... -a <agent>`.
4. Observe the trace matchers pass regardless of actual token/cost, because the fixture is empty.

## Suggested fix

Either:

- **(a)** Prioritize the roadmap item — pre-load real OTel data into the `traces` fixture (e.g., via the existing `fetchTracesByRunIds`) when `agentConfig.useTraces` is set; **or**
- **(b)** Until (a) lands, make the fixture **fail loudly** instead of returning silent zeros — e.g., throw or emit a warning when a `traces.*` accessor is read while `useTraces: true` but no data was pre-loaded. That turns a silent false-pass into an actionable signal.

Even just gating the documented examples behind a "not yet wired" note would help, but loud-failure (option b) is the safer interim behavior.

## Environment

- `@opensearch-project/agent-health` 0.5.x (observed in a 0.5.8 build)
- Code-based SDK (`test()` / `expect()` / `traces` fixture)

goyamegh:
## Summary

Fixes #230. The code-based SDK `traces` fixture was hard-coded to `emptyTracesAccessor()` regardless of the agent's `useTraces` setting, so every accessor returned `0` / `[]`. Trace-based assertions like `expect(traces.totalTokens).to.be.lessThan(10_000)` therefore passed silently against `0` even when the agent actually used 12 000 tokens — a green matcher that measured nothing.

## What changed

The runner now constructs the fixture three ways:

| Mode | Behaviour |
|---|---|
| `useTraces: false` | **Silent zeros** — opt-out preserved, existing matchers unchanged |
| `useTraces: true` + spans available | Real `buildTracesAccessor(spans)` — token / cost / span-duration aggregates reflect actual OTel data |
| `useTraces: true` + no spans / no `runId` / fetch timeout | **Loud-failure accessor** — every getter throws `traces fixture unavailable: <reason>` so the silent false-pass becomes an actionable error |

A new lightweight `fetchSpansForRun` helper polls the existing `fetchTracesByRunIds` endpoint with a configurable retry budget (driven by the agent's existing `tracePolling` block).

## Files

- **`lib/matchers/traces.ts`** — adds `unavailableTracesAccessor(reason)` alongside the existing `empty` and `build` factories.
- **`lib/matchers/index.ts`** — re-exports the new factory.
- **`services/traces/fetchSpansForRun.ts`** *(new)* — thin polling helper that returns `Span[]` (or `[]` on timeout / error).
- **`services/evaluationRunner.ts` & `services/benchmarkRunner.ts`** — both `buildFixtures()` callsites now take a pre-loaded accessor; new `loadTracesAccessor()` helper picks the appropriate factory based on `agentConfig.useTraces` and span availability.
- **`docs/SDK.md`** — "Traces fixture" section rewritten to document the three modes; roadmap item ticked.
- **`CHANGELOG.md`** — `### Fixed` entry under `[Unreleased]`.

## Tests

- `tests/unit/lib/matchers/traces.test.ts` *(new, 15 tests)* — covers all three accessors, alias attribute keys (`gen_ai.usage.*`, `llm.usage.*`, `input_tokens`/`output_tokens`), string-numeric attrs, ms vs nanosecond span timestamps, and a dedicated `regression #230` case asserting `lessThan(N)` against an unavailable fixture is no longer green.
- `tests/unit/services/traces/fetchSpansForRun.test.ts` *(new, 7 tests)* — first-attempt success, retry until spans appear, `maxAttempts` exhaustion (no throw), transient/persistent error swallowing, configured interval, `maxAttempts: 1` no-retry path.
- `tests/unit/services/evaluationRunner.traces230.test.ts` *(new, 5 tests)* — runner-level cases A–E from the verification plan:
  - **A**: `useTraces: false` → silent zeros (opt-out preserved)
  - **B**: `useTraces: true`, no spans → reads on `traces.*` throw, test fails
  - **C** *(regression)*: `useTraces: true`, 12 k tokens → `lessThan(10_000)` **fails** (pre-fix this silently passed against 0)
  - **D**: `useTraces: true`, 5 k tokens → `lessThan(10_000)` passes using real numbers
  - **E**: `useTraces: true` + missing `runId` → throws with the no-runId reason

## Verification

- `npx tsc --noEmit -p tsconfig.json` — **clean**
- 27 new tests pass; existing 4441 tests in the affected areas continue to pass; **0 new failures** introduced.
- Reverted the four production-file changes locally and re-ran the new runner regression tests: **4 of 5 fail without the fix** (Case A passes because the opt-out path always worked; B–E catch the bug). Re-applied the fix: all 27 pass.

Verification matrix and step-by-step plan: see the issue thread.

---

**Issue:** Closes #230

**Signed-off-by:** Megha Goyal <goyamegh@amazon.com>


jasonlhamazon (approved):

goyamegh:
Pushed `a3971677` to address the open review feedback.

## Copilot review comments

| # | Comment | Status |
|---|---|---|
| #1 | `trace` unused import in `services/benchmarkRunner.ts` (was on stale `cb84ddc`) | ✅ Not present on current head — the unrelated OTel context propagation work isn't in this PR |
| #2 / #3 | Poll defaults `10 × 1s` diverge from documented `30 × 10000ms` | ✅ Fixed in `a3971677` — `AgentConfig.tracePolling` docstring now accurately documents the two distinct paths (judge poller `60 × 10000ms`, SDK pre-load `10 × 1000ms`), the shared env-var overrides (`TRACE_POLL_INTERVAL_MS` / `TRACE_POLL_MAX_ATTEMPTS`), and the hard 60-attempt ceiling that both paths enforce |
| #4 | `fetchSpansForRun` swallows fetch errors | ✅ Already addressed in `3698032f` — helper now returns `{spans, lastError}` and `loadTracesAccessor` constructs a specific `"fetch failed for runId=…: <error>"` reason instead of the generic `"no spans found"` |
| #5 | OTel `test_case` span before agent execution leaked in (was on stale `cb84ddc`) | ✅ Not present on current head |

## CI status

- ✅ **CI** (build/test) — passing
- ✅ **Changelog Verification** — passing
- ❌ **Links Checker** — same pre-existing failure as `origin/main`. Not introduced by this PR. Every link in the 3 #230 CHANGELOG entries this PR adds points at a file that exists in `origin/main`. The broken links Lychee flags (`components/ComparisonPage.tsx`, `services/failureClusterService.ts`, `.claude/skills/agent-health/SKILL.md`, `server/routes/comparison.ts`) are inherited from `origin/main`'s own [Unreleased] section — the same workflow has been failing on the last two `main` pushes (#228 follow-up #237 merged through it). `a3971677` rebuilt the PR's `CHANGELOG.md` as `origin/main`'s CHANGELOG + only the 3 #230 entries, so the PR is no longer ADDING any broken links of its own.
- ❌ **PR Review (Copilot bot)** — bot signals "failure" status when it leaves review comments. Not blocking; the bot re-runs on each push.

## Pending

Nothing blocking from my side. The Links Checker noise is pre-existing on `origin/main` and not introduced here. Ready for re-review whenever convenient.

TackAdam (approved):

PR #241, opened 2026-06-04
feat(sdk): add beforeEach/afterEach lifecycle hooks for per-test provisioning

### LINKED ISSUE #229: [SDK] Add beforeEach/afterEach lifecycle hooks for per-test provisioning
## Summary

Add Playwright-style **`beforeEach` / `afterEach` lifecycle hooks** to the code SDK for per-test setup and teardown.

## Why

Per-test context that fits in the request is **already handled today** — declare it on the test (`context: [{ description, value }]`) and forward it in the connector's `buildPayload`. That path is request-scoped and parallel-safe, no change needed.

What's missing is the case a connector *can't* cover: **side-effecting provisioning with a lifecycle**. A connector is a pure request-shaper with no teardown, so it can't, per test:

- materialize an out-of-band resource the agent reads (e.g. a repo/workspace checkout on disk, a seeded DB, a started sandbox/mock service, a temp file/index), and
- tear it down afterward.

That needs a hook that runs before/after each test, isolated per test.

## Proposed shape

```js
test.beforeEach(async ({ testInfo }) => {
  // provision an out-of-band resource for THIS test (checkout, seed DB, start mock…)
});

test.afterEach(async ({ testInfo }) => {
  // clean it up
});
```

Standard semantics: per-file (and/or per-describe) registration, run in isolation around each test, with `beforeAll`/`afterAll` as the suite-scoped variants. A way to surface what was provisioned to the test/connector (e.g. a path or handle) would round it out.

## Questions
1. Do hooks fit the SDK's direction, and is this on the roadmap already?
2. Preference on surfacing provisioned values to the test body / connector (return value, `testInfo`, a `provide()`-style helper)?

goyamegh:
Closes #229.

## Why

The code SDK already covers per-test data that fits in the request payload (the `context` field forwarded by the connector's `buildPayload`). What it can't express is **side-effecting, per-test setup with a teardown lifecycle** — connectors are pure request-shapers with no `afterEach`, so they can't:

- materialize an out-of-band resource the agent reads (a workspace checkout on disk, a seeded DB, a started sandbox/mock service, a temp file/index), and
- tear it down once the test finishes.

This PR adds Playwright-style hooks for that. Plan was discussed and agreed in the issue: https://github.com/opensearch-project/agent-health/issues/229#issuecomment-4617343636

## What

New SDK surface:

```js
const { test, beforeAll, afterAll, beforeEach, afterEach, expect } = require('@opensearch-project/agent-health');

let suiteRoot;
beforeAll(() => { suiteRoot = fs.mkdtempSync('/tmp/my-suite-'); });
afterAll(() => fs.rmSync(suiteRoot, { recursive: true, force: true }));

beforeEach(({ provide, testInfo }) => {
  const dir = fs.mkdtempSync(path.join(suiteRoot, `${testInfo.name}-`));
  provide('workspaceDir', dir);
});
afterEach(({ provisioned }) => {
  fs.rmSync(provisioned.workspaceDir, { recursive: true, force: true });
});

test('uses workspace', { prompt: '...' }, async ({ result, provisioned }) => {
  expect(fs.existsSync(provisioned.workspaceDir)).to.equal(true);
});
```

Also accepted as `test.beforeEach(...)` / `test.afterEach(...)` etc. for Playwright parity (declaration-merged on the existing `test` function — TS users get types, JS users get runtime methods).

### How `provide()` answers Q2 from the issue

- **Parallel-safe.** Each test gets its own fixtures object, so `provide()` writing into `fixtures.provisioned` is automatically test-scoped — no cross-talk between concurrent tests.
- **No magic returns.** Returning a value from `beforeEach` and stitching it into fixtures by name is fragile (typing, ordering, multiple hooks). `provide()` is explicit and composes across multiple `beforeEach` registrations.
- **`testInfo` is read-only metadata** — `{ name, benchmarkPath, sourceFile, testCaseId }`. Useful for naming temp resources or tagging logs.

## Architecture

The novel piece is `services/hookOrchestrator.ts`. Two invariants under the runner's bounded-concurrency dispatch:

1. **`beforeAll` runs exactly once per scope**, even when N tests in the scope start in parallel. Implemented as a promise-based **once-latch**: the first arrival creates the promise; everyone else awaits it.
2. **`afterAll` runs exactly once after the last test in the scope completes** (success or failure). Implemented as a remaining-tests counter, decremented on every test completion.

Scope = `(sourceFile, describePath)`. Scope chain for a test in describe `A > B`: `[file, 'A', 'A > B']`. `beforeAll`/`beforeEach` run outer→inner, `afterEach`/`afterAll` run inner→outer (Playwright/Jest semantics).

Hook errors never crash the runner — they surface as `MatcherResult` entries (with `method: 'code-assertion'`) so the existing per-matcher UI panel renders them next to assertion failures with no UI change. `afterEach`/`afterAll` always run, even when the body or a `beforeEach` threw.

When no hooks are registered, `createHookOrchestrator(...)` returns a `NoopHookOrchestrator` and the hot path stays branchless — existing tests pay zero overhead.

## File changes

| File | Purpose |
|---|---|
| `lib/testCases/types.ts` | New types: `TestInfo`, `HookKind`, `HookFn`, `RegisteredHook`. Extended `TestFixtures` with `testInfo`, `provisioned`, optional `provide`. |
| `lib/testCases/define.ts` | New `hookRegistries` map + four registrars + `getRegisteredHooks`. Namespace-merged `test.beforeEach` etc. |
| `lib/testCases/loader.ts` | `LoadResult.hooks` populated; SDK exports surface the new registrars to CJS fixtures. |
| `lib/testCases/index.ts`, `lib/index.ts` | Public exports. |
| `services/hookOrchestrator.ts` | **New.** Orchestrator + noop variant + factory. |
| `services/sourceResolver.ts` | `ResolvedSources` gains `hooksByFile` + `testHookScopes` maps. |
| `services/{evaluationRunner,benchmarkRunner}.ts` | Build orchestrator once per run; wrap matcher session with `beforeTest`/`afterTest`. |
| `server/routes/storage/{evaluationRuns,benchmarks}.ts` | Forward the new maps to the runner. |

## Tests

3 new test files, ~30 cases:

- `tests/unit/lib/testCases/define.hooks.test.ts` — registration of each hook kind, `test.X` static-method form, describe-path capture, file isolation, `clearRegistry`, experimental warning.
- `tests/unit/lib/testCases/loader.hooks.test.ts` — loader returns hooks alongside test cases (CJS path), `test.X` form, describe path attribution, no-hooks file, reload-doesn't-duplicate.
- `tests/unit/services/hookOrchestrator.test.ts` — once-latch under N=3 concurrency, `beforeAll` failure → all-tests-aborted, drain counter for `afterAll`, `afterAll` runs even when every test fails, ordering (outer→inner before / inner→outer after), `afterEach` runs after body throw and after `beforeEach` throw, hook errors→MatcherResult, `provide()`/`provisioned` isolation across concurrent tests, `testInfo` stamping, factory short-circuit to noop.

Final numbers (locally):
- All 133 tests in the touched areas pass.
- 5 pre-existing `services/sourceResolver.codeImport.test.ts` failures fixed as a side-effect (loader return shape was already mocked but extension was defensive-checked).
- 86 unrelated test failures remain on this branch — all pre-existing on `origin/main` (React `act` API mismatch in `usePersistedSet` etc.). I verified by running the same suite on `origin/main` before changes; failure list is identical.

Build (`npm run build`, `npm run build:lib`, `tsc --noEmit`) all clean.

## Out of scope (intentionally — flagged in the issue plan)

- `expect.soft()` (separate roadmap item).
- Per-hook timeouts beyond the existing `TestOptions.timeout`.
- Worker-level hooks — the runner has no worker abstraction yet.
- A new `MatcherMethod: 'hook'`. Reusing `'code-assertion'` keeps the UI panel unchanged; a follow-up can split it out if the signal proves useful.

## Demo

`evals/sdk-hooks-demo.eval.js` — runnable end-to-end example using a temp directory as the out-of-band resource.

```
AGENT_HEALTH_PORT=4002 npx @opensearch-project/agent-health benchmark \
  -f evals/sdk-hooks-demo.eval.js -a demo
```

## Pre-PR checklist

- [x] All commits have DCO signoff (`git commit -s`)
- [x] `CHANGELOG.md` updated under `## [Unreleased]` with PR link
- [x] `npm run build:all` succeeds
- [x] All hook-related tests pass (`tests/unit/lib/testCases`, `tests/unit/services/hookOrchestrator.test.ts`, `tests/unit/services/sourceResolver`, `tests/unit/services/evaluationRunner`)
- [x] `npm audit --audit-level=high` reports no new vulnerabilities (2 pre-existing, unchanged)
- [x] New source files have SPDX license headers


goyamegh:
## Review-comment round-up

Force-pushed `74ce9b53` after rebasing onto `origin/main` (which now includes #234's traces-fixture work). Both commits squash-clean against the latest `main`. 104/104 tests pass locally (`hookOrchestrator.test.ts`, `define.hooks.test.ts`, `loader.hooks.test.ts`, `loader.test.ts`, `codeSdkIssue245.integration.test.ts`, `benchmarkCodeSdk.integration.test.ts`).

### Addressed in code

| # | Where | Resolution |
|---|---|---|
| 8 | `services/hookOrchestrator.ts` `ScopeState.beforeAllResults` | **Removed.** Was leftover from an earlier design that planned to inspect suite-level results separately; the "replay onto every test" lookup uses `scope.beforeAllRun` directly. |
| 9 | `services/hookOrchestrator.ts` `scopeDescribePaths` | **Removed.** Was a memoisation cache for `chainFor(...)`'s describePath splitting that the chain walk doesn't actually use. |
| 6 | `lib/testCases/types.ts` `provide` JSDoc | **Rewrote** to match runtime: calling `provide` outside `beforeEach` is a runtime error (the orchestrator hands the body fixtures with `provide: undefined`), not a no-op. Pointer added to `fixtures.provisioned[key]` for reads. Aligned with `tests/unit/services/hookOrchestrator.test.ts > 'provide is undefined inside the body (only beforeEach gets it)'`. |
| 7 | `docs/SDK.md` `afterEach` example | **Added a `typeof === 'string'` guard** so the example doesn't throw when `beforeEach` aborted before `provide`. The runnable demo `evals/sdk-hooks-demo.eval.js` already had the same guard; only the docs example was unsafe. |
| 3 | `services/evaluationRunner.ts` comment on `testHookScopes` | **Rewrote.** The "fall through to file-level hooks only" wording was wrong — the orchestrator's chain walk requires a real `sourceFile` on the descriptor, so a missing scope entry means no hooks match at all (the file-level hooks need the file to be set on the descriptor). |

### Discussed but deferred to follow-up — comments 1, 2, 4, 5

These all surface the same root issue: hooks run **after** the connector/agent invocation, inside the deterministic-eval block. So `beforeEach` can't "materialize a workspace the agent reads" — by the time it runs, the agent has already finished. The fixtures factory passes `result: {} as any` because the real `EvalResult` doesn't exist yet at hook-call time, but `beforeTest()` runs before the body sees the overlay, so any hook reading `fixtures.result` would see the placeholder. You're right on both points.

**Why this PR doesn't fix it:**

- Reordering hooks to run **before** the agent invocation is genuinely the right thing for the issue #229 use case ("out-of-band resource the agent reads"). It's also a non-trivial behaviour change: tests that today silently relied on `beforeEach` being a post-agent fixture would now fire pre-agent.
- The fix isn't complete without a connector contract change. Even if hooks ran first, today there's no way to hand `provisioned` values to the connector — no `buildPayload({ provisioned })` parameter, no env-var convention. So `mkdtemp` works in the body but the agent still can't see the dir.
- There's a parallel `await evalFn(arg)` call site in `services/benchmarkRunner.ts` (catch on comment 2) — both runners need the same change. That duplication is what issue #246 (filed during this PR's discussion) is about, and resolving the ordering in one runner without the other would mean the asymmetry hides the regression on the path that has fewer tests.
- The scoped instruction for this PR was to keep the runner functionality identical to what's already running in production; reshaping the hook ordering changes externally-visible behaviour for any test that has a `beforeEach`.

**What's tracked:**

- I'll file a follow-up issue covering the three things together: (a) move hook execution to wrap the agent invocation, not the body; (b) add `provisioned` to the connector contract; (c) split `HookFixtures` from `TestFixtures` so the type doesn't lie about `result` being available before the body runs.
- Issue #246 already covers the runner-duplication side of the picture; the new follow-up will reference it and add the hook-ordering scope on top.

In the meantime, the hook integration as it stands is useful for body-driven assertions (the dominant pattern in the test suite that ships with this PR) and `expectedOutcomes` / `expectedTrajectory` forwarding for #245 is independent of the ordering question. Both work end-to-end against the live system — verified via `cp-oncall-V2215793027` benchmark runs against the Claude Code agent producing 25 matcher entries (4 LLM-judge + 21 chai/custom/hook), all passing.

Ready for re-review.

TackAdam (approved):

Issue #245, opened 2026-06-04
Code-based SDK: test() body is never executed by benchmark — .eval.js tests can't run end-to-end


ashwin-pc:
## Summary

The experimental code-based test SDK — `test(name, options, async ({ result, expect, judge }) => { ... })` — does not run end-to-end under `agent-health benchmark -f <file>.eval.js`. The **body is never executed**: assertions and `judge()` calls in the body have no effect on the result. Combined with the fact that `options.expectedOutcomes` is not forwarded to the server (see #243), a code-based test currently cannot be graded at all — neither by its own `expect()`/`judge()` nor by a server-side evaluator.

## Version

`agent-health` 0.5.9 (also reproduces on 0.5.8).

## Expected behavior

Running `agent-health benchmark -f ./foo.eval.js -a <agent>` should:
1. invoke the agent for each registered `test(...)`, then
2. execute that test's body against the result, so `expect(result.agentOutput)...`, `expect(result.trajectory)...`, and `await judge(result, criteria)` determine (or contribute to) the case pass/fail.

This is the headline value of a code-based SDK: assertions live with the test.

## Actual behavior

The registered body is stored but never called. `benchmark` imports only the test-case *data* and grades server-side via `/api/evaluate` using the selected `-e` evaluator. The body’s `expect()`/`judge()` are silently ignored.

### Evidence (from the published bundle)

`test()` stores the body as `evaluate` and pushes it to the registry:
```js
function test(name, optionsOrBody, maybeBody) { ... let evaluate; ... 
  registry.push({ name, options, evaluate, sourceFile, benchmarkPath }); }
```
…but `evaluate` is **never invoked** anywhere in `lib/dist` or `cli/dist` — it appears only in the two locations above (declaration + `registry.push`). There is no `evaluate(...)` / `.evaluate(...)` call site.

The import mapping that `benchmark` sends to the server keeps only:
```js
{ name, category, difficulty, initialPrompt: tc.options.prompt,
  context: tc.options.context, labels, sourceFile, sourceHash,
  description: tc.options.description }
// note: no expectedOutcomes / expectedTrajectory, and no evaluate body
```

So pass/fail comes solely from the server evaluator (`POST /api/evaluate`), which requires `expectedOutcomes` — a field the code-based path never sets.

## Consequences

- `expect(...)` assertions in the body (e.g. on `result.agentOutput` or `result.trajectory`) are no-ops.
[...]

Issue #242, opened 2026-06-04
Judge/evaluator validation errors are silently reported as completed runs with metrics=0 and a misleading reason


ashwin-pc:
## Summary

A **non-retryable judge/evaluator validation error** is reported as a `completed` run with all metrics `0`, and the surface explanation shown to the user is misleading. "Evaluator could not run" is indistinguishable from "agent scored 0", which corrupts benchmark results and wastes debugging time.

## Version

`agent-health` 0.5.8 (CLI `benchmark -f <file> -a <agent>`, code-based SDK test using `test()`/`judge()`).

## What happened

A code-based test invoked the agent successfully and called `judge(result, criteria)`. The judge could not run because of a hard validation error, yet the run was recorded as completed/failed with a misleading reason. The report record contained:

```
status            : "completed"
metricsStatus     : "error"
metrics           : { accuracy: 0, faithfulness: 0, trajectory_alignment_score: 0, latency_score: 0 }
llmJudgeReasoning : "Waiting for traces to become available..."     # <- misleading surface message
traceError        : "Judge evaluation failed: Bedrock Judge validation error (not retryable):
                     Missing required field: expectedOutcomes or expectedTrajectory"
```

The benchmark summary showed `0/1 passed (0%)`, identical to a genuinely poor agent run.

## Why this is a problem

1. **Silent degradation of a hard error into a score.** A non-retryable configuration/validation failure (`metricsStatus: "error"`) is folded into a normal `completed` result with `metrics: 0`. Consumers (CLI summary, UI, exported report) cannot tell "the evaluator was misconfigured / never ran" from "the agent answered terribly". This silently poisons aggregate pass rates.
2. **Misleading surface message.** The user-visible `llmJudgeReasoning` says *"Waiting for traces to become available…"* while the real, terminal cause is *"Missing required field: expectedOutcomes"* (only present in `traceError`). The displayed reason actively points debugging in the wrong direction.

## Expected behavior

- When the evaluator cannot run (validation error, non-retryable), the result should carry a distinct **`error` / `inconclusive`** status — not `completed` with `metrics: 0`.
- Such cases should be **excluded from pass-rate aggregation** (or counted as errored), not as failures.
- The user-facing reason should reflect the **actual** terminal error, not a generic "waiting for traces" message when no further trace fetch will change the outcome.

## Repro sketch

```js
// foo.eval.js
const { test, judge } = require('@opensearch-project/agent-health');
[...]

PR #252, opened 2026-06-04
feat(traces): trace correlation conventions + run-detail UX overhaul

goyamegh:
### Summary

Adds trace correlation conventions so any agent type's spans automatically correlate with agent-health's `test_case` eval span (single trace tree by default), plus a focused batch of run-detail UX fixes verified against real Claude Code, Kiro, and demo runs.

### Why

When agent-health ran a test case, the `test_case` eval span and the agent's own spans landed in OpenSearch as **two unrelated trace trees** with different `traceId`s — the run-report Traces tab showed only the eval span and missed 50+ agent spans (`claude_code.interaction`, `claude_code.llm_request`, `claude_code.tool.execution`, etc.). Per-agent eyeballing required hand-crafted OpenSearch queries on `service.name` + time window, which any teammate sharing the cluster would foul. This PR establishes a layered convention any new agent connector can opt into in one line.

### Three layered correlation strategies

| Strategy | Mechanism | Result |
|---|---|---|
| **A** W3C trace context | `TRACEPARENT` env (subprocess) / `traceparent` header (HTTP) | Single unified trace tree (eval span = parent of agent's root) |
| **B** Run-ID attribute | `gen_ai.request.id == runId` | Loose link, separate trees joined by query |
| **C** Service-name + time-window | `service.name=<agent>-agent AND time IN [run.startedAt − slack, run.endedAt + slack]` | Best-effort fallback, always-on (cost-noise tradeoff favored visibility) |

Per-connector defaults registered: `claude-code` / `kiro` / `pi` / generic `subprocess` propagate via env; `agui-streaming` / `rest` / `openai-compatible` / `langgraph` propagate via header. Each declares a `traceContext.serviceName` so Strategy C works out of the box. Users override per-agent in `agent-health.config.ts`.

### Companion fixes batched in

* **Claude Code parser**: `parseJsonEvent` was checking `event.type === 'tool_result'` at the top level, but Claude Code wraps tool results in `type: 'user'` messages with `tool_result` content blocks → tool **calls** appeared in the trajectory but tool **results** were silently dropped. Verified: regression test case jumped from 0 → 19 `tool_result` steps.
* **Run-detail right-pane UX**: removed the "Overview" tab (everything it surfaced is already in the compact run-card header), default Traces sub-view is now the trace tree (was the info card), Pretty/Raw toggle on span attributes for JSON-shaped strings.
* **Test case definition collapsible**: drops at the top of `TestCaseInspectorPanel` so both `TestCaseDetailPage` and `RunInspectorPage` share the same widget. SDK tests show `sourceFile` path; JSON tests show full untruncated JSON. Auto-collapse main app sidebar on these dense multi-pane pages.
* **Always-on correlation**: removed the opt-in checkbox; the run-report always issues all three correlation clauses unioned via `bool.should`. The empty-by-default trace tree was a worse cost than the noise risk.
* **Inline bottom drawer for span details** (was a side panel that cramped the timeline at typical viewports), matches fullscreen UX.
* **Escape page ScrollArea**: Traces tab no longer overflows the page. Per-tab `overflow-y-auto` on the simple top-down tabs; Traces opts out and manages its own internal scroll.
* **Fullscreen Y-scroll fix**: chart wrapper had `overflow-hidden h-full` clipping tall traces. Now `overflow-x-hidden + overflow-y-auto` preserves the X-axis pan UX while letting a real vertical scrollbar engage.
* **Wider window-fallback default**: 30-min lookback when `performanceMetrics.durationMs` is missing on older runs.
* **Storage layer fix**: `toTestCase` was dropping `sourceFile`/`sourceHash` so `isSdk` checks were always false.
* **Removed redundant fullscreen X**: dialog already owns the only close affordance.
* **LLM Judge reasoning markdown detection**: only render as markdown when actual markdown syntax is present (bold, headings, code, lists ≥2 items, links, blockquotes). Plain prose with a single leading "1." renders as `whitespace-pre-wrap` so Bedrock judge output stops fragmenting into "another box".

### Verification end-to-end

* `tsc --noEmit` clean
* `npm run build:server` succeeds
* **38 test suites / 821 tests pass** across all directly-touched dirs (connectors, traces, tracesService, telemetry, components/traces)
* Manually verified: pointing the upstream Claude Code binary at a local OTLP sink produced **1 root + 11 LLM + 19+19+19 tool spans on a single trace** (vs 1 lonely eval span before)

### Tests added

* `tests/unit/services/connectors/base/BaseConnector.test.ts` — 5 new cases for `buildTraceparentEnv()` and `injectTraceparentHeaders()`. Registers a real `AsyncLocalStorageContextManager` + `W3CTraceContextPropagator` so the helpers see actual context flow.
* `tests/unit/services/connectors/claude-code/ClaudeCodeConnector.test.ts` — fixture updates to use the real Claude Code wire format (`type:'user'` messages with `tool_result` blocks).
* `tests/unit/server/services/tracesService.test.ts` — error-message assertion updated for the new `agents` field validation.

### Docs

`AGENTS.md` gets a new **"Trace correlation conventions"** section explaining all three strategies, the per-agent override pattern, and the per-connector `serviceName` defaults table.

### Out of scope (separate PRs)

Several adjacent fixes that landed on `main-goyamegh` during this work are **not** in this PR — `judge` data-model unification, SDK `evaluatorId` honoring, evaluator-error patches (#242), inspect-page spacing (#245). They each warrant their own PR.

### Issue / Description


goyamegh:
Thanks for the careful review. All 8 comments addressed in commit `33cbdabe` (just pushed):

| # | File | Comment | Fix |
|---|---|---|---|
| 1 | `server/services/tracesService.ts` | `useUnion` only flips when *different* strategies present; multiple agents windows would AND not OR | Now counts `agents.length` toward the union threshold so 2+ windows alone trigger `bool.should + minimum_should_match: 1` |
| 2 | `components/RunDetailsContent.tsx` (judge-failure path) | `metricsStatus: 'error'` set but `traceError` not, banner shows "Unknown error" | Now also sets `traceError: message` so the trace-mode banner shows the real failure reason |
| 3 | `components/RunDetailsContent.tsx` (Strategy C service-name) | Hard-coded protocol→serviceName lookup bypasses `AgentConfig.traceServiceName` override | Resolution priority is now: `agentConfig?.traceServiceName` → protocol map → `<agentKey>-agent` fallback. The `traceServiceName` declared in `agent-health.config.ts` (or `lib/constants.ts` for built-ins) wins, restoring parity with `AgentTracesPage`. |
| 4 | `services/connectors/base/BaseConnector.ts` | `buildTraceparentEnv()` rejects all-zero traceId but emits invalid traceparent on all-zero spanId | Now also rejects all-zero spanId per W3C spec |
| 5 | `services/connectors/types.ts` | Broken doc reference to `docs/AGENT.md` | Fixed to `AGENTS.md` |
| 6 | `AGENTS.md` | Strategy C documented as opt-in via UI checkbox; current behavior is always-on | Section rewritten: describes always-on behavior, the noise-vs-empty-by-default tradeoff that motivated the promotion from opt-in, and the window-derivation rules (`durationMs + 60s` when set, 30-min lookback otherwise) |
| 7 | `CHANGELOG.md` | Same opt-in-checkbox stale text | Strategy C entry now matches shipped behavior; documents the override priority |
| 8 | `observio-sample-agent/src/main.ts` | `dotenv.config({ path: '../.env' })` resolves CWD-relative; npm scripts launched from repo root miss the parent `.env` | Both `dotenv.config` calls now use `path.join(__dirname, ...)`. Same fix applied to `main_ag_ui.ts` (had the identical bug) for consistency, per the reviewer's hint |

Verification:
* `tsc --noEmit` clean
* **38 test suites / 821 tests pass** across all touched dirs
* DCO is now green (the rebase also corrected an author/signoff email mismatch on `e3102dbd` — `goyamegh@amazon.com` was lingering from before I'd set the worktree-local git config)

Force-pushed because the DCO fix required rewriting history. Diff vs the prior push is in `33cbdabe` only.

jasonlhamazon (approved):

PR #247, opened 2026-06-04
fix(judge): surface evaluator failures distinctly instead of as agent failures (#242)

goyamegh:
## Description

Fixes #242.

A non-retryable judge / evaluator validation error used to be recorded as a normal `completed` run with `metrics: { all 0 }`, `passFailStatus: failed`, and the surface field `llmJudgeReasoning` left at the misleading `"Waiting for traces to become available..."` placeholder. The actual cause (e.g. `Missing required field: expectedOutcomes or expectedTrajectory`) was hidden in a separate `traceError` field that nothing surfaced in the Judge tab or in benchmark summaries. Aggregate pass rates absorbed the false zero; debugging pointed users at trace propagation when the real problem was a misconfigured evaluator.

### Evaluator-error surface (the core fix)
- New `services/evaluation/evaluatorError.ts` exposes `buildEvaluatorErrorPatch(kind, error)` — the canonical patch every error site uses to persist `metricsStatus: error`, a kind-prefixed `traceError`, an explicit human-readable `llmJudgeReasoning` (`**Evaluator could not run.** ... **Reason (kind):** <message>`), `passFailStatus: undefined`, and zeroed `metrics`.
- 6 call sites converted: `tracePoller.ts` (4 error paths), `evaluationRunner.ts`, `benchmarkRunner.ts`, `browserRecovery.ts`.

### Stats: `errored` is its own bucket
- `RunStats` extended from `{passed, failed, pending, total}` → `{passed, failed, pending, errored, total}`. Optional for backward compat (older runs read as `errored: 0`).
- All 4 stats-computation sites bucket `metricsStatus === error` reports as `errored` instead of `failed`.
- Pass rate denominator is now `total - errored` so a misconfigured judge can't drag the score to 0%.

### UI
- `ResultStatus` gains an `errored` variant with an amber `AlertTriangle` icon and `ERRORED` badge text. All callers (`RunInspectorPage`, `BenchmarkRunDetailPage`, `RunDetailsPage`, `RunDetailsContent`) inherit it without code changes.
- Benchmark runs list, run-detail header, eval-run detail card, and CLI summary all surface the errored count distinctly with a tooltip explaining “Excluded from pass-rate aggregation.”

### UX label cleanup
- `LLM Judge Reasoning` → `Judge Reasoning` (deterministic / SDK matcher reasonings now flow through the same field).
- `Conversation History` → `Test Case Output` (the streaming view shows tool calls + tool results, not just chat).

## Tests

[...]

PR #259, opened 2026-06-05
fix(ui): surface ERRORED status consistently across run/test/benchmark surfaces (#242 follow-up)

### LINKED ISSUE #242: Judge/evaluator validation errors are silently reported as completed runs with metrics=0 and a misleading reason

goyamegh:
## Description

Follow-up to [#247](https://github.com/opensearch-project/agent-health/pull/247) (closes [#242](https://github.com/opensearch-project/agent-health/issues/242)).

The original PR introduced the `'errored'` status — the dedicated bucket for runs where the evaluator could not produce a verdict (`metricsStatus: 'error'` on the report) — but only some of the run/test/benchmark surfaces were actually wired through to the new bucket. Reports landing on the inspector right pane, for example, showed a grey `PENDING` chip instead of amber `ERRORED` because the badge config table had no entry for the new status (you can see this exact bug in the screenshot review feedback on #247: a run with `metricsStatus='error'` and a fully-rendered `**Evaluator could not run.**` Judge tab still showed `PENDING` in the top-right pill).

This PR sweeps every UI/data surface that displays per-run / per-test / per-benchmark status and wires the new bucket through.

### Inspector surfaces
- **`TestCaseInspectorPanel`** (`components/evals3/TestCaseInspectorPanel.tsx`): added the `errored` entry to the badgeConfig + an `isErrored` derivation that wins over passed/failed so `metricsStatus === 'error'` always renders the amber `ERRORED` pill.
- **`RunInspectorPage`** (`components/evals3/RunInspectorPage.tsx`): added `erroredCount` calculation + amber `⚠ N` badge in the top header next to passed/failed.
- **`RunDetailsFlyout`** (`components/evals3/RunDetailsFlyout.tsx`): local `getResultStatus` previously mapped `metricsStatus === 'error'` to `'failed'`. Now returns `'errored'`, matching the shared helper in `ResultStatus.tsx`.

### Test-case surfaces
- **`TestCaseDetailPage`** (`components/evals3/TestCaseDetailPage.tsx`): per-test-case run history list rendered errored rows as red `XCircle`. Now renders amber `AlertTriangle`. The pass-rate denominator excludes errored runs so a misconfigured judge can't drag the per-test-case score to 0%.
- **`TestCasesPage`** (`components/evals3/TestCasesPage.tsx`): `getPassFail` folded errored runs into `'fail'`. Added a dedicated `'errored'` bucket with amber chip styling.
- **`TestCaseRunsPage`** (legacy, `components/TestCaseRunsPage.tsx`): `RunCard` rendered every non-passed run as red `FAILED`. Now distinguishes errored runs.

### Benchmark surfaces
- **`BenchmarkRunDetailPage`** (`components/evals3/BenchmarkRunDetailPage.tsx`): the Passed/Failed summary boxes had no Errored counterpart. Added one (only shown when `erroredCount > 0`) so the per-run benchmark detail page matches the Eval Run Detail card already updated in #247.
- **`EvalRunsPage`** (`components/evals3/EvalRunsPage.tsx`): per-row count `passed/failed/total` now also shows `errored` from `RunRow.errored` (reads `stats.errored` optional field, falls back to `0` for legacy runs).
- **`BenchmarkSummaryCharts`** (`components/benchmarks/BenchmarkSummaryCharts.tsx`): aggregate statistics excluded errored runs from passed/failed (good — they were silently dropped) but never surfaced the count anywhere. Now tracks `erroredCount` per run; pass-rate denominator excludes them.

### Inline / comparison surfaces
- **`QuickRunModal`** (`components/QuickRunModal.tsx`): status badge had no `ERRORED` branch. Added one between the existing `PENDING` and `passFailStatus`-based branches.
- **`RunSummaryPanel`** (`components/RunSummaryPanel.tsx`): stats counting + donut chart now have an `'errored'` slice (amber `#f59e0b`). Pass-rate denominator excludes errored runs.
- **`MetricCell`** (`components/comparison/MetricCell.tsx`): the side-by-side comparison cell folded errored results into `'Failed'` because the result row only carried `passFailStatus`. Now reads a new `result.errored` flag and renders the amber `Errored` chip distinct from `Failed`.

### Type + service plumbing
- **`TestCaseRunResult`** (`types/index.ts`): gains an optional `errored?: boolean` flag.
- **`comparisonService.ts`**: sets `errored: report.metricsStatus === 'error'` on every comparison row built from a report. The comparison cell reads this flag instead of routing through `passFailStatus` (which is `null` on errored docs and would silently fall through to `Failed` styling).

## Tests

- 1 new regression test in `tests/unit/services/comparisonService.test.ts` (`flags reports with metricsStatus="error" as errored on the row`) pins the new `errored` flag's plumbing end-to-end against the existing comparison fixtures. Asserts the flag is `false` for passed/failed reports and `true` for errored reports, and that this works even when `passFailStatus` is `null` on the underlying report.
- All existing 47 `comparisonService` tests updated for the new `errored?: false` expectation on the row shape.

## Backward compatibility

- `TestCaseRunResult.errored` is optional. Old comparison rows persisted before this change read as `errored: undefined` → falsy → existing pass/fail rendering paths unchanged.
- All `stats.errored` reads are `?? 0` so older stored runs without the field continue to render correctly (no UI regression for legacy benchmarks).
- No storage shape changes; this is a pure read-side / render-side fix.

## Issues Resolved

Follow-up to [#247](https://github.com/opensearch-project/agent-health/pull/247) which closed [#242](https://github.com/opensearch-project/agent-health/issues/242).

## Check List

- [x] All tests pass (`npx tsc --noEmit` clean, all comparisonService unit tests green — 48/48)
- [x] New functionality has been documented in the CHANGELOG


goyamegh:
## Force-pushed `fb638e50` — Copilot review feedback addressed

The single Copilot review comment on `components/RunSummaryPanel.tsx` was a real correctness bug, not just stylistic. Fixed inline:

### Before

```ts
if (result.status === 'completed' && result.reportId) {
  const report = reports[result.reportId];
  if (report) {
    reportCount++;                                                   // ← always counted
    const score = getRunOverallScore(report.metrics);
    if (score !== null) totalScore += score;                         // ← always aggregated

    if (report.metricsStatus === 'error') {
      errored++;
    } else if (report.passFailStatus === 'passed') {
      ...
```

The metricsStatus check came **after** the score aggregation, so:

1. **Errored reports** (`metricsStatus: 'error'`, zeroed metrics from `buildEvaluatorErrorPatch`) still contributed `0` to `totalScore` and bumped `reportCount`. The donut's center number `Avg Score` was being dragged toward zero by every misconfigured judge run — exactly the scenario the dedicated `errored` bucket was supposed to keep separate.
2. **Pending / calculating reports** (trace-mode in flight) carry placeholder metrics (`{ accuracy: 0, ... }` from the runner's TRACE-mode return). Those were also being aggregated as if they were real scores.

### After

```ts
if (report.metricsStatus === 'pending' || report.metricsStatus === 'calculating') {
  pending++;
  return;                          // skip — no score yet
}
if (report.metricsStatus === 'error') {
  errored++;
  return;                          // skip — zeroed metrics aren't a real score
}

// Real verdict — only here do we contribute to the average.
const score = getRunOverallScore(report.metrics);
if (score !== null) {
  totalScore += score;
  scoredCount++;                   // separate counter, not `reportCount`
}
if (report.passFailStatus === 'passed') passed++;
else failed++;
```

The new ordering mirrors `lib/runStats.ts`'s canonical bucketing exactly (which the comment correctly cited as the source of truth):

| metricsStatus | Old: reportCount | Old: totalScore | New: scoredCount | New: totalScore |
|---|---|---|---|---|
| `'pending'` / `'calculating'` | ✓ counted | ✓ added (placeholder zero) | — | — |
| `'error'` | ✓ counted | ✓ added (zeroed metrics) | — | — |
| `'ready'` / undefined | ✓ counted | ✓ added (real score) | ✓ counted | ✓ added (real score) |

Also renamed `reportCount` → `scoredCount` so the variable name reflects what it actually counts. The returned stats object now also exposes `pending` for completeness.

### Impact

- For a benchmark run with 1 passed (90%), 1 failed (40%), and 1 errored (zeroed): `Avg Score` was previously `(90 + 40 + 0) / 3 = 43%`; now it's `(90 + 40) / 2 = 65%` — actually reflects judge output instead of placeholder zeros.
- Same fix applies to in-flight pending/calculating runs that briefly populate the panel before traces arrive.

### Verified

- `npx tsc --noEmit` clean.
- `tests/unit/services/comparisonService.test.ts` — 48/48 passing (the new regression test for the `errored?` flag plumbing remains green).
- All 16 files in this PR still compile, no other surfaces affected by the change.

jasonlhamazon (approved):

TackAdam (approved):

Issue #256, opened 2026-06-05
[RFC] Test SDK v2 — a real Playwright-style harness (control inversion, single engine, trace-driven insights)


ashwin-pc:
# RFC 004: Test SDK v2 — a real Playwright-style harness

| Field        | Value                                                              |
|--------------|--------------------------------------------------------------------|
| **Status**   | Draft                                                              |
| **Author**   | Ashwin P Chandran                                                  |
| **Created**  | 2026-06-05                                                         |
| **Supersedes** | the experimental `test()/judge()/expect()` SDK (#207, #198)      |
| **Resolves** | #245, #243, #242, #244, #238, #248, #230, #232, #246              |

## 1. TL;DR

The current SDK is *"a thin layer over the existing UI/JSON execution path"* (its own words, #238). That single decision — keep the framework in charge of invoking the agent **eagerly**, and only call the user's body afterwards to *assert on* a result — is the root cause of almost every bug filed against it.

The fix is one idea, applied without compromise: **invert control.** The test body owns execution. The framework supplies *capabilities* (an `agent` fixture, `expect`, `judge`, `traces`) exactly the way Playwright supplies `page` — and the body decides when and how to use them. There is **one** execution engine behind both the CLI and the server, **one** grading surface, and **no** global mutable state.

This collapses the current "two runners × two entry points × two grading systems × eager invocation × global session" combinatorial mess into a single linear pipeline.

---

## 2. Why the current SDK fails (one root cause, many symptoms)

The current execution model (verified in `services/evaluationRunner.ts`, `services/benchmarkRunner.ts`, `lib/testCases/*`, `server/routes/storage/benchmarks.ts`):

```
connector.buildPayload(prompt, context)   ← framework reads test() options
  → invoke agent                          ← framework, eagerly, before any user code
  → buildEvalResult(...)                  ← framework
  → startSession() [GLOBAL singleton]     ← framework
  → evalFn({ result, judge, traces })     ← user body runs LAST, only asserts
  → endSession()                          ← framework collects verdicts
```

Five structural problems fall out of this:

| # | Structural flaw | Symptoms it causes |
|---|-----------------|--------------------|
| A | **Eager, framework-owned invocation.** The body runs *after* the agent, so it can never influence the prompt, seed state first, invoke twice, or do multi-turn. | #238 (no `invoke`), #248 (hooks run after the agent, so "seed a resource the agent reads" can't work) |
[...]

Issue #248, opened 2026-06-04
[SDK] Hook execution should wrap the agent invocation, not just the body


goyamegh:
## Summary

The lifecycle hooks introduced by #229 / PR #241 currently run **inside the deterministic-eval block, after the connector has already returned**. So `beforeAll` / `beforeEach` fire *after* the agent has run, which means the headline use case from #229 ("materialize an out-of-band resource the agent reads") doesn't actually work as intended. This issue tracks the three coupled changes needed to make hooks first-class around the agent invocation, surfaced by the Copilot review on PR #241.

No bug today for body-driven assertions (writing markers, asserting on the trajectory after the fact); this only matters for hooks whose work needs to land *before* the agent runs.

## Concrete observed behaviour

From [services/evaluationRunner.ts](https://github.com/opensearch-project/agent-health/blob/main/services/evaluationRunner.ts), with PR #241 merged:

```ts
// 1. Agent runs first.
let report = skipAgentInvocation
  ? synthesizeEmptyReport(...)
  : await runEvaluationWithConnector(...);

// 2. Then we enter the deterministic-eval block …
if (hasDeterministicEval) {
  // … and only here do hooks fire.
  const before = await hookOrchestrator.beforeTest(desc);
  …
  await evalFn(arg);                                 // body
  const after = await hookOrchestrator.afterTest(desc, fixtures);
}
```

Result: `beforeEach` can't `mkdtemp` a workspace dir for the agent to read, can't seed a DB before the agent connects, can't start a sandbox before the connector spins up the LLM. Same code path duplicated in [services/benchmarkRunner.ts](https://github.com/opensearch-project/agent-health/blob/main/services/benchmarkRunner.ts) (separately tracked in #246 as the parallel-runner duplication that lets these regressions through).

### 1. Move hook execution to wrap the agent

```
beforeTest()  ──→  build connector payload  ──→  agent runs  ──→  body  ──→  afterTest()
```

`beforeAll` / `beforeEach` run before `runEvaluationWithConnector(...)`. `afterEach` / `afterAll` run after both the agent and the body, in `try/finally` so they always fire on body or hook failure.

When `beforeAll` / `beforeEach` aborts, the agent is skipped (we already abort the body today); the test is marked failed and `afterEach` / `afterAll` still run.

### 2. Surface `provisioned` to the connector

Without this, even with reordering the agent has no way to read the values `beforeEach` wrote. Two options to evaluate:

[...]

Issue #246, opened 2026-06-04
Unify run surfaces: one document, one URL, one UI for all runs


goyamegh:
## Summary

Today there are three different ways an evaluation run can show up in the system, and they don't share a document shape, a URL, or a UI. This issue tracks consolidating all three layers — **data plane** (which document gets written), **URL surface** (where you land when you share a link), **UI surface** (which page renders) — onto a single canonical shape.

The original scope of this issue was just the data-plane piece (CLI file mode → `evaluation-runs`). Discussion has expanded it to cover the URL and UI consolidation those storage docs feed into, because all three asymmetries have the same root cause: legacy code paths that pre-date the unified `EvaluationRun` document. Filing them together because they're best designed as one — fixing only the data plane leaves the UI showing the same data through three different layouts; fixing only the UI is impossible while CLI file mode still produces a separate document type.

This was noticed while testing #229 + #245 on the running system. Filing as a separate follow-up so #229/#245 can land on their own merits.

> **Spec revision (post-discussion):** the original draft proposed demoting `describe()` to a per-test grouping label. That's reversed in this version. **`describe()` blocks remain a way to create benchmarks**, with the same identity rule as the CLI's `-n` flag — they're two equivalent inputs for "associate these tests with a saved benchmark". The actual bugs are (a) the orphan-tests file-default benchmark created without user intent, (b) `Benchmark.runs[]` growing unbounded as a nested array, and (c) two parallel runner pipelines that don't share regression tests. The fix is to keep `describe()` ↔ benchmark as a real operation, but route everything through `EvaluationRun` as the canonical execution document. See "Proposed end-state" below.
>
> **All runs of an evaluation share one view, regardless of cardinality.** A run with one test case (today: `/api/evaluate` produces a bare `EvaluationReport`) is conceptually a 1-test `EvaluationRun`. After this issue lands, every entry point — including `/api/evaluate` — writes an `EvaluationRun` document; the run detail page renders 1-test and N-test runs identically (the left list just has fewer rows). There's no separate "standalone report" UI, no separate URL family, no separate document type. Whether a run is benchmark-attached or orphan is a property (`benchmarkId` set/unset), not a layout. Promotion (orphan → benchmark) is one of three equivalent ways to attach a run to a benchmark, and is exposed as a contextual button on the run detail page — not a separate concept.

---

### 1. Data plane — six entry points, three storage shapes

| Entry point | Endpoint hit | Document written | URL pattern | Body invocation runner |
[...]

Issue #244, opened 2026-06-04
Allow defining custom evaluators in code/config (SDK parity with connectors)


ashwin-pc:
## Summary

Custom **connectors** can be defined in code and version-controlled (`registerConnector(...)` + `defineConfig({ connectors: [...] })` in `agent-health.config.ts`). Custom **evaluators** have no equivalent code/SDK/config path — they appear to be created and versioned only as server-side state (via the web UI: `/evaluators/new`, `/evaluators/:id/edit`, `/evaluators/:id/versions`). There is no exported `registerEvaluator`/`defineEvaluator` and no `defineConfig({ evaluators: [...] })` authoring path.

## Version

`agent-health` 0.5.8.

## Context / motivation

We're building a domain benchmark where the agent emits a structured outcome plus free-text reasoning. We want:
- deterministic, label-based scoring of the outcome (done in a code-based eval), and
- a **judge rubric** scoring agent-specific reasoning dimensions, e.g. a handful of weighted metrics like `dimension_a`, `dimension_b`, `dimension_c`, `dimension_d`.

The judge rubric is exactly an **evaluator** (system prompt + weighted metrics), which the framework already models well — `list evaluators` shows system evaluators (e.g. `system-factuality`) with `systemPrompt` + `scoringConfig.metrics[{name, weight, scale}]` + versions.

## Problem

Because evaluators can only be authored as server-side objects (UI/API), a custom rubric:
- is **not version-controlled** in the repo alongside the connector and test cases,
- is **not reproducible** from a clean clone / CI,
- drifts from the code that depends on it (the `-e <evaluator-id>` selection in the benchmark refers to an id that exists only on a particular server).

This is inconsistent with how connectors (and the rest of the config) are authored, and it breaks the "config-as-code / reproducible benchmark" model.

## Request

Provide an SDK/config authoring path for custom evaluators, parity with connectors. For example:

```ts
import { defineConfig, defineEvaluator, registerEvaluator } from '@opensearch-project/agent-health';

const myRubric = defineEvaluator({
  id: 'my-reasoning-rubric',
  name: 'My reasoning rubric',
  systemPrompt: `...rubric...`,
  scoringConfig: {
    metrics: [
      { name: 'dimension_a', weight: 0.3, scale: 100 },
      { name: 'dimension_b', weight: 0.3, scale: 100 },
      { name: 'dimension_c', weight: 0.2, scale: 100 },
      { name: 'dimension_d', weight: 0.2, scale: 100 },
[...]

PR #260, opened 2026-06-05
RFC 004: Test SDK v2 — control inversion, judge v2, custom + agentic judge, single code-import path

ashwin-pc:
# RFC 004 — Test SDK v2 (Playwright-style harness)

Implements [RFC 004 / #256](https://github.com/opensearch-project/agent-health/issues/256): a redesign of the experimental code-based test SDK around control inversion (the test body owns agent invocation), a pluggable run-scoped judge, custom evaluators, and a single code-import execution path.

> **Draft / stacked.** This branch is stacked on **#257** (run-level `evaluatorId` in `judge()`), so the diff currently includes #257's commits — they drop out once #257 merges and this rebases on `main`. It's also large (multi-phase); happy to split into per-phase PRs if preferred.

## What's in here

| Area | Change |
|---|---|
| **Phase 1 — concurrency-safe sessions** | `runInSession()` (AsyncLocalStorage) replaces the module-global matcher session so concurrent test bodies don't cross-contaminate verdicts. |
| **Phase 2 — control inversion** | New `agent` fixture: the body calls `const result = await agent.run(prompt?)` (one invocation per test, enforced). Extracted a pure `invokeAgent()`; both runners rewired; captured result reflected into fixtures so hooks see it (#248). All `evals/*.eval.js` migrated. `agent.run(prompt, { env })` is now forwarded to the connector (merged into `connectorConfig.env`) instead of being dropped. |
| **Phase 3 — judge v2** | Non-throwing `Verdict` with `gate`/`observe` roles, `.orThrow()`, `expect(v).toPass()`, `errored` status (vs `failed`), skip (`AH_SKIP_JUDGE` / `{skip}`), content-addressed cache. |
| **Phase 4 — evaluators + `agent` trace judge (#244)** | `defineEvaluator()` + `evaluate` fixture (in-process custom checks); a pluggable **`agent` (trace) judge** — an LLM judge with a restricted, read-only, run-scoped trace-tool pack (`query_spans`/`query_logs`) so it verifies claims against the run's **real OTel spans/logs**, not just the trajectory text. Selected per-run via `evaluator.inferenceConfig.provider: 'agent'`. |
| **Phase 5** | `result.traces` exposed on the run result. |
| **Phase 6 — single code-import path (#245/#246)** | Unified the benchmark-run code-import resolution into `resolveCodeFnMapForStoredTestCases()`; **`benchmark -f *.eval.js` now executes the SDK body** (routes to `code-import`) instead of importing inert data. |
| **Phase 7 — loader unification (#232)** | One authoring surface shared across `.js`/`.ts`/`.mjs` loaders (kills the `.js`-vs-`.ts` export drift). |
| **Migration tooling** | `agent-health migrate sdk-v2` codemod rewrites v1 `({ result })` bodies → `({ agent }) => { const result = await agent.run(); … }` (AST-based, `--dry-run`, idempotent). |
| **Fixes surfaced while validating end-to-end** | `evaluationRuns.list()` queried `docType` (analyzed text) instead of `docType.keyword` — silently returned 0, hiding **every** evaluation run; UI Runs list now merges top-level eval-runs with legacy benchmark-embedded runs (temporary, documented — see below). |

Advances (per the RFC): #245, #243, #242, #244, #238, #248, #230, #232, #246. Builds on merged #247 and open #257.

## The `agent` (trace) judge — in-process, self-contained
The trace judge runs **in-process via the pi SDK** (`createAgentSession`), not by spawning a CLI — so there's no subprocess, no stdout parsing, no extension file, no env-var scoping, and nothing to install on `PATH`:
- pi ships as an **`optionalDependency`** (`@mariozechner/pi-coding-agent`) loaded by dynamic import — a library import, self-contained per install.
- Trace tools are registered via an **in-process extension factory**; `runId` is captured by closure and the judge is **allow-listed to only** `query_spans`/`query_logs` (hard-scoped — it cannot read the project filesystem or other runs).
- Model selection prefers the run's configured judge model (region-aware Bedrock inference-profile match — bare Claude 4.x ids fail on-demand), else a recent Claude.
- Verdicts carry no `improvementStrategies` (per RFC: recommendations live in the insights layer).
- The provider id is `agent` (the pi implementation detail is not exposed in config/types).
- `/api/judge` returns `400` for `provider: 'agent'` without a `runId` (fail loud rather than emit a trajectory-only no-op verdict).

## Run-model note (intentional temporary merge)
There are two disjoint run records — legacy `benchmark.runs[]` (`run-…`) and top-level evaluation-runs (`eval-run-…`). `EvalRunsPage` now **merges** both so neither is invisible. This is **deliberately temporary**, documented on `RunRow`: the long-term goal is to converge on the evaluation-run model (converge the benchmark write path → eval-runs, migrate embedded runs via `migrate evaluation-runs`, then drop `benchmark.runs[]` and the merge).

## Validation
Migrated + ran a real consumer eval project end-to-end through the runner against a live subprocess agent + OpenSearch (storage + observability):
- **Contextual (Bedrock) judge**: control inversion, `expect` gates, non-throwing `judge` with gate-role reasoning, `benchmark -f` → code-import, run visible in the UI Runs list.
- **`agent` (trace) judge, organic run**: 0 CLI spawns (fully in-process), `query_spans` invoked against the run's real spans, both gate verdicts pass grounded in span data ("Examination of the actual spans reveals…"), run persisted + viewable. A side-by-side comparison showed the trace judge correcting a contextual false-negative (the contextual judge only saw a truncated trajectory; the trace judge found the actual reasoning in the spans).

## Test / CI status
`build-and-test (18 / 20 / 22)` is **green** — the full unit + integration suites pass on this branch (same as #247 on `main`). New unit coverage for the in-process judge (`traceJudgeTools`, model matching, NDJSON/text extraction), the `env` forwarding, and the `runId` guard.

> An earlier revision of this description claimed 4 component suites fail "on `main`." **That was incorrect** — they fail only in a *local* `node_modules` resolution artifact; CI's clean install passes them on both `main` and this branch. This PR introduces no test regressions.

`Code-Diff-Analyzer` is an advisory AI reviewer.

## Notes
- **Changelog**: collapsed to concise `[Unreleased]` entries; reset to mainline to clear union-merge link rot (a generated changelog would be a better long-term fix).
- **Deployment of the `agent` judge**: pi needs model credentials (same Bedrock/Anthropic auth as the rest of agent-health); the published tarball must keep the SDK reachable for installed consumers.
- **`npm audit`**: the 2 `high` advisories (`react-router`) are pre-existing frontend deps, **not** introduced by the optional pi dependency.


ashwin-pc:
Thanks for the automated review passes. Triage of the flagged items, with what changed in `1dcdab4`:

### Acted on

**`AgentRunOptions.env` was silently dropped** (PR Code Suggestions → "Forward env from AgentRunOptions"). Real contract gap — fixed in `1dcdab4`. Note the suggested "merge into the invocation test case" wouldn't have worked: `TestCase` has no `env` field, so it'd be a no-op. The actual path is `AgentRunOptions.env` → `invokeAgent` → `connectorConfig.env` (per-call wins over static config) → `SubprocessConnector` spawn env. Covered by new tests in `tests/unit/services/evaluation/index.test.ts` and `tests/unit/services/evaluationRunner.deterministic.test.ts`.

**Missing `runId` validation on the trace judge** (PR Code Suggestions → "Validate runId presence"). Added a `400` guard in `1dcdab4`. Framing as a fix is right, but it's a fail-loud improvement (the trace tools have nothing to scope to without `runId`), not a security fix — see below.

### Won't fix — with rationale

**Code-Diff-Analyzer: "cross-tenant data access" (medium), `judge.ts`.** There are no tenants. Agent Health is a single-user localhost tool (a Bedrock proxy the browser can't call directly) — no auth middleware, no `req.user`, no tenant model anywhere in the server. `runId` is a correlation key, not a security boundary, so there's nothing to leak across. We added the `400` guard so a missing `runId` fails loudly instead of running a no-op trace judge with nothing to query. Stricter isolation on shared OTel clusters is available per the trace-correlation conventions in AGENTS.md by scoping a connector's `serviceName` to a tenant value.

**PR Code Suggestions: "Move cache after skip check" (9/10).** Stale against current HEAD. `runJudge` already orders the SKIP early-return (judge.ts:322) *before* the cache lookup (judge.ts:343), and skipped verdicts are never cached — `verdictCache.set` only runs on the post-fetch success path (judge.ts:414). No change needed.

**PR Reviewer Guide: `bindJudge({ skip: false })`.** By design. `skip:false` is already the default, and a per-call `skip:false` cannot override a global `AH_SKIP_JUDGE=1` regardless of the early-return guard, because `runJudge` checks `options?.skip || judgeSkippedByEnv()` (env is an independent OR) and `mergeOptions` uses `??`. Fixing the guard wouldn't deliver the described behavior; to clear a bound default, call the unbound `judge`.

### Holding for a follow-up

**PR Code Suggestions: restrict `isCodeFile` to `.eval.*` (`sourceResolver.ts`).** Not applying blindly — this predicate runs on stored test cases' `sourceFile`, and the CLI import path (`benchmark -f ./foo.ts`) can legitimately store a plain `.ts`; tightening here could stop re-loading those modules' hooks/evaluate fns. There are also two separate `isCodeFile` definitions (here + `benchmark.ts`) to reconcile first.

The remaining low-severity items (FNV-1a fallback collision risk, migrate.ts glob writes for a user-invoked codemod, OTel context-undefined degradation, traces accessor error when `agent.run()` isn't called) are acceptable as-is for this tool's threat model and usage.

goyamegh:
Re-reviewed end-to-end with fresh eyes — and now also having read the test files, the storage benchmarks-route changes, the piJudgeService refactor, and `services/evaluation/index.ts`'s `invokeAgent` extraction. Several findings change, several get sharper, one new real bug surfaces.

---

## Real bugs

### 1. OTel context propagation asymmetry between the two runners — breaks W3C trace correlation (Strategy A)

This is the most interesting one. In `services/benchmarkRunner.ts` the SDK invoke callback wraps `invokeAgent` in the eval span's OTel context:

```ts
// benchmarkRunner.ts — control-inversion invoke path
const doInvoke = () => invokeAgent(agentConfig, bedrockModelId, invocationTestCase, {
  registry: connectorRegistry,
});
const inv = caseSpanContext
  ? await context.with(caseSpanContext, doInvoke)
  : await doInvoke();
```

In `services/evaluationRunner.ts` the same invoke callback does **not**:

```ts
// evaluationRunner.ts — control-inversion invoke path
const inv = await invokeAgent(agentConfig, bedrockModelId, invocationTestCase, {
  registry: connectorRegistry,
});
// ↑ no context.with(caseSpanContext, …) wrap
```

`AGENTS.md` documents Strategy A (W3C trace context propagation) as the preferred mechanism for tying an agent's spans to the eval span:

> The `test_case` span is started **before** the connector invokes the agent and is made the active OTel context. … [HTTP/SSE / subprocess] connectors then propagate the context to the agent.

That propagation (the `traceparent` header injection for HTTP/SSE connectors, the `TRACEPARENT` env var injection for subprocess connectors) only fires if the eval span is the *active* OTel context when the connector's `execute()` runs. With the wrap missing on the evaluation-runs path, every code-SDK run launched via `POST /api/storage/evaluation-runs` (which is now the headline path — `benchmark -f *.eval.js` routes here) emits agent spans with no parent linkage to the eval span. Strategy A silently degrades to Strategy C (service-name + time-window) — exactly the "near-empty trace tree by default" failure mode `AGENTS.md` says was bad enough to flip Strategy C from opt-in to always-on.

CI doesn't catch this because none of the unit/integration tests assert OTel propagation end-to-end. The pre-existing `caseSpanContext` plumbing in `executeEvaluationRun` was correct for the *legacy* path (which uses `runEvaluationWithConnector`); the new SDK path forgot to inherit it. Fix is one line — same wrap as the bench runner — but it really should be tested.

This also retroactively explains why the duplicated control-inversion block matters: one of the two divergent copies has a real OTel propagation bug the other doesn't.

### 2. `AgentRunOptions.env` is silently dropped by the runner's `invoke` callback

`lib/testCases/agentFixture.ts` explicitly documents this as a public contract:

```ts
/**
 * Structured values forwarded to the connector's payload builder (e.g. a
 * provisioned workspace dir). Stringly-typed env is the lowest common
 * denominator every subprocess connector inherits automatically.
 */
env?: Record<string, string>;
```

But both runners' `invoke` callback only forwards `context`:

```ts
const invocationTestCase: TestCase = {
  ...testCase,
  initialPrompt: prompt,
  ...(options?.context ? { context: options.context as any } : {}),
  // options.env — silently dropped
};
```

RFC 004 §4.6 ("Connector access from fixtures") leans on this exact capability. There's no test asserting the forwarding works. Either honor it (forward through `connectorConfig` or a new `request.env` field consumed in `connector.buildPayload`) or remove `env` from the public type until it's wired.

### 3. Cross-tenant `runId` exposure on `/api/judge` for the agentic trace judge

The `Code-Diff-Analyzer` bot flagged this; the latest commit (`fix(judge): make pi-agentic trace judge actually runnable`) addressed the wiring, not the threat model. Trace-judge tools are scoped via `process.env.AH_JUDGE_RUN_ID`, set from the POST body's `runId`. There's no ownership check.

The tool-side scoping is rock-solid (`runId` is read once at extension load via `const runId = process.env.AH_JUDGE_RUN_ID;` at module scope, the LLM cannot pass it as a tool param). The leak is purely on the route side. In a shared OTel cluster (which `AGENTS.md` Strategy C explicitly documents as the realistic deployment), an authenticated caller can pass any `runId` and exfiltrate spans/logs through the judge's reasoning text.

Either derive `runId` from the trajectory the request also carries (so the judging principal can only ever judge its own runs), or document this as single-tenant-only and gate the `pi-agentic` provider behind a config flag.

### 4. Codemod's idempotency guard is too coarse — silently leaves half-migrated bodies broken

`lib/testCases/codemod.ts`:

```ts
if (/\bagent\s*\.\s*run\s*\(/.test(bodyText)) {
  notes.push(`skip  ${name}: already calls agent.run()`);
  return;
}
```

This skips any body that mentions `agent.run(`. But consider a body partway through a manual migration:

```js
test('mixed', { prompt: 'p' }, async ({ result, agent }) => {
  const r = await agent.run();           // user migrated half
  expect(result.agentOutput).to.exist;   // … but left this old reference
});
```

Codemod sees `agent.run(` → returns "already migrated". The author runs their evals: the destructured `result` is the empty placeholder (because `agent` is destructured separately and the runner hands a placeholder), so `result.agentOutput === ''`, the `expect` fails, and the migration looks broken when it isn't. The right gate is: skip iff `usesAgent && !usesResult && /agent\.run\(/.test(body)` — i.e. the destructure has already been migrated *and* the call is in. Anything else should be migrated (or at least surface a "needs attention" note).

There's a unit test for "drops result but keeps agent when both were destructured" that just asserts the codemod doesn't change the body — it doesn't catch the broken `result.agentOutput` reference inside. Worth adding.

### 5. `evaluatorRegistry` is global with last-write-wins, never cleared in production

`lib/testCases/evaluators.ts` holds a `Map<string, EvaluatorFn>`. The unit tests (`tests/unit/lib/testCases/evaluators.test.ts`, `loader.surface.test.ts`) clean up via `clearEvaluators()` in `beforeEach`. The runners do not — they call `clearJudgeCache()` per run but never `clearEvaluators()`, and `services/sourceResolver.ts` doesn't either.

In production this means:
- Two `.eval.{js,ts}` files in the same evals/ directory both calling `defineEvaluator('len-check', ...)` with different bodies silently shadow based on load order. No test reproduces this.
- A long-running server process re-loads test files for each new benchmark run. Evaluators from prior runs leak into later ones.

The doc string says this is intentional ("process-global registry by id, like a shared helper library"), but compare to how the test/hook registries are partitioned by `setActiveFile()` in `loader.ts` — those *are* file-scoped. The asymmetry is surprising and the failure mode is silent.

Cheapest fix: throw on duplicate `defineEvaluator(id, ...)` (one-line check). Better fix: scope by `sourceFile` like tests/hooks. Either way, document the chosen semantics.

---

## Design / contract issues

### 6. `bindJudge({ skip: false })` short-circuits, and per-call `skip: false` cannot override env anyway

```ts
if (!defaults || (!defaults.evaluatorId && !defaults.model && !defaults.serverUrl && !defaults.skip)) {
  return judge;
}
```

`!false === true` → `bindJudge({ skip: false })` returns the unbound judge. The doc string says:

> We treat an explicit `undefined` the same as a missing field — callers who want to *clear* a bound default should pass an empty string or call the unbound `judge` directly.

That handles `evaluatorId: undefined`. It does *not* handle `skip: false`, which is a meaningful value. But even if you fixed the bind to honor `skip: false`, `runJudge` still does:

```ts
if (options?.skip || judgeSkippedByEnv()) { ... skip ... }
```

So `AH_SKIP_JUDGE=1` always wins regardless of any per-call `skip: false`. Two options:

- Document explicitly that `AH_SKIP_JUDGE` is non-overridable (and pin it in `tests/unit/lib/testCases/judge.test.ts` — the existing test only checks the env path skips, not the precedence).
- Or make per-call `skip: false` an explicit "do not skip even if env says so" override (`if (options?.skip === true || (options?.skip !== false && judgeSkippedByEnv()))`).

Right now it's neither documented nor consistent.

### 7. The `evals/sdk-evaluator-id-smoke.eval.js` style mismatch

The file uses the v1 `async function ({ result, judge })` destructure even though every other eval got migrated. The codemod skipped it (no `prompt`, body doesn't really use `result`). The destructure pattern is now inconsistent with the rest of `evals/`. Either drop the unused `result` from the destructure or have the codemod handle no-prompt-but-also-not-using-result tests. Cosmetic but it'll trip the next person who reads it.

### 8. `tracesView` regression is real for body-only tests

For the no-prompt + deterministic body path, `loadedTraces` starts as `unavailableTracesAccessor("traces are only available after agent.run() has been called")`. Any read throws.

Pre-PR, the same path called `loadTracesAccessor(agentConfig, undefined)`, which (for `useTraces: false` agents) returned silent zeros. That's the #230 bug for `useTraces: true` — but for `useTraces: false`, the silent-zero behavior was correct. The new code throws unconditionally before `agent.run()` is called.

In the `evals/` directory none of the data-only tests read `traces`, so this doesn't break anything in-tree. But the `tests/unit/services/evaluationRunner.traces230.test.ts` Cases A–E now all `await agent.run()` first. There's no test for "body never calls `agent.run()`, reads `traces.totalTokens`, expects the specific 'available after agent.run()' message." Worth adding so the throw message is locked in as the contract — and worth a one-line note in the migration guide that data-only tests cannot read `traces`.

### 9. Dead code in `buildFixtures` helper after control inversion

Both runners' `buildFixtures` got a new `defaults` parameter:

```ts
function buildFixtures(
  result: EvalResult,
  traces: TracesAccessor = emptyTracesAccessor(),
  defaults?: { evaluatorId?: string; model?: string },
): TestFixtures {
  return {
    result,
    judge: bindJudge(defaults),
    ...
  };
}
```

But the new control-inversion path constructs fixtures inline (`fixtures = { ...before.fixtures, result: emptyResult, agent: agentFixture, ... judge: bindJudge({...}), ... }`) — `buildFixtures` is no longer called from there. It survives only for the legacy path that doesn't use `bindJudge`. Either delete the `defaults` parameter (it'll never be set) or remove `buildFixtures` entirely if no one calls it now. Pure cleanup, but it's confusing for the next reader.

---

## Test gaps that map to the bugs above

### 10. No test asserts OTel context propagation through `agent.run()`

This is what would have caught (1). A unit test would mock `context.with` and assert that `invokeAgent` is called with the eval span as the active context — straightforward to add. Without it, the runner-asymmetry is undetectable in CI. Recommend pinning this contract on both runners in a single shared test fixture (which would naturally pressure the engine extraction to land).

### 11. No test asserts `env` is forwarded through the connector's payload builder

Would have caught (2). Mock the connector's `buildPayload` to capture its argument, assert the synthesized `TestCase` (or `ConnectorRequest`) carries `env`. Currently nothing in `tests/unit` does this.

### 12. No test for two `.eval.{js,ts}` files colliding on a `defineEvaluator` id

Would surface (5). The collision is silent today; either lock the throw-on-duplicate contract or lock the last-write-wins contract — but pin one.

### 13. No test for a half-migrated body running through the codemod

The current codemod tests (`tests/unit/lib/testCases/codemod.test.ts`) cover full v1, no-prompt, no-result, idempotent. They don't cover the half-migrated case (body has both `result` and `agent.run()`), which is exactly the case (4) describes. Add a test that proves either: (a) the codemod migrates it correctly, or (b) it surfaces a "manual review needed" note instead of silently passing.

---

## Things I want to give explicit credit for

### The integration test for run-level evaluator binding is excellent

`tests/integration/cli/benchmarkEvaluatorId.integration.test.ts` is the kind of test I rarely see in this codebase: it spawns the *actual* `node cli/dist/index.js benchmark -f ... -e definitely-does-not-exist` subprocess, waits for the run to complete, fetches the persisted `matcherResults` via the HTTP API, and asserts the *exact* error string `Evaluator not found: definitely-does-not-exist` appears in the failed matcher's `errorMessage`. The "round-trip sanity check" at the end (`expect(FIXTURE_CONTENT).not.toContain('definitely-does-not-exist')`) is paranoid in the right way.

This is the test pattern that catches CLI/server asymmetry — exactly what #245's root cause was. More tests like this please. (And the `httpGet` workaround comment about Jest 30 + undici localhost race is clearly hard-won knowledge worth keeping.)

### The session interleave test really tests interleaving

`tests/unit/lib/matchers/session.test.ts`:

```ts
const sessionA = runInSession(async () => {
  recordVerdict(v('A1'));
  await tick();              // yield — B runs here under the old global model
  recordVerdict(v('A2'));
  ...
});
const sessionB = runInSession(async () => {
  await tick();
  recordVerdict(v('B1'));
  ...
});
```

`tick = setImmediate` is the right primitive for forcing an actual macrotask boundary between the two. Under the old module-global `activeSession`, this test would have produced `[A1, B1, A2, B2, A3]` in *one* of the two result arrays. The fact that it's split correctly is a real assertion about ALS, not just "the API doesn't throw". Good.

### Judge unit tests cover the right surface

`tests/unit/lib/testCases/judge.test.ts` has 16 cases including: per-call evaluatorId on body, runId forwarding present-vs-absent for `ResultLike` vs trajectory-array, legacy form, non-throwing failed verdict, `orThrow` chainable on pass, `judge.observe` records observe-role, errored verdict on HTTP 500, skip option doesn't fetch, env-skip doesn't fetch, cache hits identical inputs, no cross-pollution across distinct claims, bind injects evaluatorId, per-call wins over bound, bind-undefined returns identity, explicit per-call `undefined` falls through.

That's the matrix the contract needs. Missing only: the `bindJudge({ skip: false })` case (per #6) and the env-skip-non-overridable case.

### `services/evaluation/index.ts`'s `invokeAgent` extraction

Clean. Owns connector resolution, `beforeRequest`/`afterResponse` hooks, timing, auth merge — everything that should be shared between the legacy judge path and the new fixture path. `runEvaluationWithConnector` now reduces to: invoke + judge + report-shape glue. Worth keeping as a refactor even if every other phase were dropped.

### `parsePiJudgeJson` / `extractFromNdjson` extraction

The pi NDJSON handling (`pi --mode json` streams one event per line, not a single document) had been swallowed inside `evaluateWithPi`'s try/catch. Extracting `extractFromNdjson` and unit-testing it (`tests/unit/server/services/piJudgeService.extractNdjson.test.ts`) is the right move — and it's now the parsing path the agentic trace judge uses too. Bonus: `omitBasePack` is the right config knob for letting the trace judge skip the sample-agent extension instead of reaching for `process.env` hacks.

### `EvalRunsPage.tsx` merge is honest about being temporary

The `RunRow` doc comment is unusually explicit:

> NOTE (run-model convergence, RFC 004 single-engine): there are currently TWO disjoint run records … This is **deliberately temporary**, documented on `RunRow`: the long-term goal is to converge on the evaluation-run model …

The dedup-by-id is defensive (the namespaces are disjoint by construction), the `inspectPath` helper is correctly switched on `kind`, and the search/agent/time filters are mirrored across both branches. This is the right amount of compromise for a UI that needs to stay functional during a multi-PR migration.

---

## Recommendation

Once items (1)–(5) are addressed I'd be comfortable approving. Of those, (1) and (2) are one-line fixes with straightforward tests (10) and (11). (3) needs a design call (derive runId server-side vs. config-flag the provider). (4) is a tightening of an `if`-clause + one new codemod test. (5) is one line in `defineEvaluator` (`if (evaluatorRegistry.has(id) && evaluatorRegistry.get(id) !== fn) throw …`) plus a documented decision.

ashwin-pc:
Thanks for the deep re-review — this was exactly the right level of scrutiny. Addressed in `c7764aa` (with the earlier env/runId fixes in prior commits). Item-by-item:

## Real bugs

1. **OTel propagation asymmetry — fixed.** You were right that it was real, and it was actually *worse* than the framing: `executeEvaluationRun` created **no** `test_case` span at all (not just a missing wrap), so the evaluation-runs path silently relied on Strategy C. `evaluationRunner.ts` now starts a `test_case` span (synthesizing the benchmark shell the span helper needs) and wraps **both** the `agent.run()` invocation and the legacy `runEvaluationWithConnector` call in `context.with(caseSpanContext, …)`, mirroring `benchmarkRunner.ts`, and finalizes/ends the span on success and error. Pinned by a unit test that injects a sentinel eval-span context and asserts the invoke runs inside it (test gap **10**).

2. **`env` dropped — already wired + now tested.** Both runners already merge `options.env` into `invokeAgent`'s options (it lands in `connectorConfig.env`); added the runner-level assertion you asked for in (test gap **11**) so the wiring is locked.

3. **Cross-tenant `runId` — fixed via trajectory binding.** Went with "derive server-side": `/api/judge` now returns **403** when the requested `runId` isn't one the submitted trajectory carries (the SDK always derives `runId` from the judged result, so legitimate calls match). When the trajectory carries no `runId` we can't corroborate it, so that case is documented single-tenant-only. Tests cover the 403 and the matching happy path.

4. **Codemod guard — fixed.** Now skips **only** fully-migrated bodies (`agent.run()` + no `result` binding); a body that calls `agent.run()` but still destructures `result` gets a `review` note instead of "already migrated". New test for the half-migrated case (test gap **13**).

5. **`evaluatorRegistry` last-write-wins — fixed (throw + scope).** `defineEvaluator` now throws on a conflicting duplicate id (same id, different fn); same-fn re-register stays a no-op. The code-import loader (`sourceResolver.ts`) `clearEvaluators()` before each batch so prior-run evaluators don't leak and re-loads don't trip the guard. New collision test (test gap **12**).

## Design / contract

6. **`bindJudge({ skip: false })` + env precedence — fixed (tri-state).** `skip` is now `true` → skip, `false` → force-run **even if `AH_SKIP_JUDGE` is set**, omitted → defer to env. `bindJudge` compares `skip` against `undefined` so `{ skip: false }` binds instead of short-circuiting. Documented in `docs/SDK.md` and pinned by two tests.

7. **`sdk-evaluator-id-smoke.eval.js` style — resolved by the merge.** It now destructures `{ judge }` only (no unused `result`).

8. **`tracesView` body-only — locked the contract.** Added a test asserting a body that never calls `agent.run()` and reads `traces.totalTokens` throws `traces are only available after agent.run() has been called`, plus a note in the migration/availability docs that data-only tests cannot read `traces`.

9. **Dead `buildFixtures` — removed** from both runners (it had zero call sites after control inversion; the `defaults` param was never set).

Full unit suite: 4094 pass (+10 new), only the 4 pre-existing component suites fail (a local `node_modules` artifact — green on a clean CI install). Thanks again for catching (1) and (3) — both were genuinely shipping bugs.

goyamegh (approved):

Issue #264, opened 2026-06-08
Unify trace correlation strategies (A/B/C) so the agent (trace) judge actually sees the agent's spans


goyamegh:
## Summary

The agent (trace) judge (RFC 004 §4.4, server/services/piAgenticJudgeService.ts) and the run-detail Traces tab use **two different correlation strategies** to find the agent's OTel spans, and the strategies don't agree on which identifier to use. The result: when a user picks the agent trace judge from the UI Judge Model dropdown, the judge either fails (validation error) or runs against an empty span set, because the runId it scopes to (the connector's subprocess id) doesn't match what the agent actually emits.

This needs a unified correlation strategy across:
- The runner (which `runId` does it stamp on the report?)
- `/api/traces` and `/api/logs` (what scoping field do they index by?)
- `query_spans` / `query_logs` tools the agent (trace) judge uses
- The Traces tab (UI manual-correlation fallback)

Filing this as a follow-up so the evaluator-prompt-plumbing PR (split agent vs judge model, plumb saved evaluator's systemPrompt + scoringConfig.metrics through every provider) can land on its own merits. The agent trace judge surface is wired end-to-end in that PR but the resulting verdict is grounded in 0 spans because of this correlation gap.

## Reproduction

Concrete failed run on the eval-prompt-plumbing tunnel: `report-1780894835104-gvw5p4v1u` (Claude Code agent + AES Oncall evaluator + Agent Trace Judge model). The run document has:

```
agentKey:         claude-code         (useTraces: true)
modelId:          claude-opus-4.6
judgeModelId:     agent-trace-judge   ← user picked from UI dropdown
agent runId:      subprocess-1780894835961   ← THIS is the runner's runId
spans count:      0                   ← polled twice, found nothing under that runId
traceError:       runId is required for the agent (trace) judge provider — its trace tools scope to it
judgeDebug:       null                ← judge never executed
```

Two stacked bugs surface:

1. **`callBedrockJudge` doesn't forward `runId` to /api/judge.** Fixed in the evaluator-prompt-plumbing PR — small wire-up change in `services/evaluation/bedrockJudge.ts`.

[...]

Issue #190, opened 2026-05-11
Enhancement: Store traceId as top-level field on run documents


goyamegh:
## Summary

Currently, to correlate a run with its trace, we either:
1. Look up the `traceId` from already-stored spans (`report.spans[0].traceId`)
2. Fall back to querying by `gen_ai.request.id` (deprecated custom attribute)

This issue proposes storing `traceId` as a top-level field directly on the run document in OpenSearch for faster lookups.

## Motivation

- Eliminates parsing the spans array just to get the traceId
- Enables indexed lookups on the `traceId` field for faster queries
- Reduces query complexity (single-phase `term` query vs two-phase aggregation fallback)
- Aligns with OTel standard where traceId is the primary correlation key

## Proposed Changes

1. Add `traceId?: string` field to `TestCaseRun` type
2. In `tracePoller.ts`: persist traceId via `asyncRunStorage.updateReport(reportId, { traceId })` when traces are first found
3. In `asyncRunStorage.ts`: handle `traceId` in the `updateReport()` field mapping
4. In `RunDetailsContent.tsx` and CLI: check `report.traceId` before falling back to `report.spans[0].traceId`

## Context

This is Phase C of the trace correlation unification work:
- Phase A (done): Two-phase query fix — use `gen_ai.request.id` to find traceId, then fetch all spans
- Phase B (done): Prefer traceId from stored spans at read time, deprecate `gen_ai.request.id`
- **Phase C (this issue)**: Store traceId as first-class field for O(1) access

## Priority

Low — current Option A approach (extracting from stored spans) works correctly. This is a performance enhancement for when scale requires it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

krisfreedain:
@TackAdam @goyamegh @ps48 @jasonlhamazon - please take a look 
 
[Catch All Triage](https://github.com/opensearch-project/.github/blob/main/TRIAGING.md) - [1](https://github.com/krisfreedain) [2](https://github.com/andrross)

PR #265, opened 2026-06-09
fix(judge): honor saved evaluator prompt/metrics on every provider + separate agent vs judge model + surface complete judge output

goyamegh:
## Summary

Fixes a class of bugs where a **saved evaluator's `systemPrompt` and `scoringConfig.metrics` were silently dropped** on most judge providers, and the agent's model and the judge's model were conflated into a single field. Also surfaces the complete judge output in the UI and wires the agent (trace) judge to the agent's real OTel spans.

Two commits:
1. `fix(judge): plumb saved evaluator's systemPrompt + scoringConfig.metrics through every provider`
2. `feat(judge): separate agent vs judge model + agent-trace-judge correlation + complete judge output`

## Problem

- **Evaluator prompt drop:** `/api/judge` resolved the saved evaluator, but only the `bedrock` and `openai-compatible`/`litellm` providers actually used `evaluator.systemPrompt` + `scoringConfig.metrics`. The `claude-code`, `pi`, `agent`, and `agentic` providers fell back to a hardcoded `JUDGE_SYSTEM_PROMPT` and the legacy 4-metric schema. Editing a saved judge prompt produced byte-identical output on those providers.
- **Dropped judge fields:** any JSON the model emitted beyond `pass_fail_status`/`reasoning`/`metrics`/`improvement_strategies` (e.g. `failure_tags`, `weighted_score`, `confidence`) was discarded by the typed response shape.
- **Agent vs judge model conflation:** a single `modelId` was used for both agent invocation and judge selection, so picking a judge-only model (e.g. an agentic/trace judge) also handed it to the agent, which Bedrock rejected with "The provided model identifier is invalid".
- **Agent (trace) judge couldn't see the agent's spans:** its `query_spans`/`query_logs` tools scoped by `runId` only (Strategy B), which matched agent-health's own eval span but not subprocess-agent instrumentation (claude-code emits under its own correlation).

## What changed

**Evaluator prompt plumbing**
- `/api/judge` passes the resolved evaluator into all providers; each uses `evaluator.systemPrompt` (when non-empty) in place of the hardcoded baseline. Provider-specific tool-use addendums (agentic, agent-trace) are appended on top, not substituted.
- New shared `judgeResponseParser` — dynamic metric extraction from `evaluator.scoringConfig.metrics` (incl. rubric-style `scores` objects) and capture of every other emitted key into `JudgeResponse.extraFields`.
- New env-gated `judgeDebug` (AH_JUDGE_DEBUG) capturing the exact system prompt / user prompt / raw response, persisted on `LLMJudgeResponse`.

**Agent vs judge model split**
- New `judgeModelId` on `TestCaseRun`/`BenchmarkRun`/`EvaluationRun`, `/api/evaluate` + `/api/storage/evaluation-runs` bodies, the SDK request types, and CLI `--judge-model`.
- Resolution priority: `judgeModelId` > `BEDROCK_MODEL_ID` env > agent model (BC fallback). Agentic-provider judges ignore it and pick their own model.
- UI: "Agent Model" dropdown filtered to LLM providers (bedrock/openai-compatible/litellm); separate "Judge Model" dropdown with "Use evaluator default" sentinel.

**Agent (trace) judge correlation (tactical fixes toward #264)**
- `query_spans`/`query_logs` now union Strategy B (runId) with Strategy C (service.name + run wall-clock window) so the agent's real spans are found.
- Persist `runId` and a real W3C `traceId` (not the subprocess id) on the run document across create + placeholder-update + trace-mode polled-judge paths.
- Guard: a null/undefined `runId` no longer produces a `terms: [null]` clause that OpenSearch rejects wholesale (which was zeroing the run-detail Traces tab).

**UI**
- "LLM Judge" tab renamed "Judge Evaluation"; judge output rendered flat (identity, parsed metrics, every extraFields key, system/user prompt, raw response) so the complete output is visible upfront.

## Testing

- `build:all` clean; `npm audit --audit-level=high` reports 0 vulnerabilities.
- New unit suites: `judgeResponseParser.test.ts`, `judgeDebug.test.ts`, `evaluatorPromptPlumbing.test.ts`, `judgeModelIdResolution.test.ts`; updated `piAgenticJudgeService.test.ts` (trace-tool addendum invariant), routes/judge tests, constants/utils (model catalog).
- New integration suites: `judgeModelId.integration.test.ts` (round-trip on both endpoints + placeholder-update regression), `judgeModelIdFlag.integration.test.ts` (CLI flag end-to-end). 54/54 integration tests for the touched paths pass.
- New SDK evals: `judgeModelId-smoke.eval.js`, `judgeModelId-and-extra-fields.eval.js`.
- The 4 failing component suites (EvaluatorEditPage, RunDetailsContent, RunDetailsPage, SettingsPage) are pre-existing stale-mock failures that fail identically on the base commit — separately fixed in #263. This PR adds zero new failures.

## Related issues

- Refs #264 (unify trace correlation strategies — this PR lands the tactical Strategy-C wiring; the full `buildSpanQuery` refactor is the follow-up).
- Refs #190 (persist traceId on run documents — partially addressed; W3C traceId now stamped when available).


jasonlhamazon (approved):

goyamegh:
Addressed the review feedback in 10d92d1c:

- **`JudgeModelSelect` `string | undefined` (Copilot ×3):** reverted the public `value`/`onValueChange` to a plain `string`, using `''` as the "use evaluator default" sentinel (mapped to a Radix-safe non-empty internal sentinel, since `<SelectItem value="">` is disallowed). Existing required-`string` callers (benchmark dialogs) are unaffected; `TestCaseDetailPage` maps `'' ↔ undefined` at its own state boundary.
- **SSRF on discovery URLs (Code-Diff-Analyzer, medium ×2):** new `validateDiscoveryUrl` requires https + an allowlisted host before `ANTHROPIC_API_KEY` / `GITHUB_TOKEN` is forwarded (`api.anthropic.com`; `models.github.ai`/`api.githubcopilot.com`/`github.com`). localhost allowed for test proxies; `AH_ALLOW_CUSTOM_MODEL_ENDPOINTS=1` is the documented opt-out. Non-allowlisted host → 400, no credential sent.
- **`judgeDebug` default (low):** auto-on is now gated on `NODE_ENV === 'development'` (was `!== 'production'`) so CI/staging with `NODE_ENV` unset don't persist full prompts; explicit `AH_JUDGE_DEBUG=1` still works.

On the qodo "possible issue" re: `report.judgeModelId ??= run.judgeModelId` only forwarding `undefined` rather than the documented fallback chain — that's **by design**. The `evaluator.inferenceConfig.modelId → BEDROCK_MODEL_ID` fallback is applied at judge-*resolution* time (runner / `/api/judge`), not at *persistence* time. The stamp here is the audit record of what the customer selected; leaving it `undefined` when nothing was selected is correct (recording a synthesized id would be misleading).

PR #269, opened 2026-06-10
feat(config): support storage/observability in defineConfig() (TS config)

### LINKED ISSUE #261: Allow storage/observability cluster config in defineConfig() (single TS source of truth)
## Summary

`defineConfig()` (in `agent-health.config.ts`) cannot express **storage** or **observability** cluster config today. Those live only in a separate `agent-health.config.json` (read via `getStorageConfigFromFile()`) or `OPENSEARCH_STORAGE_*` / `OPENSEARCH_LOGS_*` env vars. This forces consumers to maintain **two config files** for one project, which diverges from the Playwright-style "single TS config is the source of truth" model the SDK is otherwise moving toward.

Proposal: add optional `storage` and `observability` fields to `UserConfig`, have the CLI-launched server consume them from the in-process loaded TS config, and encourage reading env-specific values from `process.env` inside the TS so a single committed `agent-health.config.ts` carries no environment-specific values or secrets.

## Why two files exist today

| File | Consumed by | Contains | Lifecycle |
|---|---|---|---|
| `agent-health.config.ts` | CLI/SDK process (esbuild/tsx) | **code** — connector instances, hooks, reporters, judge | hand-authored, loaded at startup |
| `agent-health.config.json` | long-running server + admin UI | **data** — storage/observability cluster endpoints, auth | read **and written back** at runtime |

Concrete blockers:
1. `lib/config/types.ts` `UserConfig` exposes `server`, `connectors`, `agents`, `models`, `testCases`, `reporters`, `judge`, `telemetry`, `remoteServers`, `codingAgentAnalytics`, `extends` — but **no `storage` / `observability`**. There is no TS code-path to author cluster config.
2. The JSON is machine-writable: the admin UI writes it back via `server/routes/storage/admin.ts`. Config that the UI mutates at runtime needs a serializable target (you can't safely round-trip into hand-authored TS).
3. Cluster config is resolved in `server/middleware/dataSourceConfig.ts` with priority `agent-health.config.json` → `OPENSEARCH_STORAGE_*` env → null (file-based fallback). The loaded TS config is never consulted.

When the CLI starts the server from the consumer cwd, it already loads the user's TS in-process (`server/app.ts` registers `config.connectors[]`), so the server *does* have access to the TS config object — it just doesn't read storage/observability from it.

## How other tools handle this

The JS tooling ecosystem converged on **one code-config file + `process.env`/`.env`**, no second JSON:

- **Playwright** — single `playwright.config.ts` is the sole source of truth (projects, reporters, `webServer`, `use`), loaded by its own TS loader; env-specific/secret values come from `process.env` (dotenv). No stateful server+UI rewrites config, so no writable JSON is needed.
- **Vite** (`vite.config.ts` + `loadEnv`), **Jest** (`jest.config.ts`), **ESLint flat config**, **Next.js**, **Drizzle** — all single `*.config.{ts,js}` + env.

Convention: *authored config = code (TS); environment-specific/secret values = `process.env`/`.env`.* agent-health's second JSON exists only because of the runtime-mutable server + admin UI.

## Proposed change

1. **Extend `UserConfig`** (`lib/config/types.ts`) with optional `storage?: StorageClusterConfig` and `observability?: ObservabilityClusterConfig` (reuse the existing types from `server/middleware/dataSourceConfig.ts` / config service).
2. **Server consumes the loaded TS config.** When the server is launched by the CLI with an in-process resolved config, feed `config.storage` / `config.observability` into the storage-resolution chain.
3. **Resolution precedence** (make explicit and documented):
   `agent-health.config.json` (UI-written, runtime) → `agent-health.config.ts` (`defineConfig`) → `OPENSEARCH_STORAGE_*` / `OPENSEARCH_LOGS_*` env → file-based fallback.
   Keep the JSON highest so the admin UI's runtime edits still win; the TS becomes the committed default.
4. **Encourage env-driven TS** in docs/examples, e.g.:
   ```ts
   export default defineConfig({
     connectors: [prioritizerConnector],
     storage: {
       endpoint: process.env.OPENSEARCH_STORAGE_ENDPOINT!,
       authType: 'sigv4',
       awsRegion: 'us-east-1',
       awsService: 'es',
       awsProfile: process.env.AWS_PROFILE,
     },
   });
   ```
   so the committed `.ts` carries no environment-specific values or secrets.

## Outcome

- Single authored config file (`agent-health.config.ts`) for projects that don't mutate storage via the admin UI — matching the Playwright model the SDK is converging on.
- `agent-health.config.json` becomes optional (only present when the admin UI writes runtime config).
- No behavior change for existing JSON/env users (precedence preserves their setup).

## Notes

- Surfaced while migrating a consumer project to the code SDK: a shared-cluster, private-repo project ended up committing `agent-health.config.json` purely to share cluster config, which is awkward when `agent-health.config.ts` already exists as the authored source of truth.
- Related to the Test SDK v2 direction (RFC 004 / #256).

uhinze:
## Description

Adds optional `storage` and `observability` cluster config to `defineConfig()` (`agent-health.config.ts`), so a single committed TS config can be the source of truth instead of also maintaining `agent-health.config.json`. Implements the proposal in #261.

Today `defineConfig()` cannot express storage/observability — those come only from `agent-health.config.json` or `OPENSEARCH_STORAGE_*` / `OPENSEARCH_LOGS_*` env vars. A project that already has an `agent-health.config.ts` ends up committing a second JSON file purely to share cluster config.

## Changes

- `lib/config/types.ts`: add optional `storage?: StorageClusterConfig` and `observability?: ObservabilityClusterConfig` to `UserConfig` (and `ResolvedConfig`).
- `lib/config/loader.ts`: carry these through `mergeConfigs`, and publish the TS-authored cluster config in-process via `getStorageConfigFromCode()` / `getObservabilityConfigFromCode()` (the server resolver has no access to the loaded config object). Cleared in `clearConfigCache()`.
- `server/middleware/dataSourceConfig.ts`: consult the code config between the JSON and env layers in `resolveStorageConfig` / `resolveObservabilityConfig`.
- `server/services/configService.ts`: `getConfigStatus()` gains a `'code'` source so the UI/CLI report TS-configured clusters accurately instead of "not configured".
- `docs/CONFIGURATION.md`: document the fields, the `process.env` guidance, and the precedence.

## Resolution precedence (explicit)

`agent-health.config.json` (UI-written, runtime) → `agent-health.config.ts` (`defineConfig`) → `OPENSEARCH_STORAGE_*` / `OPENSEARCH_LOGS_*` env vars.

The JSON stays highest so the admin UI's runtime edits keep winning; the TS config becomes the committed default for projects that don't mutate storage via the UI. **No behavior change for existing JSON/env users.**

## Example

```ts
export default defineConfig({
  storage: {
    endpoint: process.env.OPENSEARCH_STORAGE_ENDPOINT!,
    authType: 'sigv4',
    awsRegion: 'us-east-1',
    awsService: 'es',
  },
  observability: {
    endpoint: process.env.OPENSEARCH_LOGS_ENDPOINT!,
    authType: 'sigv4',
    awsRegion: 'us-east-1',
    awsService: 'es',
    indexes: { traces: 'otel-v1-apm-span-*' },
  },
});
```

## Testing

- `npx tsc --noEmit` clean.
- Added unit tests in `tests/unit/lib/config/loader.test.ts` (code accessors default to null; `defineConfig` preserves `storage`/`observability`; `clearConfigCache` resets the published config).
- Existing config/service suites pass (`tests/unit/lib/config`, `tests/unit/lib/dataSourceConfig.test.ts`, `tests/unit/server/services`): 657 passed, 0 failed.

## Notes

- Surfaced while migrating a real deployment (Amazon OpenSearch Service domain with IAM FGAC) to the code SDK — the `.ts` `observability` block was silently ignored and the UI reported "Observability data source not configured".
- Related to the Test SDK v2 direction (#256 / RFC 004); happy to reshape if this should land within that redesign.

Closes #261

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -214,13 +214,13 @@ export default defineConfig({
 });
 ```
 
-> **Storage/observability precedence.** These are resolved in the order
+> **Storage/observability precedence.** A cluster is resolved in the order
 > `agent-health.config.json` (written by the Settings UI at runtime) →
 > `agent-health.config.ts` (`defineConfig`, shown above) →
-> `OPENSEARCH_STORAGE_*` / `OPENSEARCH_LOGS_*` env vars → file-based fallback.
-> The JSON stays highest so the admin UI's runtime edits keep winning; the TS
-> config is the committed default for projects that don't mutate storage via the
-> UI. Keep secrets/endpoints in `process.env` so the committed `.ts` carries none.
+> `OPENSEARCH_STORAGE_*` / `OPENSEARCH_LOGS_*` env vars. The JSON stays highest
+> so the admin UI's runtime edits keep winning; the TS config is the committed
+> default for projects that don't mutate storage via the UI. Keep
+> secrets/endpoints in `process.env` so the committed `.ts` carries none.
 
 ### Config File Options
 
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -191,17 +191,36 @@ export default defineConfig({
     },
   ],
 
-  // Override storage (can also use env vars)
+  // Storage cluster (evaluation data). Read environment-specific values and
+  // secrets from process.env so the committed file carries none.
   storage: {
     endpoint: process.env.OPENSEARCH_STORAGE_ENDPOINT,
-    username: 'admin',
-    password: process.env.OPENSEARCH_STORAGE_PASSWORD,
+    authType: 'sigv4',          // 'none' | 'basic' | 'sigv4'
+    awsRegion: 'us-east-1',
+    awsService: 'es',           // 'es' (managed) | 'aoss' (Serverless)
+  },
+
+  // Observability cluster (agent traces/logs/metrics).
+  observability: {
+    endpoint: process.env.OPENSEARCH_LOGS_ENDPOINT,
+    authType: 'sigv4',
+    awsRegion: 'us-east-1',
+    awsService: 'es',
+    indexes: { traces: 'otel-v1-apm-span-*' },
   },
 
   // Custom test cases location
   testCases: './my-tests/*.yaml',
 });
 ```
+
+> **Storage/observability precedence.** These are resolved in the order
+> `agent-health.config.json` (written by the Settings UI at runtime) →
+> `agent-health.config.ts` (`defineConfig`, shown above) →
+> `OPENSEARCH_STORAGE_*` / `OPENSEARCH_LOGS_*` env vars → file-based fallback.
+> The JSON stays highest so the admin UI's runtime edits keep winning; the TS
+> config is the committed default for projects that don't mutate storage via the
+> UI. Keep secrets/endpoints in `process.env` so the committed `.ts` carries none.
 
 ### Config File Options
 
--- a/lib/config/index.ts
+++ b/lib/config/index.ts
@@ -32,5 +32,7 @@ export {
   findConfigFile,
   getConfigFileInfo,
   clearConfigCache,
+  getStorageConfigFromCode,
+  getObservabilityConfigFromCode,
   DEFAULT_SERVER_CONFIG,
 } from './loader';
--- a/lib/config/loader.ts
+++ b/lib/config/loader.ts
@@ -26,6 +26,7 @@ import type {
   JudgeConfig,
   TelemetryConfig,
 } from './types';
+import type { StorageClusterConfig, ObservabilityClusterConfig } from '@/types/index.js';
 
 /**
  * Default server configuration
@@ -193,6 +194,10 @@ function mergeConfigs(
     reporters,
     judge,
     telemetry,
+    // Cluster config authored in defineConfig() (optional). Consumed by the
+    // server's data-source resolution chain below the UI-written JSON.
+    ...(userConfig.storage ? { storage: userConfig.storage } : {}),
+    ...(userConfig.observability ? { observability: userConfig.observability } : {}),
   };
 }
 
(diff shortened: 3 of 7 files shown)


uhinze:
Thanks for the automated review. Addressed in e582ad7:

**Storage drift false-positive (the substantive one):** fixed. `getConfigStatus()` no longer folds `codeStorage` into `resolvedStorageConfig` when computing the drift key. Drift compares the runtime storage backend against the config the storage adapter actually initializes from (file/env), so including `defineConfig()` storage there could flag drift when the runtime state is correct. Code config still participates in resolution and in the `configured`/`source` reporting.

On the two low-priority suggestions, I left them as-is intentionally:

- **`as StorageClusterConfig` on the `{ endpoint }` partial:** this is pre-existing code I didn't introduce, and it's only used to derive a cache key via `configToCacheKey`, which reads `endpoint` (+ optional auth fields). Tightening it is reasonable but out of scope for this change and would be a behavior-neutral refactor better done separately.
- **Module-level mutable state for the published cluster config:** this mirrors the existing `cachedConfig` / `cachedConfigPath` pattern in the same file and is cleared in `clearConfigCache()`. Node is single-threaded and the loader is a singleton keyed by `cachedConfigPath`, so a `Map`-by-path adds complexity without a correctness gain here.

Happy to take either of those if you'd prefer them in this PR.

uhinze:
Closing this — taking a different approach. Thanks for the reviews.

uhinze:
Agent went nuts, this is not a good PR. Config loading probably needs refactor

uhinze:
Addresses #261 by the way

Issue #96, opened 2026-03-09
Getting Started flow for Traces


goyamegh:
**Parent:** #88 — [META] Developer Experience & Adoption

## Description

Guide new users on how to connect tracing to their agents and view traces in Agent Health.

## Requirements

- Instrumentation guide for existing agents (OpenTelemetry GenAI semantic conventions)
- Step-by-step for new users who haven't instrumented yet
- In-app guidance when traces page is empty (zero-state)
- Link to sample trace data for exploration without a live agent
- Cover common agent frameworks: LangGraph, CrewAI, custom REST agents

## References

- [Backlog doc](https://quip-amazon.com/aeVaAkAz2Viq)

sibasispadhi:
Thanks @goyamegh. Here's the RFC for standalone mode.

I'll focus on items #1 and #2 (standalone mode + file-based storage).
Items #3 (UI demo banner) and #4 (doctor command for traces) could follow separately.

---

### Proposed approach

Two new files, minimal changes to existing ones:

**1. `server/adapters/file/ObservabilityModule.ts`** — implements `IObservabilityModule`
using JSON files on disk (same `agent-health-data/` directory pattern as
`FileStorageModule`). Spans stored as individual JSON files, with a lightweight
index file for fast filtering. The `NormalizedSpan` shape stays identical to
`tracesService.ts` output; no UI changes needed.

**2. `server/services/otlpReceiver.ts`** — a lightweight OTLP HTTP/JSON receiver
(`POST /v1/traces` on a dedicated configurable port, defaulting to `:4318`) that
decodes incoming spans and writes them to `FileObservabilityModule`. Running on its
own port means the path follows the OTLP HTTP standard (`/v1/traces`), keeping it
a drop-in replacement for any existing OTLP HTTP exporter config. This bypasses the
OTEL Collector + Data Prepper pipeline entirely, which is intentional for the
standalone use case.

**3. CLI flag** — `--standalone` on the default command and `serve` subcommand.
When set: boot with `FileObservabilityModule`, start the OTLP receiver, and show
a "Standalone Mode" indicator in the Settings UI. No config file or env vars needed.

The result:
```bash
npx @opensearch-project/agent-health --standalone
# ✓ UI at http://localhost:4001
# ✓ OTLP HTTP receiver at http://localhost:4318/v1/traces
# ✓ Traces stored to agent-health-data/traces/
For Docker users, nothing changes — the existing Compose stack and OpenSearch path are untouched.

[...]

PR #301, opened 2026-06-18
feat(traces): local filesystem trace backend + pluggable observability adapter (#96)

goyamegh:
## What & why

Closes the local-evaluation half of #96: **traces now work with no OpenSearch observability cluster.** When none is configured, agent-health runs its own mini-collector and stores spans on disk, so the Traces page and the run-report Traces tab work out of the box — the file-backed parallel to how eval *runs* already fall back to `agent-health-data/`.

Built as a **pluggable data-source adapter** so filesystem and OpenSearch are interchangeable backends, with room for more later (Jaeger/Tempo/ClickHouse/…). This realizes the `IObservabilityModule` interface the repo already declared but never implemented (the storage plane already had file/opensearch adapters; observability was the one plane still hard-wired to OpenSearch).

## How it works

- **`POST /v1/traces`** — embedded OTLP/HTTP **JSON** receiver (the format connectors already export). Parses `ResourceSpans` → internal `Span[]` → `TraceStore`.
- **`TraceStore`** — `agent-health-data/traces/{traceId}.json`, atomic writes, merge/dedupe by spanId, **kept forever** (customers delete manually; graduate to OpenSearch when local storage grows).
- **`FileObservabilityModule`** — realizes `IObservabilityModule`, replicating the OpenSearch correlation semantics in-memory: A (`traceId`) / B (`gen_ai.request.id`=runId) / C (`service.name`+time window), plus `session.id` / time / serviceName / textSearch, union via `minimum_should_match` + offset pagination.
- **`OpenSearchObservabilityModule`** — the existing read path (`fetchTraces`/`fetchLogs`/`checkTracesHealth`) behind the same interface (PR A, pure refactor — zero behavior change).
- **`getObservabilityModule(req)`** — file backend is the **zero-config default** when no cluster is configured; `/api/traces` + `/api/traces/health` now report `backend: 'opensearch' | 'file'`. Demo samples remain the empty-store browse fallback.

## Precedence safety (load-bearing invariant)

> If an OpenSearch observability cluster is configured, it is the **only** trace source.

- Backend is chosen by **config presence at startup, not health**.
- A configured-but-broken cluster surfaces an **error** (like the storage backend / #281) — **never** a silent fallback to disk.
- The receiver **drops** payloads while a cluster is configured (no split-brain between disk and cluster).

## Tests

- Unit: `otlpTransform` (mapping / hex+base64 ids / status / events), `TraceStore` (write/merge/readAll/retention/env-dir), `FileObservabilityModule` (A/B/C/union/pagination), `OpenSearchObservabilityModule` (delegation/health).
- Route: `/api/traces` + `/api/traces/health` updated for file-default + `backend` field.
- Integration (supertest, real stack): OTLP ingest → file → read round-trip, runId correlation, file-backend health, and the **cluster-configured drop** (no split-brain).
- `tsc` ✓ · `npm run build` (vite) ✓ · `build:server` ✓ · 1737 server/traces tests pass.

## Notes / follow-ups

- JSON-first receiver; protobuf can be layered on later.
- Docs: `docs/CONFIGURATION.md` gains a "Where things live" section (`.agent-health/` settings vs `agent-health-data/` data vs `evals/` source).
- Next (separate PR): auto-point local connectors' `OTEL_EXPORTER_OTLP_ENDPOINT` at `/v1/traces` in file mode so eval runs capture traces with zero setup.

Refs #96


goyamegh:
Thanks for the review — addressed the actionable items in `5cf4fc66`:

- **`doctor.ts` regression (good catch):** `/api/traces/health` returning `status:'ok'` in file mode made the CLI doctor print "OpenSearch traces index is accessible". It now reads the `backend` field and reports *"Local filesystem trace storage (agent-health-data/traces)"* for `backend:'file'` (still handles the older `sample_only` shape).
- **`normalizeId`:** now returns `''` for non-hex / non-round-trippable base64 instead of leaking the raw string; spans with a missing/invalid `traceId` **or** `spanId` are skipped (an empty `spanId` would collide in the store), and `SpanEvent.name` defaults to `''`.
- **`TraceStore`:** unusual/overlong trace ids are now **sha256-hashed** for bounded filenames (was hex-encoding the whole id); `writeSpans` skips spans without a `spanId` and the existing-span merge guards the key.
- **`FileObservabilityModule`:** time-range filters use `!== undefined`, so a valid `startTime`/`endTime` of `0` (epoch) is honored rather than treated as absent.
- **CHANGELOG:** reconciled the adapter-seam entry with the combined PR (dropped the stale "returns null / later PR / no behavior change" wording).
- **Tests:** added coverage for invalid-id skipping, epoch-`0` bounds, bounded-filename hashing, and `spanId`-less write skips.

Not changed:
- **Mend Security Check (red):** no dependencies were added/changed in this PR, so this is a pre-existing/transitive org finding rather than something this diff introduces.
- **Deterministic trace-id hashing → existence leak:** low severity here — the trace files are local, gitignored runtime state under `agent-health-data/`, not a multi-tenant surface. Happy to switch to a salted/HMAC scheme if you'd prefer.

TackAdam (approved):

PR #303, opened 2026-06-18
feat(traces): context-window utilization (Ctx%) in trace summary

goyamegh:
## What & why

Adds a **Ctx%** signal to the trace summary: how close the busiest LLM request in a trace came to filling the model's context window. This is a standard column teams reach for when comparing agents/workflows on an identical task (the same task can vary 2-3x in context pressure depending on prompt/context strategy), and we already collect everything needed to compute it from OTel `gen_ai.usage.*` attributes.

## Correctness: peak, not sum

The load-bearing decision here is that Ctx% is computed from **`peakInputTokens`** — the largest single-request input-token count in the trace — **not** the summed `inputTokens` across the whole trace. Each agent turn re-sends the running context, so summing input tokens routinely exceeds the window by many multiples and is meaningless as a utilization figure. The peak request is the real high-water mark against the window.

## Changes

- **`services/traces/traceSummary.ts`** — `TraceSummary` gains `peakInputTokens` (max single-span input tokens). Existing `inputTokens`/`outputTokens`/`totalTokens` sums are unchanged.
- **`lib/contextUtilization.ts`** (new, pure) —
  - `resolveContextWindow(model)` resolves a model's `context_window` from the existing model registry by **either** registry key (`claude-opus-4.6`) **or** provider `model_id` (`us.anthropic.claude-opus-4-6-v1`), case-insensitive — because traces report the `model_id` while the rest of the app uses the short key.
  - `contextUtilizationPct(peakInputTokens, model)` — percentage, `undefined` for unknown model / missing tokens, **not clamped** (>100 is a real over-budget signal).
  - `formatContextPct(pct)` — one decimal under 10%, whole numbers above.
- **`components/traces/TraceSummaryStrip.tsx`** — renders a `Ctx%` chip next to the token count. Omitted when the model is unknown (no misleading 0/100%); rendered red with an over-budget tooltip when utilization ≥ 100%.

## Tests

- **Unit** — `tests/unit/lib/contextUtilization.test.ts` (resolve by key + model_id, case/whitespace, non-Claude window, unknowns; pct compute, >100, undefined paths; formatting). `tests/unit/services/traces/traceSummary.test.ts` gains a `peakInputTokens` assertion proving it is the MAX (100) not the sum (150). 21/21 pass.
- `npx tsc --noEmit` ✓ · `npm run build` (vite) ✓

## Notes / follow-ups

The chip reads off the trace summary strip (Agent Traces inline + fullscreen headers). A natural follow-up is a sortable `Ctx%` column in the Benchmark Runs overview so it sits alongside time / tool-calls for cross-agent comparison.


TackAdam (approved):

goyamegh:
## Screenshots — context-window utilization (Ctx%)

Captured from a local build with a synthetic OTLP trace POSTed to the file trace backend (one agent root + two `chat` LLM spans on `claude-opus-4-6`, peak input **150k** tokens, plus a smaller 42k call).

**1. Agent Traces list** — the seeded trace

Agent Traces list

**2. Expanded trace — the new `Ctx%` chip in the summary strip**

Expanded trace with Ctx% chip

**3. Zoom of the summary strip**

Ctx% chip zoom

The strip reads `2 LLM · 1 agent · 203k tok · 75% ctx · us.anthropic.claude-opus-4-6-v1`. Note **75%** = peak request (150k) ÷ 200k window — **not** the 192k input *sum* (which would read 96%). That's the peak-not-sum correctness this PR is built on.

<sub>Images hosted on a stable assets branch (SHA-pinned) so the links survive merge/squash. PNGs are not added to this PR's diff.</sub>

PR #311, opened 2026-06-18
fix(pkg): resolve SDK from a .ts config/eval/workflow loaded as CJS (add require+default exports)

goyamegh:
## What

Adds `require` + `default` conditions to `exports["."]` in `package.json` (it had only `import` + `types`).

## Why

A `.ts` `agent-health.config.ts` / `*.eval.ts` / `*.workflow.ts` loaded via `tsx` is compiled to **CJS**, so its `import … from @opensearch-project/agent-health` resolves through the **`require`** condition. With only `import`/`types` defined, Node throws:

```
No "exports" main defined in …/@opensearch-project/agent-health/package.json
```

…so the SDK cannot be loaded from a TypeScript config/eval/workflow at all (hit while running an eval *and* a workflow from a `.ts`-configured project). Pointing `require`/`default` at the same ESM entry (Node ≥ 22 supports `require(esm)`) makes both module systems resolve. ESM consumers are unaffected.

## Test
- `require(@opensearch-project/agent-health)` and `import` both resolve.
- No change for existing ESM consumers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Inspired by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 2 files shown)


goyamegh:
## Verification ✅

The bug this fixes: a `.ts` config/eval/workflow loaded via `tsx` (CJS) could not resolve the SDK — Node threw `No "exports" main defined …`. After adding `require`+`default` to `exports["."]`:

```
$ node -e "const m=require(@opensearch-project/agent-health); console.log([source,workflow,test,defineConfig].filter(k=>k in m).join(, ))"
require ok — exports: source, workflow, test, defineConfig
```

`require()` (CJS) and `import` (ESM) both resolve now; ESM consumers unchanged.

TackAdam (approved):

Issue #271, opened 2026-06-12
Config v2: single authored .ts + .agent-health/ runtime state (user+project scoped), JSON ignored when .ts present


goyamegh:
## Summary

Make `agent-health.config.ts` the **single file a human authors**, and demote today's `agent-health.config.json` to **runtime state the app owns** — renamed to `.agent-health/state.json`, gitignored, undocumented as "config", and **ignored entirely when an authored config file is present**. Add **user + project scoping** to `.agent-health/` (like git/npm config).

This is the complete design; it ships in **two PRs** (see "Sequencing").

Builds on #261 (which made `storage`/`observability` work in `.ts`) and #270 (docs accuracy).

## The model: two modes, picked by presence of an authored config file

| Mode | Trigger | Source of truth | Settings UI (data sources) |
|---|---|---|---|
| **Code-first** | `agent-health.config.{ts,js,mjs}` exists (project **or** user) | `.ts` + `.env` | read-only (test-connection + copy-snippet) |
| **UI-first** | no authored config file anywhere | `.agent-health/state.json` (+ `.env`) | writable (persists to state) |

Single rule (supersedes #261's `JSON > TS > env`): **authored config present ⇒ state file ignored.** No merge, no drift, no shadowing.

## File layout & scoping (user + project)

```
~/.agent-health/                 # USER scope
  agent-health.config.ts         #   optional user-global authored config
  state.json                     #   user-global runtime state (UI-written)
<cwd>/.agent-health/             # PROJECT scope
  state.json                     #   project runtime state (UI-written)
<cwd>/agent-health.config.ts     # project authored config
```

Resolution (each tier overrides the one below):
- **Code-first:** project `.ts` -> user `~/.agent-health/*.ts` (existing `extends`) -> `.env`.
- **UI-first:** project `.agent-health/state.json` -> user `~/.agent-health/state.json` -> `OPENSEARCH_*` env -> file-storage fallback.

State keys merge per top-level key (`storage`, `observability`, `customAgents`, `debug`, `remoteServers`), project over user. UI/CLI writes target **project** scope by default (`--global` writes user).

## Phase 1 — core (no UI change required)
- New `lib/config/statePaths.ts`: path + mode resolver (`projectStatePath`, `userStatePath`, `hasAuthoredConfig`, `isCodeFirstMode`, `readLayeredState`, `writeState`).
[...]

PR #314, opened 2026-06-18
feat(config): combine agent-health-data/ into .agent-health/data/ (config-v2 follow-up)

goyamegh:
## What

Moves the generated-data dir `agent-health-data/` **under** the app-managed `.agent-health/` as `.agent-health/data/`, so config-v2 `state.json` and all generated outputs (runs, benchmarks, test-cases, analytics, evaluators, profiles, skill-evals, uploaded-skills) live in **one** gitignored, app-managed directory.

## Why

After config-v2 (#271/#280), `.agent-health/` is already *the* app-managed dir (`state.json`). `agent-health-data/` is a sibling holding the same *kind* of thing — generated, disposable, gitignored. Two ignore entries + two "where does my stuff live?" answers collapse to one: `.agent-health/` = `state.json` + `data/`.

## How

- Single owner in `lib/config/statePaths.ts`: `projectDataDir()` -> `<cwd>/.agent-health/data` (+ `legacyProjectDataDir()`).
- Repoint every call site: file storage backend (`StorageModule`), `profile.ts`, `skills.ts` (x4, incl. backward-compat for the legacy path prefix), `skill.ts` log.
- **One-time, idempotent migration** in `configMigration.ts`: if `<cwd>/agent-health-data/` exists and `<cwd>/.agent-health/data/` does not, move it once — existing run history is preserved, no-op thereafter.
- `.gitignore` collapses to `.agent-health/`; `docs/CONFIGURATION.md` updated.

## Test

- New unit tests for `projectDataDir`/`legacyProjectDataDir`; config + `StorageModule` suites green.

Follow-up to config-v2 (#280).


goyamegh:
## Verification ✅

- `projectDataDir()` resolves generated data under the app-managed dir; legacy path preserved as the migration source:
```
projectDataDir(/tmp/proj)       -> /tmp/proj/.agent-health/data
legacyProjectDataDir(/tmp/proj) -> /tmp/proj/agent-health-data
```
- Unit tests green: config suite **21/21** (incl. new `projectDataDir`/`legacyProjectDataDir` tests) + `StorageModule` **9/9**; `build:all` green.
- One-time migration is idempotent (moves an existing `agent-health-data/` once, no-op thereafter).

TackAdam (approved):

goyamegh:
## Updated: removed the migration shim; completed the move

Since the `agent-health-data/` layout was **never released**, this now **fully removes** it instead of migrating from it:

- Dropped the one-time `migrateLegacyDataDir()` shim + `legacyProjectDataDir()` / `LEGACY_DATA_DIRNAME` helpers — no migration needed for an unreleased format.
- The first cut only moved storage/profile/skill outputs; it left the **file trace store** (`TraceStore` / `FileObservabilityModule` / `otlpReceiver`) and the **skill-eval workspace prefix in both the API and the frontend** (`SkillsPage.tsx`) still writing to `agent-health-data/` — which would have *fought* the boot migration (it moves the dir, then the trace store recreates it). Both now point at `.agent-health/data/`.
- `AGENT_HEALTH_DATA_DIR` kept as a test override (default is now `.agent-health/data`); `doctor` / `/api/traces` status messages + docs updated.

Net: one app-managed, gitignored dir owned by `projectDataDir()` — no legacy paths anywhere.

Verified locally: `npm run build:all` green, `tsc --noEmit` clean, 28 unit + 4 OTLP file-trace integration tests pass.

TackAdam (approved):

jasonlhamazon (approved):

PR #315, opened 2026-06-18
fix(cli): keep stdout clean for --output json / pipes (startup logs to stderr, quiet dotenv)

goyamegh:
## What

Route the `[Connectors] … registered` startup logs to **stderr** (`console.error`) and pass `quiet: true` to dotenv, so the CLI's **stdout** carries only command output.

## Why

`agent-health list <resource> --output json` (and any piped/scripted use) emitted diagnostics on **stdout** before the JSON:

```
[Connectors] Browser-safe connectors registered: …
[Connectors] Server connectors registered: …
[dotenv@…] injecting env (0) from .env -- tip: …
```

…so `… --output json | jq` / `python -m json.tool` fails to parse. Diagnostics belong on stderr; data on stdout.

## Test

`agent-health list commands --output json` now parses cleanly; the three lines appear on **stderr** only.


goyamegh:
## Verification ✅

`list <resource> --output json` now emits parseable JSON (was corrupted by `[Connectors]`/`[dotenv]` lines on **stdout**):
```
$ agent-health list commands --output json 2>/dev/null | python3 -m json.tool   # valid JSON — 18 commands
$ agent-health list commands --output json 2>/tmp/err >/dev/null                 # the diagnostic lines now on stderr
```

TackAdam (approved):

PR #316, opened 2026-06-19
fix(traces): time-range picker + real service-name filter on Live Traces

goyamegh:
## What

Two fixes to the **Live Traces** page (`components/traces/TracesPage.tsx`):
1. **Time-range picker** — `5m / 15m / 1h / 6h / 24h / 7d` (default **1h**), replacing the hard-coded last-5-minutes window.
2. **Service filter derived from real `service.name`s** — built from the OTel `service.name`s actually present in fetched spans (accumulated across fetches), replacing the dropdown built from agent *display names*.

## Why

The page only ever queried the **last 5 minutes**, and its Service dropdown used agent display names (e.g. `"Claude Code"`) that never match the emitted `service.name` (e.g. `claude-code-agent`). Result: a quiet/empty view (recent or ingestion-lagged spans fall outside 5 min), and selecting an agent returned **zero** rows. After the fix, widening the window surfaces history and the Service filter actually matches (e.g. `claude-code-agent`, `agent-health`).

## Test

Frontend `build` green; widening the range surfaces spans and the Service dropdown lists the real services present.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Inspired by [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Fixed
+- **Live Traces: time-range picker + real service-name filter** ([components/traces/TracesPage.tsx](components/traces/TracesPage.tsx)): the Live Traces page was a fixed last-5-minutes window, and its Service filter was built from agent *display names* (which never match the OTel `service.name`) — so recent/lagged spans were invisible and picking an agent returned nothing. Adds a time-range selector (5m / 15m / 1h / 6h / 24h / 7d, default 1h) and derives the Service dropdown from the `service.name`s actually present in fetched spans (accumulated across fetches), so e.g. `claude-code-agent` appears and filtering works.
+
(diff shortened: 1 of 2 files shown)


goyamegh:
## Verification ✅

Frontend `build` green. Behavior:
- **Time-range selector** (5m / 15m / 1h / 6h / 24h / 7d, default **1h**) replaces the fixed 5-minute window.
- **Service filter** is now derived from the OTel `service.name`s actually present in fetched spans (accumulated across fetches), so it matches real services (e.g. `claude-code-agent`) instead of agent display names.

_A live screenshot of the Live Traces page would show real internal trace spans, so per policy it is shared privately rather than on this public PR._

TackAdam (approved):

goyamegh:
Closing: this targeted `/traces` ("Live Traces"), which is an **orphaned route** (no nav link). The page users actually use is `/agent-traces` (`AgentTracesPage`), which already has a time-range picker + agent->service mapping. The real fix (map the `claude-code` agent to its emitted OTel `service.name` so the filter works, and remove the dead `/traces` page) is in a new PR. Superseded.

PR #322, opened 2026-06-19
fix(storage): no silent file fallback in error state + validate authType

goyamegh:
## Summary

Two storage-config **robustness** fixes surfaced while end-to-end testing #314's file↔OpenSearch precedence. Both are about failing **loudly and honestly** instead of silently degrading.

### 1. No silent file fallback when a configured cluster is unreachable
When an OpenSearch storage cluster is configured but unreachable, the runtime state is `backend: 'error'` — but the module singleton stayed the default `FileStorageModule`. So `/api/storage/*` entity CRUD **silently read/wrote local files** while `/api/server-info` reported `error`: a split-brain that violates the documented invariant *"a configured cluster is always authoritative; a broken cluster surfaces an error, never a silent fallback to disk."*

**Fix** (`server/routes/storage/index.ts`): a guard returns **503** for entity CRUD (test-cases / benchmarks / runs / evaluators / analytics / reports) while storage is in the error state. It is mounted **after `adminRoutes`**, so the config/recovery endpoints stay reachable to diagnose/fix:
- `GET /api/storage/health`, `POST /api/storage/test-connection`
- `POST /api/storage/config/storage`, `POST /api/storage/config/retry`
- `POST /api/storage/config/use-file-storage` (explicit fall back to local files)

### 2. Unrecognized `authType` no longer silently degrades to basic auth
A cluster configured with an invalid `authType` (e.g. `'aws'` — a common mistake; valid values are `'sigv4' | 'basic' | 'none'`) fell through to the basic-auth branch and produced an opaque `Response Error` (401/403).

**Fix** (`server/services/opensearchClientFactory.ts`, `server/routes/storage/admin.ts`): `createOpenSearchClient` now
- **infers `sigv4`** when `authType` is omitted but `awsProfile`/`awsRegion` are present (the common AWS case — basic auth could never have worked there), and
- **throws a clear error** for an unrecognized `authType` instead of silently using basic auth.

`POST /api/storage/config/storage` validates `authType` up front and returns **400** with the valid options, so the Settings UI gives immediate, actionable feedback.

## How they were found
While verifying #314, an isolated instance pointed at OpenSearch via `state.json` kept returning `Response Error`. Root cause: `authType: 'aws'` → basic auth (finding 2). After fixing that, a *different* misconfig showed the UI still serving local file data while reporting "error" (finding 1).

## Tests
- `tests/unit/server/services/opensearchClientFactory.test.ts` — +2 cases: infers `sigv4` when `authType` omitted + AWS fields present; throws on `'aws'` (and does **not** construct a basic-auth client).
- `tests/unit/server/routes/storage/errorStateGuard.test.ts` (new) — 503 for entity CRUD in error state; admin/recovery routes stay reachable; normal file-mode passthrough.
- `tests/unit/server/routes/storage/index.test.ts` — updated mount-order assertion for the new guard.

```
$ npm run build:all          # ✓ green
$ npx jest tests/unit/server/routes/storage/ \
    tests/unit/server/services/opensearchClientFactory.test.ts \
    tests/unit/server/services/storageInitializer.test.ts
  Test Suites: 11 passed, 11 total
  Tests:       290 passed, 290 total
```

## Backwards compatibility
- No dependency changes (server code + tests only).
- `authType` inference only triggers when it was **omitted** and AWS fields are present — basic-auth users (username/password, no AWS fields) are unaffected.
- The 503 guard only fires when a cluster is configured **and** unreachable; pure file-storage mode and healthy-OpenSearch mode are unchanged.

## Notes
- `npm audit --audit-level=high` reports 5 high advisories that are **pre-existing on `main`** (this PR changes no `package.json`/lockfile).
- Integration/e2e suites that need real Bedrock/OpenSearch/Playwright are left to CI (the full matrix is the source of truth).

## Checklist
- [x] DCO signoff
- [x] `CHANGELOG.md` updated under `[Unreleased] → ### Fixed`
- [x] `npm run build:all` succeeds
- [x] New source file has SPDX header
- [x] Unit tests added for both fixes

🤖 Authored with assistance from Claude.


goyamegh:
## Verification

**Build** — `npm run build:all` (frontend + server + cli + lib): ✅ green, 0 TS errors.

**Tests** — the two fixes + regression of all related storage suites:
```
$ npx jest tests/unit/server/routes/storage/ \
    tests/unit/server/services/opensearchClientFactory.test.ts \
    tests/unit/server/services/storageInitializer.test.ts
  Test Suites: 11 passed, 11 total
  Tests:       290 passed, 290 total
```

New/changed cases:
- `opensearchClientFactory.test.ts`
  - ✓ infers `sigv4` when `authType` omitted but `awsProfile`/`awsRegion` present
  - ✓ throws on unrecognized `authType` ('aws') and does NOT construct a basic-auth client
- `errorStateGuard.test.ts` (new)
  - ✓ 503 for entity CRUD when a configured cluster is unreachable
  - ✓ config/recovery admin routes stay reachable in error state
  - ✓ no block in normal file mode

**Behavior matrix (manual, isolated instance):**

| State | `/api/server-info` backend | `/api/storage/test-cases` | local file data |
|-------|----------------------------|---------------------------|-----------------|
| file mode | `file` | 200 (file data) | served |
| OpenSearch reachable (`sigv4`) | `opensearch` | 200 (cluster data) | hidden, preserved on disk |
| OpenSearch unreachable | `error` | **503** (was: 200 file data) | preserved on disk |
| `authType: 'aws'` save | — | — | **400** at save (was: opaque `Response Error`) |

**Audit:** `npm audit --audit-level=high` → 5 high, all **pre-existing on `main`** (this PR changes no `package.json`/lockfile).

TackAdam (approved):

jasonlhamazon (approved):

goyamegh:
Thanks for the review — all five addressed in `7630ab94`:

1. **`configToCacheKey()` authType inference** — extracted a shared `resolveAuthType(config)` helper now used by **both** `createOpenSearchClient()` and `configToCacheKey()`, so an inferred-SigV4 config is keyed `sigv4|<endpoint>|<region>|<profile>|<service>` (not `basic|…`). Fixes the client-pool collision + drift-detection inconsistency. Regression test added (`uses a sigv4 key … when authType is inferred`).

2. **Startup crash on invalid `authType`** — `initializeStorageFromConfig()` now builds `configKey` + the client **inside** its `try/catch` (with a guarded `client?.close()`), so an invalid on-disk/TS/env `authType` resolves to a controlled `{ backend: 'error' }` state instead of throwing out of `createApp()`. Regression test added (`invalid authType → error state, no throw`).

3 & 4. **Weak file-mode test** — the `getStorageModule` stub now provides `isConfigured()`, and the file-mode case asserts **`200` + `{ testCases, total }`** (previously only `!== 503`, which a 404/500 would have passed).

5. **Observability route** — `POST /api/storage/config/observability` now validates `authType` and returns **400** with the valid options, matching the storage route.

`npm run build:all` green; 293 related tests pass.

PR #390, opened 2026-08-14
fix(benchmark): forward run-level judgeModelId through the benchmark-execute path

goyamegh:
### Description

Fixes the silent judge-model drop on the benchmark-execute path.

`POST /api/storage/benchmarks/:id/execute` → `executeRun`'s classic (non-SDK) path passed `{ evaluatorId, skipJudge }` to `runEvaluationWithConnector` but **dropped `judgeModelId`** — the only caller that did. The sibling paths (`runSingleUseCase`, `executeEvaluationRun` used by `/api/storage/evaluation-runs`) forward it correctly, which is why the existing `judgeModelIdFlag` integration test never caught this. The judge then silently resolved via the fallback chain (`BEDROCK_MODEL_ID` env → the agent's own model), so a run configured with a specific judge model could be scored by a completely different judge — in the worst case (demo-modeled agent + no `BEDROCK_MODEL_ID`) the mock judge, producing untrustworthy pass rates with no signal.

**Empirical proof of the bug**: executing a benchmark with `judgeModelId: "us.anthropic.totally-bogus-model-v99"` *completed with a real verdict* on the unfixed server — the bogus model was never used.

### Changes

- **`services/benchmarkRunner.ts`** — forward `judgeModelId: run.judgeModelId` in the classic path (matching the SDK path's `bindJudge` and `runSingleUseCase`), and stamp `judgeModelId`/`evaluatorId` onto the report before saving (the connector return path doesn't carry them; the trace-mode polled judge reads them off the report).
- **`server/services/storage/index.ts`** — `saveReportWithClient` now persists `judgeModelId` and `evaluatorId` (parity with `saveReportWithModule`), so the run-detail audit trail shows which judge scored each report on this path too.

### Tests

- **Unit** (`tests/unit/services/benchmarkRunner.test.ts`): `executeRun` forwards run-level `judgeModelId`/`evaluatorId` to the judge options and stamps them on the saved report.
- **Integration** (`tests/integration/server/routes/benchmarkExecuteJudgeModel.integration.test.ts`): full HTTP round-trip through `benchmarks/:id/execute` with agent model = Bedrock key and `judgeModelId: demo-model`. Only the run-level judge model resolves to the mock provider, so the mock reasoning in the persisted report proves the value reached the judge (every fallback would produce real LLM prose or an error). Validated **red on the unfixed server → green on the fixed one**. Self-skips without OpenSearch storage (the route 400s in file mode) and cleans up all created data.

### Check List


goyamegh:
**Verification proof**

Bug repro on the **unfixed** server — a bogus judge model should error the judge, but the run completed with a real verdict (the value never reached the judge):

```
POST /api/storage/benchmarks/<id>/execute
  { "agentKey": "demo", "modelId": "demo-model",
    "judgeModelId": "us.anthropic.totally-bogus-model-v99" }
→ data: {"type":"completed", ... "stats":{"passed":1,"failed":0,...}}   # should have errored
→ persisted report: judgeModelId: None, llmJudgeResponse.modelId: None  # audit trail also lost
```

New integration test, **red against the unfixed server**:

```
✕ run-level judgeModelId reaches the judge and is persisted on the report (12527 ms)
    Expected: "demo-model"
    Received: undefined
```

**Green against a server built from this branch** (same OpenSearch storage cluster):

```
✓ run-level judgeModelId reaches the judge and is persisted on the report (5277 ms)
Tests: 1 passed, 1 total
```

Unit suite:

```
tests/unit/services/benchmarkRunner.test.ts   52 passed, 52 total
  ✓ forwards run-level judgeModelId to the judge and stamps it on the saved report (classic path)
```

Full suites on the branch: unit 4521 passed · integration 746 passed · e2e 399 passed (2 pre-existing `skills.spec.ts` env-dependent failures reproduce identically on a clean `origin/main` checkout).

TackAdam (approved):

PR #356, opened 2026-07-06
respect TLS_SKIP_VERIFY environment variable

joshuali925:
### Description
Following readme I got 
```json
{"error":"storage_unavailable","message":"Configured OpenSearch storage (https://localhost:9200) is unreachable: unable to verify the first certificate. Refusing to silently read/write local file data. Fix the cluster then POST /api/storage/config/retry, or POST /api/storage/config/use-file-storage to switch to local files.","backend":"error","configuredEndpoint":"https://localhost:9200"}
```
since certs are self signed in docker, i set the tls skip verify variables, but still got error

This PR fixes reading the environment variables.

### Issues Resolved
_List any issues this PR will resolve, e.g. Closes [...]._

### Testing & Coverage
- [ ] Unit tests added/updated
- [ ] Integration tests added/updated (if backend changes)
- [ ] E2E tests added/updated (if UI changes)
- [ ] `npm run build:all && npm run test:all` passes locally
- [ ] Patch coverage >= 90% (new/changed lines are tested)
- [ ] No overall coverage decrease (ratchet enforced by Codecov)

### Compliance
- [ ] DCO signoff on all commits (`git commit -s`)
- [ ] CHANGELOG.md updated
- [ ] SPDX license header on new files

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #392, opened 2026-08-21
Restricted-bash evidence judge: full-fidelity, zero-copy evidence for the agentic judge

ashwin-pc:
### Description

This PR gives the agentic judge complete, inspectable evidence without giving it a
general-purpose process or filesystem escape hatch.

#### Problem

- The plain `pi` judge is unrestricted: its subprocess has full `bash`, `edit`, and
  `write` capabilities.
- The safer agentic judge currently receives only an inline `compactTrajectory`.
- That trajectory is truncated, so the judge can degrade or make unverifiable claims
  when correlated traces are unavailable.
- A judge should verify claims against the complete evidence, not a lossy narrative.

**Depends on / used by:** used by [#453](https://github.com/opensearch-project/agent-health/pull/453) (the compiler's claim evaluator); [#445](https://github.com/opensearch-project/agent-health/issues/445) tracks this lineage.

##### 1. Deliberately narrow command surface

The judge receives one `bash`-named tool, but it is an in-process interpreter, not an
OS shell. Its complete command set is `jq`, `grep`, `rg`, `sort`, `cat`, `ls`, `find`,
`head`, `tail`, `wc`, `uniq`, `cut`, `tr`, `sed`, `echo`, and `pwd`; pipelines,
`&&`/`||`/`;`, and scratch-only redirection are supported
([`restrictedBash.ts:20-23`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/server/services/restrictedBash.ts#L20-L23)).

- `grep` and `rg` always perform fixed-string matching. `-E`, `-G`, and `-P` are
  rejected with guidance to use fixed strings (`-F` semantics)
  ([`:902-936`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/server/services/restrictedBash.ts#L902-L936)).
- `sed` has no substitution or regex surface; it accepts only `sed -n 'N,Mp'`
  line ranges ([`:1066-1074`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/server/services/restrictedBash.ts#L1066-L1074)).
- `jq` is the real `jq-wasm` engine; all other commands are the fixed JavaScript
  built-ins above. jq runs inside the preemptible worker boundary
  ([`:1077-1128`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/server/services/restrictedBash.ts#L1077-L1128)).
- Variables, command substitution, backticks, subshells, backgrounding, glob
  expansion, `cd`, and unknown commands are rejected.

**Residual risk:** `jq-wasm` still exposes jq's regex/filter engine; its safety boundary is hard worker preemption and byte/time budgets, not regex analysis.

##### 2. Preemptible execution and judgment-wide budgets

Each parsed pipeline—including traversal and every built-in—runs in one persistent
`worker_threads` worker per judgment. The main thread waits for the worker's `ready`
message before arming the command deadline; if the deadline is crossed it awaits
`worker.terminate()` before returning the timeout result. A later command can create a
fresh worker ([`restrictedBash.ts:510-640`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/server/services/restrictedBash.ts#L510-L640)). jq's nested WASM worker is also terminated on completion or timeout
([`:1085-1128`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/server/services/restrictedBash.ts#L1085-L1128)).

Default limits are:

| Scope | Default |
|---|---:|
| One command/pipeline | 2,000 ms |
| One judgment | 200 commands |
| One judgment | 60,000 ms cumulative wall time |
| One input file | 10 MiB |
| Aggregate input to one command | 25 MiB |
| Returned output from one command | 50 KiB |
| Scratch writes | 100 MiB / 500 files |

The defaults and overrides are defined at
[`restrictedBash.ts:31-36`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/server/services/restrictedBash.ts#L31-L36) and
[`:249-256`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/server/services/restrictedBash.ts#L249-L256), and judgment admission checks are at
[`:510-521`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/server/services/restrictedBash.ts#L510-L521). Timeout, command-count, total-time, and output breaches are returned as
`details.resourceBreach` in the tool result
([`evidenceJudgeTools.ts:46-56`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/server/services/evidenceJudgeTools.ts#L46-L56)). Oversized inputs fail before the read.

The implementation launches **zero OS child processes**; this is also asserted
statically by the adversarial suite.

```mermaid
flowchart LR
  J[Evidence judge] --> C[Restricted command]
  C --> A[Budget admission]
  A --> W[Persistent command worker]
  D[2 s deadline] -. terminate on breach .-> W
  W --> E[Read-only evidence]
  W --> S[Scratch only]
  W --> R[Result + breach metadata]
  R --> J
```

##### 3. Zero-copy mount hardening

Trace files and the post-run workspace are virtual read-only mounts; no symlink or
workspace copy is created in the judgment directory
([`judgeEvidence.ts:82-89`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/server/services/judgeEvidence.ts#L82-L89),
[`:131-145`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/server/services/judgeEvidence.ts#L131-L145)). Mount admission and every later content read are tied to the admitted filesystem objects:

1. Mount creation canonicalizes the source, recursively snapshots `(dev, ino)` for
   every admitted entry, rejects symlinks, and rejects regular files with `nlink > 1`
   ([`restrictedBash.ts:262-332`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/server/services/restrictedBash.ts#L262-L332)).
2. Each directory-mount access must remain canonical, underneath the allowed root,
   and equal the mount-time `(dev, ino)` snapshot
   ([`:384-403`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/server/services/restrictedBash.ts#L384-L403)).
3. Content is opened with `O_NOFOLLOW`; post-open `fstat` rechecks file type,
   `(dev, ino)`, and—inside directory mounts—`nlink`, closing the validation/open race
   ([`:414-444`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/server/services/restrictedBash.ts#L414-L444)).
4. Directory-mount hardlinks are rejected both at admission and if an alias is added
   later. Writes remain confined to quota-limited `scratch/`.

**Residual risk:** This is not a mount namespace: bind/overlay/FUSE content already present inside the trusted workspace when the inode snapshot is built is admitted, so mount provenance remains part of the workspace trust boundary.

**Residual risk:** The identity snapshot is not a byte snapshot; in-place mutation of an admitted inode is not detected, so correctness still relies on the documented post-quiescence/immutability contract.

**Residual risk:** The recursive mount-admission snapshot runs before the command worker/deadline and has no entry-count budget; an extremely large trusted workspace can still make initialization expensive.

#### Adversarial test inventory

| Test | What it proves | Source |
|---|---|---|
| Catastrophic-looking grep/rg patterns | `(a\|b){1,10}{1,10}`, `(a+)+`, and `(a\|aa)+$` are data, not regex, in both engines; all six queries finish within the asserted 8 s suite bound | [`restrictedBash.test.ts:243-252`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/tests/unit/server/services/restrictedBash.test.ts#L243-L252) |
| Regex flags and sed substitution | `grep -E/-P`, `rg -G`, and sed substitution are rejected structurally | [`:254-258`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/tests/unit/server/services/restrictedBash.test.ts#L254-L258) |
| Busy jq/WASM | A billion-element jq range hits a 100 ms deadline, exits nonzero, and returns within 750 ms | [`:346-354`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/tests/unit/server/services/restrictedBash.test.ts#L346-L354) |
| 300k-line sort pipeline | A 50 ms deadline preempts `sort | uniq | sort`; elapsed time is <500 ms and the worker-exit hook fires | [`:302-317`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/tests/unit/server/services/restrictedBash.test.ts#L302-L317) |
| Deep metadata traversal | A 150-level `find` is terminated under a 10 ms deadline; elapsed time is <410 ms and worker exit is observed | [`:319-331`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/tests/unit/server/services/restrictedBash.test.ts#L319-L331) |
| Command flood and judgment budgets | One worker services 50 cheap commands in <5 s; separate assertions reject command 201/cumulative-time exhaustion (with smaller test limits) and expose breach kinds | [`:333-368`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/tests/unit/server/services/restrictedBash.test.ts#L333-L368) |
| Worker startup/exit/reuse | A fresh worker's bootstrap is excluded from a 50 ms command deadline; forced-exit hooks plus 50-command reuse guard lifecycle regressions. This is a lifecycle proxy, not a direct active-handle count. | [`:302-344`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/tests/unit/server/services/restrictedBash.test.ts#L302-L344) |
| Traversal and symlink pivots | Rejects `..`, absolute paths, encoded traversal, sibling-file pivots, and symlink swaps without disclosing target bytes | [`:127-135`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/tests/unit/server/services/restrictedBash.test.ts#L127-L135), [`:162-229`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/tests/unit/server/services/restrictedBash.test.ts#L162-L229) |
| Hardlink aliases | Rejects file mounts with external hardlinks, directory mounts containing hardlinks, and aliases added after snapshot | [`:260-283`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/tests/unit/server/services/restrictedBash.test.ts#L260-L283) |
| Input/output exhaustion | Per-file and aggregate input caps fail before reads; output truncation carries narrowing guidance | [`:285-300`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/tests/unit/server/services/restrictedBash.test.ts#L285-L300), [`:370-376`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/tests/unit/server/services/restrictedBash.test.ts#L370-L376) |
| Zero child processes | Judge interpreter modules contain no `child_process`, `spawn`, `exec`, `execFile`, or `fork` use | [`:378-383`](https://github.com/opensearch-project/agent-health/blob/06137f4d417df6f925e3e799e839ffacbf1e8a1f/tests/unit/server/services/restrictedBash.test.ts#L378-L383) |

#### Review history: goyamegh B1 and mount hold

The first response added regex heuristics and Promise-level timeouts. The current
implementation instead changes the execution structure across `6e652e2`, `fc17fcf`,
`d24013c`, `65d9c02`, and `06137f4`:

| Review item | Current closure | Status / remaining boundary |
|---|---|---|
| JS regex screening could not govern grep/sed/jq engines | grep/rg are fixed-string only; sed regex/substitution is absent; regex flags are rejected | Structurally closed for grep/rg/sed. jq regex/filter work is bounded only by worker preemption. |
| Unbounded command flood / cumulative work | 200-command and 60 s cumulative wall budgets, 2 s per-command deadline, one reusable command worker | Closed for tool-issued commands; no OS subprocess fan-out exists. |
| Promise rejection did not prove work stopped | Deadline is armed after worker readiness and `terminate()` is awaited; tests observe worker exit and wall bounds | Closed for command execution. Worker bootstrap has a separate 30 s startup bound. |
| Path traversal and symlink TOCTOU | Canonical-root checks plus `O_NOFOLLOW` and post-open identity checks | Closed for post-admission pathname/symlink/inode-replacement pivots. |
| Hardlink disclosure | `nlink > 1` is rejected at snapshot and after open, including aliases added after directory-mount admission | Closed for regular-file hardlink aliases visible to this process. |
| Bind/overlay/FUSE aliasing | Every admitted entry is pinned by canonical path and `(dev, ino)` | A namespace-equivalent provenance check is not claimed; pre-existing nested mounts remain in the trusted input boundary. |
| Lost snapshot semantics | Mount-time identities prevent replacement/addition from becoming readable | Not a byte snapshot; in-place writes remain governed by the post-quiescence contract. |
| Metadata traversal | `find`, `ls`, and all command-side traversal execute inside the terminable worker | Command-side traversal is bounded; initial recursive mount admission remains unbudgeted. |

#### Evidence behavior and compatibility

The evidence bundle still contains the complete original trajectory, normalized
per-step files, testcase/run metadata, scratch space, and correlated canonical
file-mode trace mounts. Cluster traces are queried but not materialized. The
unrestricted plain `pi` judge remains deprecated.

> [!WARNING]
> **Breaking file-mode trace-store change:** the on-disk layout changes from
> `{traceId}.json` to `*-<id>.ndjson`. Legacy JSON files are not read or migrated;
> users must re-emit affected traces or delete the legacy files.

Known gaps retained from the original scope:

- File mode does not yet have a logs store; `query_logs` remains cluster-only.
- A follow-up can add a restricted PPL query tool for cluster mode.
- Behavioral span-query proof still needs trace-emitting benchmark cases.

### Testing & Coverage

- [x] Unit/adversarial tests added for command semantics, hard preemption, budgets,
  input/output limits, traversal, symlinks, hardlinks, mounts, and zero child processes
- [x] Integration coverage retained for file-mode OTLP → canonical NDJSON → read path
- [x] Focused verification at `06137f4`: 2 suites / 43 tests passed (`restrictedBash` + `judgeEvidence`)
- [ ] E2E tests added/updated (no UI change)
- [ ] Patch coverage >= 90% (left to Codecov)
- [ ] No overall coverage decrease (left to Codecov)

### Compliance

- [x] DCO signoff on all commits (`git commit -s`)
- [x] `CHANGELOG.md` updated
- [x] SPDX license header on new files

By submitting this pull request, I confirm that my contribution is made under the
terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off
your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).


ashwin-pc:
Added per-run connector workspace evidence in 4ef2362. The judge now prefers connector `metadata.workspaceDir` over static `connectorConfig.cwd` (with fallback), so temporary fixture runs audit the directory that actually executed.

ashwin-pc:
## Code-Diff-Analyzer triage

Follow-up commits `aee862e` and `39025e1` address the actionable findings:

| Finding | Resolution |
|---|---|
| Private CodeArtifact URL for `jq-wasm` | **Fixed.** Lockfile now uses the canonical public npm tarball. `npm view jq-wasm@3.0.0-jq-1.8.2 --registry=https://registry.npmjs.org/` reports the same tarball and exact lockfile integrity (`sha512-jgWSEBJSd0lYR4Q5...Ksp2g==`), MIT license. Dependency Review, Mend, and `npm audit` security-scan all pass. |
| Fixed worker string + `jq-wasm` could execute tampered code | **Supply-chain premise fixed; execution design retained.** The worker source is a fixed application string, takes only structured `workerData`, has no shell/`child_process`, and is terminated on the configured deadline. `jq-wasm` is a normal integrity-pinned dependency, not untrusted runtime input. A worker is required because WASM can block the main event loop. |
| Client-controlled `workspaceDir` could snapshot arbitrary server directories | **Fixed.** `/api/judge` now ignores request metadata paths and permits only server-owned configured agent `cwd`; added a regression test using `/etc`. Trusted in-process connector evaluation can still use its per-run workspace. |
| `evidenceDir` absolute path returned in API debug response | **Fixed.** Removed from `judgeDebug`; keep-evidence remains an explicit server-side diagnostic option and logs the path only server-side. |
| `AH_JUDGE_SERVER_URL` could redirect internal trace requests | **Not a request-controlled SSRF.** It is process-start configuration controlled by the deployer, analogous to other backend endpoint environment variables. The request cannot influence it; its intended purpose is selecting this server's internal route origin. |
| Optional dependency bin changed from `dist/cli.js` to `./dist/cli.js` | **Fixed.** Reverted this unrelated lockfile churn. |
| Current mandatory dependency-addition flag for public `jq-wasm` | **Reviewed / accepted.** The bot says this is mandatory regardless of legitimacy. Public npm metadata, exact SHA-512 integrity, MIT license, Dependency Review, Mend, audit, and real jq tests across Node 18/20/22 provide the requested verification. |

The Node 18 matrix exposed that jq-wasm expects global Web Crypto; `39025e1` installs Node 18's `node:crypto.webcrypto` in the worker before import. The same jq evidence tests now pass under Node 18 locally.

goyamegh:

This is the highest-risk PR in the current batch (34 files, judge-subsystem rewrite + a breaking file-mode trace-storage layout change), so I read the core new files directly (`restrictedBash.ts`, `judgeEvidence.ts`, `evidenceJudgeTools.ts`, `TraceStore.ts`, the `judge.ts`/`piAgenticJudgeService.ts` wiring) in addition to running the adversarial pass, and verified a few of the review's claims against the actual source/installed SDK before including them below.

**Migration & CI coverage (explicitly checked, per the review brief):**
- **No migration/fallback for legacy traces, and it's intentional**: `TraceStore.ts`'s new read paths only look at `*.ndjson`; old `*.json` files are silently ignored (not deleted, not converted). This matches the PR's own `⚠️` breaking-change note ("Legacy JSON files are not read or migrated; users must re-emit affected traces or delete the legacy files"), so it's a disclosed tradeoff, not an oversight.
- **CI does cover the new NDJSON path**: `tests/unit/server/adapters/file/TraceStore.test.ts` (bucketing by run/session/trace id, flattening, `raw`-field round-trip) and `tests/integration/server/otlp-file-traces.integration.test.ts` (real OTLP POST → NDJSON on disk → read back via `/api/traces`) both exercise the new format directly.
- **One thing I checked and ruled out as a false alarm**: I initially suspected that registering a custom `bash` tool via the inline extension factory while also passing `tools: ['bash', ...]` to the pi SDK could let the SDK's own real, unrestricted built-in `bash` tool (which does spawn processes) leak into the session alongside/instead of the restricted one. I traced this through the installed `@earendil-works/pi-coding-agent` SDK's compiled `agent-session.js`: it builds the final tool registry by inserting built-ins first, then overwriting same-named entries with extension-registered tools — so the custom restricted `bash` always wins the name collision and the real built-in is never invoked. Not a vulnerability; flagging so it doesn't need re-checking.

**Findings from the adversarial pass:**

|---|---|---|
| Blocker (worth resolving before merge) | `restrictedBash.ts`'s `outputCapBytes`/`timeoutMs` cap the *returned* text, but not memory/CPU used *while* a command runs: `cat`/`sort`/`uniq`/`cut`/`sed` slurp entire files and build full output before truncation, and `grep`/`sed`/`rg` run synchronous `new RegExp(...)` against attacker-influenced patterns (the judge model supplies the command string) — a pathological pattern or a very large evidence/mounted file can block the event loop or spike memory before the timeout/cap ever applies, since a synchronous regex scan can't be preempted by the `Promise.race` timeout. | Worth considering: a regex complexity guard (or a non-backtracking engine) for `grep`/`sed`, and streaming/chunked reads with an early-bail size check before building full in-memory output. |
| Blocker (worth resolving before merge) | Related to the above: `judgeEvidence.ts`'s `copyWorkspace()` recursively copies the *entire* run workspace into a fresh `os.tmpdir()` per judgment with no size/file-count cap before `makeReadOnly()` — a large workspace (cloned repo, `node_modules`, build artifacts) means a judge could copy gigabytes per judgment, and there's no janitor sweeping abandoned tmpdirs left by a crash/OOM (or by `AH_JUDGE_KEEP_EVIDENCE`) — under concurrent judging this is a plausible disk-exhaustion path distinct from the bash tool's own scratch quota (which only limits *new* writes, not the initial copy). | Consider a workspace-copy size cap (skip/truncate beyond N MB) and/or a periodic sweep of `agent-health-judge-*` tmpdirs older than some TTL. |
| Major | `server/routes/judge.ts` correctly refuses to trust a caller-supplied `evidenceContext.workspaceDir` directly, but it still uses the caller-supplied `evidenceContext.agentKey` to pick *which* server-configured agent's `connectorConfig.cwd` gets copied into evidence — a caller who knows another configured agent's key could get that agent's workspace snapshotted into their own judgment bundle. | Worth checking whether the API caller is expected to only ever supply their own run's `agentKey` (if so this may be low-risk in practice), or whether the lookup should also verify the caller's `runId`/session owns that agent invocation. |
| Major | `TraceStore.writeSpans()`'s read-merge-write-per-file pattern already existed pre-PR, but bucketing by `run.id`/`session.id` instead of `traceId` means many more spans across a whole run now share one canonical file — concurrent OTLP POST batches for the same run/session racing on that one file's read-merge-rename can silently drop the loser's spans (no locking). | Consider a lock (even a simple file-based one) per canonical file, or documenting that concurrent same-run OTLP ingestion isn't safe yet. |
| Major | Mount discovery for trace evidence (`fetchTraceJudgeSpans(runId, serverUrl, agents)` → `size: 500`) caps at 500 spans; `judgeEvidence.ts` only mounts the canonical files backing *those* returned spans (`canonicalFilesForSpans`). For a run with >500 spans spread across multiple correlation buckets (e.g. partial instrumentation where some spans carry `session.id` and others fall back to `traceId`), the "full-fidelity" evidence bundle could be missing some of the run's canonical NDJSON files entirely. | Might be worth a follow-up to enumerate canonical files by run/session identity directly (not by first-500-spans) when building the mount table, if runs with >500 spans are a realistic scenario. |
| Minor | The custom interpreter reimplements `grep`/`sed`/`sort`/`cut`/`tr` (not just `cat`/`ls`/`find`/`jq`) — the "zero child processes" constraint clearly justifies *some* bespoke interpreter, but the wider the reimplemented surface, the more places a subtle behavioral difference from real GNU tools (or a parsing bug) can either be exploited or just quietly mislead the judge. | No specific bug found beyond the regex-DoS point above; just flagging that narrowing the command surface (or leaning more on `jq` for filtering) would reduce both attack surface and maintenance cost, if that's compatible with the intended judge workflows. |
| Minor | Test coverage: the traversal/symlink/quota/timeout tests in `restrictedBash.test.ts` are strong, but there's no test yet for evidence cleanup when `loadPiSdk()`, `resourceLoader.reload()`, or `session.prompt()` throws mid-judgment (the `finally` in `piAgenticJudgeService.ts` looks structurally correct for this, so likely fine — just not asserted), and no test for a large/many-file workspace snapshot or the `evidenceContext.agentKey`-selects-another-agent's-workspace path above. | Consider adding targeted tests for these paths given they're the ones most likely to only show up in production load. |

**Overall**: the direction — avoid `child_process` entirely, materialize immutable evidence, mount trace data read-only instead of copying — is sound, and the path-traversal/symlink/mount-pivot defenses are genuinely well built and well tested (verified `restrictedBash.test.ts` covers `../` escapes, absolute paths, symlink-swap TOCTOU, and quota enforcement directly). The main things I'd want a second look at before merge are the resource-exhaustion paths (unbounded workspace copy, un-preemptible regex/large-file processing) and the >500-span mount-discovery gap, since those undercut the "complete, full-fidelity evidence" claim under realistic load rather than being purely theoretical. The trace-store breaking change is clearly disclosed and tested for the new format, so that part looks merge-ready on its own.

ashwin-pc:
Thanks for the thorough review — all findings from it are addressed on this branch:

- **Restricted-command resource limits** (`801aa48`): regex guardrails (pattern-length cap, nested-quantifier rejection before matching) plus per-file and aggregate input-byte limits checked *before* reads, with errors that suggest `-F`, bounded expressions, or narrowing via `find`/`head`.
- **Judge authorization** (`a7fab41`): `agentKey` is now derived from the stored evaluation-run/report record. A caller-supplied key that differs from trusted metadata returns 403, and only the trusted key selects the configured workspace.
- **Workspace-copy exhaustion** (`9fd1fe8`): rather than capping the copy, the copy is gone. The run workspace is mounted **read-only** into the judgment tree at `evidence/workspace/`, reusing the mechanism this PR already uses for canonical trace NDJSON. There are no bytes to bound and nothing to clean up; traversal, absolute/sibling escapes, symlinks, changed canonical roots, and writes are all rejected. A test asserts a 32 MiB workspace mounts with no workspace inode created in the judgment tree.

Two trace-store findings are being handled as follow-ups rather than patched here, because the better fixes are structural:

- **Concurrent NDJSON writes**: instead of locking the read-merge-write, ingestion moves to **append-only** writes with read-time dedup by span id — the race stops existing rather than being guarded.
- **500-span evidence discovery cap**: canonical files will be enumerated by run identity in the file adapter (mounting is inherently file-mode), with the query path retained and paginated for cluster deployments. Agreed that evidence completeness shouldn't depend on a presentation limit.

Narrowing the emulated command surface and the additional cleanup/large-workspace/cross-agent tests are tracked as follow-ups.

goyamegh:
## Codex Review: Blocker Verification for 801aa487d7, a7fab41015, 9fd1fe8704

**Blocker Status after 3 fix commits:**

### ❌ **B1 (Regex DoS + Spawn Exhaustion): NOT CLEARED**

**Findings:**
- `validateRegex()` is a heuristic that only blocks one narrow pattern family (parenthesized group `*`/`+` followed by another quantifier). Misses other DoS-inducing shapes like `(a|b){1,10}{1,10}` or catastrophic alternation.
- More critically: regex *screening* in JavaScript then execution in different engines (`grep`, `sed`, `jq-wasm`) is not a credible closure. Each engine has its own backtracking semantics.
- **Zero caps** on total commands, concurrent subprocesses, directory walks, or cumulative CPU time. A malicious judge still exhausts resources via command floods (e.g., 1000 individually "safe" `grep` calls, each burning seconds).
- Timeout implementation unproven: `DEFAULT_TIMEOUT_MS = 10_000` exists, but no proof that all subprocesses are forcibly killed or that `jq-wasm` workers actually terminate. Promises rejected ≠ workers killed.

**Must-fix:** Per-judgment command/concurrency/CPU budgets + credible timeout enforcement for all spawned/threaded workers.

---

### ✅ **B2 (Workspace Copy Exhaustion): CLEARED (with caveats)**

**Findings:**
- `copyWorkspace()` function is removed; workspaces are now mounted via `fs.realpath()` reference only. Zero disk amplification from the original blocker.
- Per-file and aggregate input size limits still enforced during reads from mounted workspace.

**BUT:** Removal of copy-based snapshots introduces **new risks** (see below).

---

## 🚨 New Risks Introduced by Mount-Only Strategy

**`HIGH` severity:**

1. **Bind/Overlay/FUSE Escape** — `validateDirectoryMountPath()` uses pathname ancestry (`inside(mount.sourceRoot, realpath(path))`), not inode/mount ancestry. Paths remain under `sourceRoot` logically while bytes come from external mounts. A workspace can alias sensitive host files without changing resolved pathnames.

2. **Hard-Link Disclosure** — Unaddressed. A hard-linked file inside the workspace can alias a sensitive file outside without changing the pathname. If the workspace creator shares a UID with the judge, this is a file-disclosure vector. `realpath` doesn't detect hard links.

3. **TOCTOU Symlink Swap** — Code checks `isSymbolicLink()` (likely on `fs.stat`, which doesn't see symlinks). Even if it's `fs.lstat`, validation and read/open happen as separate operations against a live, mutable workspace. A symlink can be inserted after validation but before read.

4. **Lost Snapshot Semantics** — Evidence is no longer an immutable bundle but a live reference to the workspace. A judge can observe different bytes across commands or catch files mid-write, weakening evidence integrity for evaluation purposes.

**`MED` severity:**

5. **Unbounded Metadata Traversal** — Even without byte copying, `find`, `ls`, and glob operations can enumerate millions of inodes on a tree the workspace creator controls. I/O and CPU exhaustion remain possible.

---

## Recommendation

**Hold remains in place.** B1 is not cleared and mount-based security is weaker than the initial design on path/inode traversal and race conditions.

**To lift hold, require:**
1. **B1 fix:** Command count budget, concurrency limit, CPU-time budget per judgment + credible proof that all workers are forcibly terminated on timeout.
2. **Mount safety:** Replace path-based validation with inode/mount ancestry checks or read-only snapshot mounts.
3. **Hard-link closure:** Document UIDs or use a separate mount namespace to prevent hard-link aliasing.
4. **Test suite:** Add adversarial tests for hard links, nested mount escapes, TOCTOU symlink swaps, and command floods.
5. **Spec update:** If evidence is no longer immutable, document that in the evaluation spec.

**Summary:** The PR fixes the immediate disk-copy problem but moves the risk surface to path traversal, inode aliasing, and unbounded command execution. Blocker B1 is still open.

ashwin-pc:
Thanks — I agreed B1 needed a structural fix rather than more regex screening.

1. **Regex engines:** `6e652e2` removes regex from `grep`/`rg` (always fixed-string) and reduces `sed` to `-n 'N,Mp'`; regex flags/substitution are rejected. Tests cover the cited catastrophic shapes in both grep engines and sed rejection (`tests/unit/server/services/restrictedBash.test.ts:243-258`). `jq-wasm` still has jq regex/filter capability; it is bounded by preemption, not regex analysis.

2. **Forced termination:** `d24013c` moves each parsed pipeline, including traversal and built-ins, into one persistent `worker_thread`. `65d9c02` arms the deadline only after `ready`; `06137f4` makes bootstrap portable. On breach the caller awaits `worker.terminate()`. A 300k-line sort under a 50 ms budget is terminated at ~54 ms (test bound 500 ms) and fires the exit hook (`:302-317`); deep `find` and busy jq have equivalent wall assertions (`:319-331`, `:346-354`).

3. **Judgment budgets/spawn exhaustion:** `fc17fcf` adds defaults of 200 commands, 60 s cumulative wall time, 2 s/command, 10 MiB/file, 25 MiB aggregate input, and 50 KiB output. Budget breach kinds are returned in tool metadata. Tests cover worker reuse/flood and count/time refusal (`:333-368`). The judge modules statically assert zero `child_process`/spawn/exec/fork (`:378-383`).

4. **Mounts:** `6e652e2`/`fc17fcf` snapshot canonical `(dev,ino)`, reject symlinks and `nlink>1`, then use `O_NOFOLLOW` plus post-open `fstat`/`nlink`. Traversal, swaps, and pre/post-snapshot hardlinks are tested (`:127-229`, `:260-283`).

Residual boundary: this is not a mount namespace; pre-existing bind/overlay/FUSE content is trusted at admission. It is an identity snapshot, not a byte snapshot, so in-place mutation still relies on workspace quiescence. Initial recursive snapshot traversal is not command-budgeted.

CI note: the only remaining red is `codecov/project`; it compares against main's post-#487 baseline (+8.4% from the restored e2e suite) and reads low on every open PR in the repo. Patch coverage on this branch is 83%.

goyamegh:

PR #393, opened 2026-08-21
perf: lazy report loading + infinite scroll for run inspector and runs list

goyamegh:
### Description

The run-inspect page (`/evaluations/benchmarks/:id/runs/:id/inspect`) fetched **every full report document** (trajectory + judge output, ~0.3–2 MB each) just to paint pass/fail status badges. An 84-test-case run shipped **~68 MB across ~90 requests and took ~10 s to first paint** locally — minutes over a slow link. `/evaluations/runs` had the same disease worse: a full-report N+1 across the first 100 runs just to count annotations.

**Measured on the same 84-case run: 10.4 s → 2.7 s first paint, 68 MB → 0.7 MB, ~90 requests → 5.**

#### Changes

- **Server**: `GET /api/storage/runs?ids=…&fields=…` — optional top-level field projection on the existing batch path (`id` always included, unknown fields skipped). No-`fields` behavior unchanged (still strips `rawEvents`).
- **Server**: `GET /api/storage/test-cases?ids=…` now fans out its per-id lookups in **parallel** (was sequential — ~7 s alone for 84 ids at ~80 ms/OpenSearch round-trip).
- **Client**: `asyncRunStorage.getReportSummariesByIds()` — lightweight status/verdict batch, chunked at 100 ids/request.
- **RunInspectorPage**: one summary batch for all statuses (header tallies complete immediately), full report fetched on-demand for the **selected row only**, test-case list windowed with **infinite scroll** (100 rows/page, sentinel + IntersectionObserver), and an **error + Retry state** instead of an infinite skeleton when the API is briefly unavailable (e.g. server restart).
- **EvalRunsPage**: run rows windowed with **infinite scroll** (50/page, flat + grouped modes); annotation counts loaded for **rendered rows only** via the lightweight batch.
- **BenchmarkRunsPage (evals3)**: “Load More Runs” auto-triggers when scrolled into view (button kept as fallback).

### Testing

- **Unit** (`tests/unit/services/storage/asyncRunStorage.test.ts`): summary-batch field list, `traceId→runId` mapping, chunking at 100 ids.
- **Integration** (`tests/integration/server/routes/storage/runsBatchFields.integration.test.ts`): projection returns only requested fields + id; unknown fields skipped; no-fields batch unchanged; multi-id.
- **e2e** (`tests/e2e/lazy-report-loading.spec.ts`): network-level assertion that statuses load via one `fields=` batch and full-report fetches only target the selected row; infinite scroll reveals rows on sentinel scroll for both the inspector list and the runs table.

### Check List
- [x] New functionality has been documented (CHANGELOG under Unreleased).


goyamegh:
### Verification proof

**Before/after on the same 84-test-case benchmark run (headless Chromium against a local server, OpenSearch storage):**

Before (main):
```
LOADED in 10.4 s — rows: 84
/api/storage/runs* requests: ~90 (one full report per test case), ~68 MB transferred
```

After (this PR):
```
FIRST ROWS: 2.7s | rows: 84 | storage requests: 5 | bytes: 725 KB
statuses: 65 passed / 19 failed rendered
```

Per-request profile after the fix (the two hot spots are gone — one 33 KB `fields=` batch replaces 84 full-report fetches, and the test-cases `?ids=` batch dropped from 6.7 s sequential to sub-second parallel):
```
+0.7s 200    190KB   84ms /api/storage/benchmarks/<id>
+1.0s 200    156KB  300ms /api/storage/test-cases?ids=<84 ids>
+1.4s 200     33KB  400ms /api/storage/runs?ids=<84 ids>&fields=status,passFailStatus,metricsStatus,...
+1.6s 200   1778KB  137ms /api/storage/runs/<selected report only>
```

**Tests:**
```
Unit   — tests/unit/services/storage/asyncRunStorage.test.ts: 49 passed
Integr — tests/integration/server/routes/storage/runsBatchFields.integration.test.ts:
         ✓ returns only the requested fields (plus id) when fields= is given
         ✓ silently skips requested fields missing on the document
         ✓ batch without fields still returns full docs minus rawEvents (existing behavior)
         ✓ projects multiple ids in one request
e2e    — tests/e2e/lazy-report-loading.spec.ts:
         ✓ statuses load via one lightweight batch — no per-report full fetches (2.3s)
         ✓ test-case list reveals more rows on scroll (infinite scroll) (2.5s)
         ✓ run rows are windowed and more load when the sentinel scrolls into view (2.1s)
```

Full suite: `npm run build:all` ✓, unit 1430 passed, integration passed against a server running this branch, e2e 407 passed (4 failures are pre-existing on `main` in the same environment — `skills.spec.ts` fails identically on a pristine `origin/main` checkout, and the other 2 pass in isolation / are shared-file-storage parallelism flake). No dependency changes in this diff.

jasonlhamazon (approved):

goyamegh:
### Adversarial review pass (Codex, gpt-5.4) — findings + dispositions

Ran an adversarial second-opinion review over the diff; commit `3fec6188` addresses the accepted findings, each with a unit regression test:

|---|-----|---------|-------------|
| 1 | HIGH | Batch `?ids=` path skipped `demo-*` sample data — sample benchmarks kept real statuses on the old per-row path but lost them on the batch path | **Fixed** — batch path resolves `demo-*` ids via `getSampleRun()` like `GET /runs/:id` |
| 2 | HIGH | `toTestCaseRun` never mapped `judgeModelId`, so browser-side trace-recovery judging silently fell back to the agent model | **Fixed** (pre-existing — affected the old full-fetch path too) |
| 3 | HIGH | Error+Retry state only reachable for pre-`run` failures; a failure after `run` was set rendered a broken page with no retry | **Fixed** — error gate now takes priority over partial data |
| 4 | MED | `initialSelectionDone`/window state leaked across run navigations (React Router reuses the component instance) | **Fixed** — per-run reset effect (pre-existing for selection; new for window) |
| 5 | MED | `fields=` projection could be asked for `rawEvents`, defeating the batch payload contract | **Hardened** — `rawEvents` is never projectable. Kept the generic projection itself: the same route’s list path already exposes arbitrary `fields` → `_source`, so this matches existing API surface |
| 6 | MED | Benchmark-runs sentinel “drains pagination on tall viewports” | **Rejected** — loading pages until the sentinel leaves the viewport is the intended infinite-scroll behavior, with `isLoadingMoreRuns` as per-page backpressure |

Full unit suite after fixes: 256 suites / 4605 passed.

PR #394, opened 2026-08-23
fix: make the demo/mock judge loud so it can't be mistaken for a real pass

goyamegh:
### Description

The built-in `demo` judge provider returns a semi-random pass **without calling an LLM** — useful for local dev/demos, but the only hint that a verdict wasn't real was a footnote buried at the *bottom* of the reasoning text. A real benchmark whose judge model resolved to `demo-model` (e.g. a typo, a missing `BEDROCK_MODEL_ID`, or an unpinned judge) therefore rendered as a "100% pass" in the UI/CLI with nothing actually judged (WebIR user feedback).

This PR makes the mock verdict impossible to miss:

- Sets a structured `warning` field on `JudgeResponse` — a real provider never sets this field, so callers can detect "this wasn't a real judge" programmatically instead of grepping prose.
- Moves the `⚠️ MOCK JUDGE — NOT A REAL EVALUATION` banner to the **top** of `llmJudgeReasoning` (previously a footnote at the bottom).
- Emits an always-on `console.warn` server-side naming the resolved model and how to pin a real judge, so it shows up in server logs even if nothing reads the response body.

### Changes

- **`server/routes/judge.ts`** — `generateMockEvaluation()` now sets `warning` and leads `llmJudgeReasoning` with the banner; the `/api/judge` route handler emits a `console.warn` before returning the mock result for the `demo` provider.
- **`types/index.ts`** — adds the optional `warning?: string` field to `JudgeResponse`, documented as mock-judge-only.
- **`tests/integration/server/routes/judge.integration.test.ts`** — new test asserting (1) `warning` is a string matching `/MOCK_JUDGE/`, and (2) the banner appears before `Mock Evaluation Result` in the reasoning text (regression guard for the "banner at the bottom" bug).

### Testing

- `npm run test:all` (unit + integration + e2e) — unit and integration suites pass in full (unit: all green; integration: 747/747 passed, including the new mock-judge assertions). e2e: 395/419 passed locally; the handful of failures observed were in specs unrelated to this change (`assistant-modal`, `evaluation-runner` status badges, `navigation`, `quick-run-agent-preference`, `skills`, `test-case-runs`) and reproduced under heavy concurrent CPU load from sibling test runs sharing this dev box — none touch `judge.ts`, `types/index.ts`, or the judge integration test added here. Will confirm CI runs these clean in an isolated runner.
- `npm audit --audit-level=high` — pre-existing findings in transitive deps (js-yaml, nanoid, postcss, react-router, undici, vite), unrelated to this diff (no `package.json`/`package-lock.json` changes); consistent with `main`.
- DCO signoff present on the commit; `CHANGELOG.md` updated under `## [Unreleased]`.

### Check List
- [x] New functionality has been documented (CHANGELOG)


goyamegh:
## Verification proof

**Local checks** (`agent-health-judge-demo-warning` worktree, rebased onto `main`@`ae4317f8`):

```
npm run build:all   → PASS (tsc + vite build + build:server + build:cli + build:lib all green)
npm run test:unit         → PASS
npm run test:integration  → PASS: Test Suites: 70 passed, 70 total / Tests: 747 passed, 747 total
npm run test:e2e (playwright) → 395/419 passed locally (12 failures + 1 flaky, all in specs
   untouched by this diff — assistant-modal, evaluation-runner, navigation, quick-run-agent-preference,
   skills, test-case-runs — reproduced only under heavy concurrent CPU load from sibling worktree test
   runs sharing this dev box; none reference judge.ts / JudgeResponse / the new integration test)
npm audit --audit-level=high → pre-existing findings in transitive deps only (js-yaml, nanoid, postcss,
   react-router, undici, vite); no package.json/package-lock.json changes in this diff, same as main
```

**New integration test output** (the regression guard for this fix), from the `test:integration` run above:

```
PASS tests/integration/server/routes/judge.integration.test.ts
  Judge Route Integration Tests
    ✓ should loudly flag the mock judge: structured warning field + banner at the TOP of reasoning
```

That test asserts, against a real HTTP round-trip to `POST /api/judge` with `modelId: 'demo-model'`:
```ts
expect(typeof data.warning).toBe('string');
expect(data.warning).toMatch(/MOCK_JUDGE/);
expect(data.llmJudgeReasoning).toMatch(/^\s*⚠\uFE0F?\s*\*\*MOCK JUDGE/);
expect(data.llmJudgeReasoning.indexOf('MOCK JUDGE'))
  .toBeLessThan(data.llmJudgeReasoning.indexOf('Mock Evaluation Result'));
```

**Independent confirmation from the e2e run's server logs** (real HTTP server, real `demo`-provider judge
calls triggered by unrelated e2e specs that happen to exercise `/api/judge` with a demo model — proves the
`console.warn` fires in a live server, not just in the new test):

```
[WebServer] ⚠️  [agent-health] MOCK JUDGE IN USE: judge model "mock://demo-model" resolves to the "demo"
provider — scores are NOT real (no LLM is called). Pin a real judge model (e.g. --judge-model <bedrock-id>
or evaluator.inferenceConfig) to get a real verdict.
```
(This line appeared 5 separate times across different e2e specs during the same local run, confirming the
warning is unconditional/always-on as designed, not gated behind a debug flag.)

**DCO / CHANGELOG**: both commits carry `Signed-off-by: Megha Goyal <goyamegh@amazon.com>`; `CHANGELOG.md`
`## [Unreleased]` has the entry linked to this PR.

goyamegh:

Ran an adversarial review of the diff (correctness/regression risk + over-engineering focus). Findings, severity, and disposition:

| # | Finding | Severity | Disposition |
|---|---|---|---|
| 1 | `JudgeResponse.warning` is added to the type, but the client wrapper `callBedrockJudge()` ([services/evaluation/bedrockJudge.ts](https://github.com/opensearch-project/agent-health/blob/main/services/evaluation/bedrockJudge.ts)) reshapes the raw `/api/judge` JSON into its own `JudgeResult` and forwards `rawResponse`/`extraFields`/`judgeDebug` but silently dropped `warning` one hop after the wire — nothing downstream of that wrapper could ever see it. | Medium | **Applied** — added `warning?: string` to `JudgeResult` and forwarded `result.warning` in the return, matching how the existing debug breadcrumbs are forwarded. Added unit tests (`tests/unit/services/evaluation/bedrockJudge.test.ts`) asserting forward + real-judge-omits-it. |
| 2 | The reasoning banner uses markdown (`**bold**`, `---`) that some surfaces render as literal punctuation rather than styled text (HTML report export, QuickRunModal print) — "loud" is markdown-surface-dependent. | Medium | **Rejected (this PR)** — the plain-text banner phrase "⚠️ MOCK JUDGE — NOT A REAL EVALUATION ⚠️" is still prominently visible even unstyled (arguably more visible with literal asterisks), and it's now at the *top* of the string on every surface regardless of markdown support — which is the actual regression this PR fixes (banner used to be a bottom footnote). Auditing every downstream renderer for markdown consistency is a separate, larger concern than "make the mock judge loud." |
| 3 | Unconditional `console.warn` fires on every mock-judge call; call sites that loop `/api/judge` (e.g. the skills grader) could spray repeated warnings. | Low | **Rejected** — intentional by design (the whole point is "impossible to miss"); this is a demo/dev-only code path (never hit by a real judge), and adding rate-limiting/dedup state is disproportionate complexity for a `fix:`-scoped PR. Worth revisiting only if it's reported as noisy in practice. |
| 4 | No test asserted the new `console.warn` behavior itself — a regression could silently drop the log-side warning while the integration test (which only checks the HTTP response body) stays green. | Low | **Applied** — added a unit test in `tests/unit/server/routes/judge.test.ts` that spies on `console.warn` and asserts it fires with `"MOCK JUDGE IN USE"` + the resolved model id for the demo provider. |

Verdict from codex: no evidence of an actual parser/consumer regression from reordering the reasoning text; the two applied findings close the "warning field is dead one hop past the wire" gap and add the missing log-behavior test. Findings 2–3 are accepted as valid observations but out of scope for this fix.

**Re-verified after applying findings 1 & 4** (this dev box, `flock`-serialized against sibling worktrees):
```
npm run build:all         → PASS
npm run test:unit         → PASS (255 suites, 4591 passed / 5 pre-existing skips, 4596 total)
npm run test:integration  → PASS (70 suites, 747/747)
```
New/changed tests specifically:
```
tests/unit/services/evaluation/bedrockJudge.test.ts
  ✓ should forward the mock-judge `warning` field when the route sets it
  ✓ should leave `warning` undefined when the route response omits it (real judge)
tests/unit/server/routes/judge.test.ts
  ✓ uses mock judge when demo-model is specified (now also asserts `warning: /MOCK_JUDGE/`)
  ✓ warns loudly on the server console when the mock judge is used (unconditional, not behind a debug flag)
```

ps48 (approved):

jasonlhamazon (approved):

TackAdam (approved):

PR #395, opened 2026-08-23
fix: CLI benchmark -f polls storage on long runs instead of reporting 0/0

goyamegh:
### Description

Fixes the CLI reporting `0/0` (or `fetch failed`) on long `benchmark -f` runs even though the server finished successfully.

`runUnifiedMode`'s SSE progress loop read the run's final state exactly **once**, at the moment the `/api/storage/evaluation-runs` progress stream ended (`reader.read()` throwing or the loop exiting via `done`). On a long run — e.g. a subprocess agent (Kiro / Claude Code / a custom ops agent) taking ~10 minutes per case — a proxy idle-timeout drops the SSE connection a few minutes in, well before the run reaches a terminal state server-side. The old code treated that drop as fatal and reported `0/0` (or the raw `fetch failed` error), even though the server kept executing and persisting results independently of the client connection (this surfaced from WebIR user feedback while benchmarking a real subprocess agent).

### Changes

- **`cli/commands/benchmark.ts`** (`runUnifiedMode`) — the SSE read loop now tolerates a mid-stream disconnect once the `runId` has been captured from the `started` event (only fails hard if the connection drops *before* a `runId` exists). After the loop, if the stream ended without seeing a terminal `completed`/`cancelled` event, the CLI now **polls `/api/storage/evaluation-runs/:id`** until the run reaches a terminal state, then reports the true result read from storage — instead of trusting whatever partial `completedCount` happened to be captured before the drop.
- **`cli/utils/apiClient.ts`** — two additions: `getEvaluationRun(runId)` (`GET /api/storage/evaluation-runs/:id`, `null` on 404) and `pollEvaluationRunStatus(runId, onProgress?, timeoutMs = 3_600_000)`, which wraps the existing `pollUntilTerminal` helper (same fallback mechanism `pollRunStatus` already uses for the benchmark/execute path) with a deliberately generous default timeout — subprocess agents can run many minutes per test case.

### Testing

- **Unit** (`tests/unit/cli/utils/apiClient.test.ts`): `getEvaluationRun` returns the run on 200 / `null` on 404; `pollEvaluationRunStatus` keeps polling past a `running` read until the run is `completed` (fake timers, asserts the second fetch happens), and returns immediately when the first read is already terminal.
- Full local suite run: 255/255 unit suites (4592/4597 tests, 5 skipped) and 70/70 integration suites (746/746 tests) green, including the new/existing `cli/commands/benchmark.test.ts` and `cli/utils/apiClient.test.ts`.

### Check List


goyamegh:
## Verification proof

**1. Targeted unit tests for the new polling logic** (`tests/unit/cli/utils/apiClient.test.ts`):

```
$ npx jest tests/unit/cli/utils/apiClient.test.ts -t "getEvaluationRun|pollEvaluationRunStatus" --verbose

    getEvaluationRun
      ✓ returns the run on 200 (4 ms)
      ✓ returns null on 404 (1 ms)
    pollEvaluationRunStatus
      ✓ keeps polling storage until the run is terminal (SSE-dropped-mid-run fallback) (4 ms)
      ✓ returns immediately when the first read is already terminal (1 ms)

Tests:       98 skipped, 4 passed, 102 total
```

**2. Full local suite** (`npm run build:all && npm run test:all`, rebased onto `main`@`ae4317f8`):

```
Unit:        Test Suites: 255 passed, 255 total
             Tests:       5 skipped, 4592 passed, 4597 total
Integration: Test Suites: 70 passed, 70 total
             Tests:       746 passed, 746 total
```
(includes `tests/unit/cli/commands/benchmark.test.ts` and `tests/unit/cli/utils/apiClient.test.ts`, both green)

E2E: 404 passed, 2 flaky-but-passed-on-retry (unrelated `Configure Run` dialog / test-case-runs navigation timing), 2 genuine failures in `tests/e2e/skills.spec.ts` (Skills Evaluator Page) — root-caused to this dev box's `~/.claude/skills` (user-home-scope) containing unrelated personal skills that get merged into the Skills dropdown the test scans (`server/routes/skills.ts` intentionally scans both user- and project-scope). Confirmed unrelated to this change: `skills.spec.ts` and the skills feature are untouched by this diff (`git diff origin/main..HEAD` only touches `cli/commands/benchmark.ts`, `cli/utils/apiClient.ts`, `tests/unit/cli/utils/apiClient.test.ts`, `CHANGELOG.md`), and this local artifact does not exist on a clean CI runner.

**3. Live end-to-end smoke test** — real server, real CLI, the happy (non-dropped-SSE) path through the modified `runUnifiedMode`:

```
$ AH_PORT=4231 node bin/cli.js benchmark -f examples/eval-files/demo.eval.js -a demo --stop-server

Agent Health - Benchmark Runner

- Connecting to server...
✔ Started server on port 4231
  Agent: demo
  Sources: 1 source(s)
  Model: demo-model
  Mode: Ad-hoc (no benchmark association)

- Starting evaluation run...
✔ Evaluation run completed (5/5 test cases)

  Results:
    ✓ Passed: 1
    ✗ Failed: 4
    Total: 5

  This was an ad-hoc run (ID: eval-run-1787522906923-xsfdydolm).
  Promote to benchmark with: -n "Benchmark Name"
```
(1/4 pass split is expected/known behavior of the demo eval fixture, not a bug — this run proves the refactored SSE loop + `ApiClient.getEvaluationRun` still report the correct terminal state end-to-end when the stream completes normally.)

Simulating an actual mid-run proxy idle-timeout against a live server isn't practical in an integration test without dedicated proxy-fault infra, so the disconnect-then-poll-until-terminal path itself is covered at the unit level (item 1 above), mocking a `running` → `completed` transition across two polls.

goyamegh:

Ran an adversarial review of the diff before considering this PR done. Findings and disposition:

|---|----------|---------|-------------|
| 1 | major | `pollUntilTerminal` returns the last fetch **unchecked** on timeout; `runUnifiedMode` unconditionally printed success, so a run still `running` after the 60-min poll budget was falsely reported completed. | **Applied** — see below. |
| 2 | major | Polling treats `failed` as terminal, but the caller never branched on `run.status`; a server-side `failed`/`cancelled` run recovered via polling was printed as a success. | **Applied** — same fix as #1. |
| 3 | major | A transient 404/null from `getEvaluationRun` aborts the recovery poll immediately (`pollUntilTerminal`'s `if (!item) return null`) instead of retrying. | **Rejected (tracked as follow-up)** — real, but the fix lives in the *shared* `pollUntilTerminal` helper used by `pollRunStatus`/`pollReportStatus` too; changing its 404-retry semantics is a bigger, riskier change than this PR's scope and would need its own test pass across all three callers. |
| 4 | major | New tests only exercise the thin `ApiClient` wrappers, not the actual SSE-drop branch inside `runUnifiedMode`. | **Partially applied** — `runUnifiedMode` itself is deliberately not unit-tested anywhere in this file today (it has "server dependencies" — see the existing comment above the `Source Composition` describe block); extracted the new failed/cancelled/timeout decision into a pure, directly unit-tested function (`resolveUnifiedRunOutcome` in the new `cli/utils/evaluationRunOutcome.ts`) instead of inlining it, closing the coverage gap for the part that's tractable to test without new SSE-mocking infra. |
| 5 | minor | `getEvaluationRun`/`pollEvaluationRunStatus` called out as "near-duplicates" of the existing single-report `getReportById`/`pollReportStatus`. | **Rejected** — a run (`EvaluationRun`, with a `results` map, polled via `/api/storage/evaluation-runs/:id`) and a report (`EvaluationResult`, polled via a different endpoint) are different resource shapes with different terminality predicates; both correctly reuse the one shared generic `pollUntilTerminal` helper rather than duplicating polling logic. |

**Fix for #1/#2** (pushed as `3e9647a8`): `runUnifiedMode` now branches on the recovered run's terminal status via `resolveUnifiedRunOutcome(run, completedCount)` — `failed`/`cancelled` → `spinner.fail(...)` + exit 1 with the server's error; still `pending`/`running` after the poll timeout → an explicit "timed out waiting" message + exit 1 instead of a false success; otherwise the success banner now uses a **fresh** per-case count derived from `run.results` (the old code kept printing the stale pre-disconnect SSE `completedCount`).

**Re-verified after the fix:**
- `npx tsc --noEmit`: clean.
- `tests/unit/cli/` (13 suites / 320 tests) and `tests/integration/cli/` (8 suites / 71 tests): all green, including 6 new `resolveUnifiedRunOutcome` cases (success w/ fresh count, success w/ stale fallback when `run` is `null`, failed w/ and w/o error message, timeout from `running`, timeout from `pending`).
- Live smoke test unchanged/still correct: `AH_PORT=4232 node bin/cli.js benchmark -f examples/eval-files/demo.eval.js -a demo --stop-server` → `✔ Evaluation run completed (5/5 test cases)`, `Passed: 1 / Failed: 4 / Total: 5` (same demo-fixture split as before, now computed via the fresh-count path).

ps48 (approved):

jasonlhamazon (approved):

TackAdam (approved):

PR #396, opened 2026-08-23
fix(traces): find traces by session id / dashed UUID in Agent Traces search

goyamegh:
### Description

The Agent Traces search box returned nothing when pasting in a session id (or a bare UUID). Two root causes:

1. The OpenSearch query used `query_string: "*<text>*"`, whose parser treats `-`/`:` as operators — so a dashed UUID or an ISO-timestamped `session.id` never matched.
2. It queried `attributes.session.id[.keyword]`, but the Data Prepper `otel-v1-apm-span-*` schema stores the attribute as `span.attributes.session@id` (`@` for dots), so the field was never hit on those clusters.

### Changes

- **`server/services/tracesService.ts`** — replaced the `query_string` text-search clause with schema-agnostic, case-insensitive `wildcard` clauses across `span.attributes.session@id` + `attributes.session.id[.keyword]` + `traceId` + `serviceName`, plus a `match` on span name. Reserved characters (dashes/colons) in the query are now matched literally instead of being parsed as query-string operators. Applied the same field fix to the exact `sessionId` lookup and to Strategy D correlation (the run-report Traces tab), both of which were silently empty on Data Prepper clusters for the same field-name reason.
- **`components/traces/AgentTracesPage.tsx`** — removed a redundant client-side re-filter (on `traceId`/`rootSpanName`/`serviceName` only) that was dropping the server's `session.id` matches after the server query already found them.

### Tests

- **Unit** (`tests/unit/server/services/tracesService.test.ts`): asserts the new query shape — no `query_string`, `wildcard` clauses present across the session-id/traceId/serviceName fields for a dashed-UUID query.
- **Integration** (`tests/integration/server/services/traceTextSearch.opensearch.integration.test.ts`, new): seeds a real OpenSearch index using the Data Prepper `otel-v1-apm-span-*` span shape (`span.attributes.session@id`) and confirms `fetchTraces` finds it by bare UUID substring, by the full timestamped `session.id`, by `serviceName` substring, and via the exact `sessionId` param — plus confirms a dashed query no longer errors. Skips automatically when no local OpenSearch is reachable (matches the existing `*.opensearch.integration.test.ts` convention), so file-mode/local runs without a cluster stay green.

### Check List


goyamegh:
## Verification proof

**Local checks** (`agent-health-trace-search` worktree, rebased onto `main`@`ae4317f8`), isolated ports (`AH_PORT=4221`/`AH_DEV_PORT=4220`), run recovery disabled:

```
npm run build:all         → PASS (tsc + vite build + build:server + build:cli + build:lib, and tsc --noEmit clean)
npm run test:unit         → PASS: Test Suites: 255 passed, 255 total / Tests: 5 skipped, 4589 passed, 4594 total
npm run test:integration  → PASS: Test Suites: 71 passed, 71 total / Tests: 751 passed, 751 total
npm run test:e2e (playwright) → 404 passed, 2 failed, 2 flaky, 11 skipped (10.8m). The 4 non-green results are in
   skills.spec.ts / evals3-benchmark-runs.spec.ts / test-case-runs.spec.ts — unrelated to traces/session search;
   reproduced only under this shared dev box's heavy concurrent CPU load from other parallel worktree test runs.
   No traces/session-id spec is affected.
npm audit --audit-level=high → 14 high findings, all pre-existing transitive deps (js-yaml, nanoid, postcss,
   react-router, undici, vite). This diff makes zero package.json/package-lock.json changes, so the finding
   set is identical to main's current baseline.
```

**The new real-cluster integration test genuinely runs and passes** (not just green-by-skipping). While verifying this locally I found the test's own `it`-vs-`it.skip` selection was decided at Jest's synchronous test-collection time — before the `beforeAll` that probes cluster availability had run — so all 5 assertions were **silently registering as skipped forever**, even against a live cluster. Fixed in a follow-up commit (always register a real `it`, bail out at runtime once availability is known) and confirmed against a real local OpenSearch:

```
PASS tests/integration/server/services/traceTextSearch.opensearch.integration.test.ts
  trace text-search against real OpenSearch (Data Prepper span schema)
    ✓ finds spans by the BARE UUID substring (the original failing case) (120 ms)
    ✓ finds spans by the full timestamped session.id (46 ms)
    ✓ a dashed query does not error out (query_string would have) (23 ms)
    ✓ still matches on serviceName substring (12 ms)
    ✓ exact sessionId param matches span.attributes.session@id (Strategy D / run-report path) (6 ms)

Tests:       5 passed, 5 total
```

That test seeds a Data Prepper `otel-v1-apm-span-*`-shaped span (`span.attributes.session@id` — the schema this fix targets) into a real OpenSearch index and drives it through the actual `fetchTraces()` code path, i.e. this is the fix working end-to-end, not a mocked query-shape assertion.

**Unit test** asserting the query shape sent to OpenSearch for a dashed-UUID search (no `query_string`; `wildcard` clauses across the session-id/traceId/serviceName fields, `case_insensitive: true`):

```
PASS tests/unit/server/services/tracesService.test.ts
  ✓ Strategy D: correlates agents[].sessionId on attributes.session.id, unioned with the window (#313)
  ✓ correlates sessionId via attributes.session.id (#296)
  ✓ text search matches session.id as a case-insensitive substring (dashes/colons literal)
```

**DCO / CHANGELOG**: all 3 commits on this branch carry `Signed-off-by: Megha Goyal <goyamegh@amazon.com>`; `CHANGELOG.md` `## [Unreleased] → ### Fixed` has the entry linked to this PR.

goyamegh:
## Adversarial review (codex / gpt-5.4, different model family)

Per team standard, ran a self-contained adversarial review on this PR's diff before/alongside CI, focused on OpenSearch `wildcard`-query semantics, query performance/DoS risk, and the Jest integration-test skip-timing bug fix.

**Findings (severity → applied/rejected):**

1. `HIGH` — dropping generic `attributes.*` search (old `query_string` searched `name` + any `attributes.*` field; new code only searches 5 named fields) is a real breadth reduction. **Rejected** — out of scope for this session-id/UUID-specific bug; not covered by any existing test either way; reintroducing generic attribute search via `query_string`/`simple_query_string` would reintroduce that exact parser's dash-mishandling for the attribute values (model ids, etc.) most likely to contain dashes — the same bug this PR fixes. Worth a dedicated follow-up if generic attribute search turns out to matter in practice.
2. `HIGH` — `wildcard` query against the analyzed (non-`.keyword`) `attributes.session.id` field may be semantically wrong/a no-op on `text`-mapped fields. **Verified empirically** against a live OpenSearch: a `wildcard` query for the full literal UUID pattern against an analyzed `text` field with a hyphenated value returns **0 hits** (the standard analyzer tokenizes on `-`, so the literal substring never matches a single indexed token) — confirms the clause is inert (not a false-positive risk) for the primary use case this PR targets. **Rejected removing it** — matches this file's pre-existing defensive multi-schema `term`-query pattern (already present before this PR) of including both the plain and `.keyword` field variants; removing it would be a stylistic cleanup, not a correctness fix.
3. `HIGH` — leading-wildcard (`*x*`) queries on high-cardinality keyword fields (`traceId`, `session.id`), re-fired on every debounced keystroke with no minimum length, is a legitimate latency/cost risk (full terms-dictionary scan). **Applied** — gated the wildcard/id clauses on `textSearch.trim().length >= 3`; the cheap, index-backed `match` on span `name` still runs at any length. New unit test asserts a 2-char query emits no `wildcard` clauses.
4. `HIGH` — the integration test's "cluster unreachable ⇒ skip" path would also silently apply in CI, masking a real infra regression as green. **Applied** — confirmed via `.github/workflows/ci.yml` that the `integration-tests` job provisions a real OpenSearch service container and already blocks on it reporting healthy before any test runs, so on CI "unreachable" only ever means the promised cluster died mid-job. `beforeAll` now throws when `process.env.CI` is set and the cluster is unreachable; local/dev without a cluster still degrades gracefully. Verified all three paths directly (reachable → 5/5 real passes; `CI=1` + unreachable → hard failure; no `CI` + unreachable → graceful local pass-with-warning).
5. `MED` — `match` on `name` defaults to `OR` across multi-word queries vs. the old `query_string`'s `default_operator: 'AND'`. **Rejected** — intentional, more-permissive behavior for multi-word name search, not a correctness bug (both are inside an outer `should`/`minimum_should_match:1`, so this can only add recall, not drop it).
6. `MED` — the new integration test could cover more edge cases (literal `*`/`?` input, very short queries, the unavailable-cluster branch). **Partially applied** — added a unit test for the length-guard; did not add further integration-test cases (diminishing returns given the unit test now covers the length-guard's query-shape directly, and the guard makes moot the case of a 1-2 char literal `*`/`?` reaching the wildcard clauses at all).
7. `LOW` — keeping `textSearch` in `must` (ANDed) while ID-correlation strategies go to `should` when `useUnion` is true. Reviewer explicitly called this a **non-issue** (consistent with the pre-existing `serviceName` filter pattern in the same function) — no action.
8. `LOW` — 6-clause `should` array is scope creep. Same disposition as #1 — **rejected**, tracked as a design note rather than expanded in this PR.

Local re-verification after applying #3/#4: `tsc --noEmit` clean; `npm run test:unit` 255 suites / 4590 tests all green; `npm run test:integration` with `CI=1` set (matching CI's real env) 71 suites / 751 tests all green, confirming the new hard-fail branch is inert on the happy path.

goyamegh:
## Self-caught regression + fix

The commit applying codex finding #4 (CI hard-fail on unreachable cluster) broke `release-rehearsal (npm test exactly as Release Workflow runs it)` — confirmed from the actual failed run: all 5 tests in `traceTextSearch.opensearch.integration.test.ts` failed. Root cause: I assumed `process.env.CI` reliably signals "a cluster was provisioned" because that's true for the dedicated `integration-tests` job (which health-checks a real OpenSearch service container before any test runs) — but GitHub Actions sets `CI=true` for **every** job, including `release-rehearsal`, which runs the full `npm test` on a plain runner with no OpenSearch service at all.

Reverted that specific piece back to unconditional skip-with-warning (commit `53b522b2`) — there's no environment signal visible to the test file that distinguishes "cluster specifically promised for this job" from "just some CI job" without adding a dedicated marker env var to `ci.yml`'s `integration-tests` job, which is out of scope for this PR. The other codex-driven fix (wildcard min-length guard) is untouched and unaffected.

Verified locally against both the passing and the exact failing scenario before pushing the revert:
```
CI=1, cluster reachable (matches integration-tests job)              → 5/5 real passes
CI=1, no AH_PORT, cluster unreachable (matches release-rehearsal exactly) → 5/5 pass-with-warning (was 5/5 hard failures before the revert)
```
`npm run test:unit` (255/255 suites) and `npm run test:integration` with `CI=1` (71/71 suites) both still fully green after the revert.

Net effect vs. the original file (pre-this-PR): the "always silently `it.skip()` at collection time, forever, even with a live cluster" bug is still fixed — these 5 tests now genuinely run whenever a cluster is reachable, on any job/machine. Just no CI-only hard-fail-on-missing-cluster, since I couldn't do that safely without touching shared CI workflow config.

ps48 (approved):

jasonlhamazon (approved):

TackAdam (approved):

PR #398, opened 2026-08-24
feat(comparison): deep-dive records an on-the-fly A-vs-B chart + suggested next experiments

goyamegh:
## Summary

The Comparison Deep-Dive agent (in-process pi agent that inspects both runs' real OTel spans/logs and writes a grounded markdown narrative) now **also** records, in the same LLM pass — no extra round-trip:

1. **A small A-vs-B "compare-bars" chart** of real numeric dimensions it found worth surfacing (tool calls, retries, tokens, error counts, duration). Each row is normalized to its own scale since units differ per dimension; the printed number (not bar length) carries precision.
2. **1-4 concrete "suggested next experiments"** — follow-up test-case ideas grounded in what it actually observed in that specific run pair (e.g. "force a mid-task tool failure" because one run recovered from a retry and the other never hit that path), with the same `[label](span:<runId>:<spanId>)` citation support as the narrative.

Both are optional and recorded atomically via a single new tool (`record_deepdive_extras`) the agent calls as a structured side effect — never parsed out of the free-form markdown. Either or both are omitted (no UI section rendered) when the agent found nothing worth surfacing for that part.

`POST /api/comparison/deep-dive` gains optional `chart` / `experiments` fields alongside the existing `markdown`/`modelId`/`durationMs`.

## Why

The deep-dive already grounds a narrative in real trace data; this surfaces some of that same grounded evidence as a quick-scan visual and turns "what's different" into "what should I try next" — both directly informed by what the agent actually found in this pair, not generic advice.

## Design notes / trade-offs

- **One combined tool, not two.** An earlier version had `record_metric_chart` + `record_experiment_suggestions` as separate tools; merged into one `record_deepdive_extras` (both fields optional) after review — avoids inconsistency between two independently-ordered tool calls and halves the prompt/schema surface.
- **Chart values are LLM tool-call output, not independently re-verified against the spans cited** — same trust model the existing narrative already has for its span-grounded prose claims (this isn't a new risk this PR introduces). Numeric fields are schema-constrained to `minimum: 0` (all recorded dimensions — tool calls, retries, tokens, error counts, duration — are naturally non-negative), and the renderer defensively clamps non-finite/negative values so a malformed value degrades to a dash instead of a broken layout.
- **Custom "compare-bars" component instead of Recharts** — deliberate: each row needs independent per-row scaling (units differ: calls vs. seconds vs. tokens), which doesn't fit a shared-axis grouped bar chart. Kept as plain divs, no new charting dependency.

## Testing

- Unit: `tests/unit/server/services/comparisonTraceTools.test.ts` (tool registration, capture semantics, partial/omitted-field calls), `tests/unit/server/services/comparisonDeepDiveService.test.ts` (SYSTEM_PROMPT content).
- e2e (Playwright): `tests/e2e/comparison-deep-dive-chart-experiments.spec.ts` — renders both new sections when the API returns them, renders neither when omitted (backward compat), and a defensive-guard case asserting a malformed chart value (null / negative) never produces a broken `NaN%`/negative CSS width.
- Full suite green: `npm run build:all`, `npm run test:unit` (257 suites / 4612 tests), `npm run test:integration` (71 suites / 750 tests, including the deep-dive route's deterministic 400/404 cases), plus all pre-existing comparison e2e specs.
- `npm audit --audit-level=high`: pre-existing transitive findings only (undici via the optional `@earendil-works/pi-coding-agent` dep, vite) — already present on `main`, not introduced by this change.

## Adversarial review (codex_review, gpt-5.4)

Ran before opening this PR. Applied: merged the two recorder tools into one atomic call; added `minimum: 0` schema constraint + frontend defensive clamping for malformed chart values (with a regression e2e test). Not applied: restructuring to a single final-JSON-envelope response (would break from this codebase's established markdown-final-answer + separate-tool-for-structured-output convention, used elsewhere for the same optional-pi-SDK integration); adding explicit "nothing found" messaging when the agent skips a section (existing app convention already hides unrecorded optional data silently, e.g. `MetricComparisonPanel`); live-model verification of narrative-quality regression risk (flagged as a manual follow-up, not blocking — no live LLM call was run as part of this automated review).

## CHANGELOG

Updated under `## [Unreleased] > Added`.


TackAdam (approved):

TackAdam (approved):

goyamegh:
UX follow-up per owner feedback on the deep-dive panel: removed the redundant "Performance & Outcome" bars block (its bare `100 pts`/`50 pts` Score row misread as a case count in a multi-hundred-case comparison), replaced it with a compact `Score: 100% vs 50% · Duration: 36.9s vs 29.2s · Tools: 3 vs 3` line in the panel header, and labeled the judge score in the generation prompt so the narrative itself never writes a bare `N/N` going forward (regenerate-path only).

PR #399, opened 2026-08-24
fix: stabilize file-mode benchmark imports

ashwin-pc:
## Summary
- reuse semantically identical JSON/directory-imported test cases and link named CLI runs into one benchmark history
- exclude co-located evaluation-run documents from benchmark lists
- show Overview for non-empty file storage and guard legacy benchmarks without testCaseIds

## Validation
- `npx tsc --noEmit`
- targeted unit and integration Jest suites
- `npm run build:all`

Signed-off-by: ashwin pc <ashwinpc@amazon.com>


goyamegh:

Reviewed the full diff (18 files) against the PR's stated goals (dedupe repeated CLI/file imports into one benchmark, hide co-located `evaluation-run` docs from benchmark listings, fix the file-mode Overview/onboarding gate, guard legacy benchmarks without `testCaseIds`, and route all benchmark "Run" actions through the unified `/api/storage/evaluation-runs` runner). Framed as suggestions, not blockers — flagging the ones I'd want a second look at before merge.

|---|---|---|
| Major | `reuseOrCreateImportedTestCases()` (`services/sourceResolver.ts`) trusts `definition.sourceHash` if present, so a stale/carried-over `sourceHash` on an edited import could merge into the old test-case doc instead of creating a new one — a bit at odds with the "changed definitions become new documents" intent. | Consider always recomputing the hash from content rather than trusting a caller-supplied `sourceHash`, or document why it's trusted. |
| Major | Same function calls `storage.testCases.getAll({ size: 10000 })` and does a linear/canonical-JSON scan on every file import — works today but degrades (and silently misses matches) past 10k test cases, and repeated large imports re-scan the whole corpus each time. | Worth a follow-up if test-case counts are expected to grow past ~10k; maybe an indexed lookup by `(name, category, sourceHash)` instead of full-table scan. |
| Major | Benchmark linking in `server/routes/storage/evaluationRuns.ts` (`testCaseIds` update, then `addRun` after run completion) has no optimistic-concurrency guard — two concurrent runs against the same benchmark could race on the read-modify-write. Also, the evaluation-run is marked `completed` *before* `addRun()`, so if `addRun` throws after that, the run persists as completed but is missing from the benchmark's embedded history (no retry/compensation). | Consider ordering `addRun` before finalizing the run's terminal status, or adding a version/CAS check on the benchmark update, plus a test for the failure-after-completion case. |
| Major | The OpenSearch `must_not: docType.keyword: evaluation-run` exclusion assumes `docType` is mapped as (or has) a `.keyword` subfield. On older indices without that mapping the clause could silently no-op and evaluation-run docs would keep leaking into benchmark lists. | Worth confirming the index mapping guarantees `.keyword` on `docType` for all existing deployments, or add a mapping migration/test against a legacy-shaped index. |
| Major | `useDataState.ts`'s new file-mode "hasData" heuristic only counts "benchmark with embedded runs" or "a completed evaluation run" as real data — a workspace with imported test cases/benchmarks but zero completed runs still shows onboarding. Also fetches the full benchmark list (with embedded runs) just to answer an existence check. | Minor UX nit — might be intentional, but flagging in case "imported test cases with no runs yet" should also count as "has data". |
| Minor | `sendSSE()` now swallows all write errors via a bare `catch {}` — good for surviving dead sockets, but also silently hides genuine serialization/programming bugs. | Consider a debug-level log in the catch so real bugs aren't invisible. |
| Minor | The rewritten `executeBenchmarkRun` client adapter fabricates `currentTestCaseIndex` from `startedCount` and `currentTestCase.name` from the raw ID — fine for the single-case tests added, but not obviously equivalent to the old contract under concurrent/out-of-order progress events. | Consider a concurrency-focused test if progress events can arrive out of order. |
| Minor | Coverage: the old client SSE edge-case unit tests were deleted/replaced with simpler ones; no test yet for stale-`sourceHash` reuse, >10k-test-case behavior, legacy (no-`docType`-mapping) OpenSearch indices, or the concurrent-benchmark-linking / post-completion `addRun`-failure paths mentioned above. | Nice-to-add alongside the above if you address them. |

Nothing here looks like it invalidates the overall approach — it's a solid stabilization pass — just a handful of edge cases around identity/reuse, concurrency, and legacy-data compatibility worth a look before merge.

ashwin-pc:
Thanks — all five findings are addressed:

- **Imported test-case identity** (`189d661`): `sourceHash` is always recomputed from canonical content. An incoming hash is treated as optional verification metadata only (stale ones are logged), so a stale hash can no longer drive reuse of an unrelated document.
- **Benchmark linking consistency** (`f4de381`): the benchmark projection is linked **before** terminal finalization, and linking is idempotent — file storage treats a known run id as a successful no-op and the OpenSearch update script checks run ids before appending. An interruption can no longer leave a completed run absent from benchmark history, and retries can't duplicate it.
- **Legacy OpenSearch mapping** (`7ba6ddb`): `docType` filtering now uses a shared mapping-compatible discriminator covering both the current keyword subfield and legacy text-only indices, applied to evaluation-run inclusion and benchmark exclusion, with the compatibility contract documented.
- **Import lookup scaling** (`9296be8`): the 10,000-case scan limit no longer fails silently — an incomplete scan raises an actionable error rather than risking duplicate test cases. Failure was chosen over a warning because callers already propagate import errors. Indexed lookup is deliberately deferred until deployments approach the limit.
- **Onboarding semantics** (`2c9c462`): added an explicit three-state model (`onboarding | ready-to-run | dashboard`). Imported benchmarks/test cases with no completed runs now render a distinct "ready to run" state instead of new-user onboarding; genuinely empty installs and cluster behavior are unchanged.

The swallowed SSE write errors are now logged at debug level.

goyamegh (approved):
Tested on the shared tunnel (mirror-integrated); approving per verification.

PR #400, opened 2026-08-24
Responsive mobile layout with full capability parity

ashwin-pc:
## Approach

A parity retrofit rather than a separate mobile application: the existing route/component DOM is reused at every viewport. Below `lg`, the existing sidebar becomes an off-canvas drawer with a slim top bar; shared responsive CSS compacts gutters, makes controls touch-sized, and constrains Radix overlays to the viewport.

## Table and dense-content patterns

1. **Wide data tables:** preserve every desktop column and sort header in a horizontal scroll region, with the first identity column sticky. Nothing is hidden or replaced on mobile.
2. **Dense inspectors:** reuse the same resizable panel DOM and stack the panels at phone widths. Tabs/action strips scroll horizontally, and long trajectory, judge, code, and diagnostic content wraps or scrolls.

## Parity checklist summary

- Overview: all onboarding and dashboard actions visible; cards reflow.
- Evaluation benchmarks: search, agent/time filters, import/new/sample/refresh actions in swipeable action strip; all columns/actions in table scroll.
- Test cases + detail: search, benchmark/label filters, view/group controls, import/new/sample actions reachable; all row columns and run/edit actions retained.
- Evaluation runs + detail/inspector: filters, view, selection/compare, sort headers and all columns retained; inspector list/detail stack with all tabs.
- Legacy benchmarks: all cards and run/edit actions retained; also guards legacy records without `testCaseIds`.
- Evaluators + edit: cards/forms reflow without a separate route.
- Agent traces: search/filter/agent/time/refresh toolbar scrolls; all trace columns retained with sticky Start Time.
- Settings: panels/forms reflow and all endpoint controls remain reachable.
- Compare: run search, label/status filters and re-run action scroll; wide comparison matrices retain all runs/columns.

## Screenshots and validation

Captured each route at 375x812 and 1440x1000. Local artifacts are named `<route>-mobile.png` / `<route>-desktop.png` under `.pi/web/artifacts/agent-health-mobile/` (not committed). Production frontend and server builds and `npx tsc --noEmit` pass.

## Screenshots

All captured at **375×812** (mobile) and **1440×1000** (desktop regression) against real local data. Full set available on request; key views below.

| Mobile 375×812 | What it shows |
|---|---|
|  | **Benchmarks list** — hamburger nav, sticky Name column, all columns reachable via horizontal scroll |
|  | **Benchmark detail** — stacked panels, clean section headers, compact per-case facts |
|  | **Dashboard** — Performance Trends chart renders at mobile width (fixed-height container) |
|  | **Recent Evaluation Runs** — readable rows, sticky-first-column scroll (was overlapping text) |
|  | **Test-case detail** — split panes stack; full content (5,600px+) scrollable (was trapped behind `overflow:hidden`) |
|  | **Run inspector** — trajectory/judge output wraps and scrolls |
|  | **Agent traces** — filter toolbar swipeable, full table retained |

**Desktop regression** (unchanged layout):


goyamegh:

Reviewed the full diff (12 files, mostly CSS/layout) against the PR's stated "parity retrofit" approach (reuse existing DOM at every viewport; global CSS under `.mobile-responsive-content` handles compaction/scrolling/stacking; custom off-canvas drawer for the sidebar below `lg`).

|---|---|---|
| Blocker (worth resolving before merge) | The mobile nav drawer in `Layout.tsx` (fixed `div` + `-translate-x-full`/`translate-x-0` + backdrop `<button>`) has no focus trap, no `Escape`-to-close, no `aria-modal`/`inert`, and doesn't lock body scroll — and the closed drawer stays tabbable off-screen. `components/ui/sidebar.tsx` already ships a Radix `Sheet`-based mobile path (`collapsible="offcanvas"`) with those a11y properties built in; `Layout.tsx` currently passes `collapsible="none"`, which is why that path isn't used here. | Consider switching to the existing `Sheet`-based mobile sidebar instead of a hand-rolled drawer, or add the missing focus-trap/Escape/scroll-lock behavior if there's a reason to keep the custom implementation. |
| Major | The new global CSS in `index.css` overrides Tailwind utility classes by selector name inside `.mobile-responsive-content` (e.g. `.p-6`, `.grid-cols-2/.3/.6`, `.flex.items-center.justify-between`) — since that class wraps the whole app shell, any unrelated screen using those same utility classes (e.g. `SettingsPage.tsx`, `CodingAgentsPage.tsx`) picks up the retrofit rules "for free," with no scoping to the pages this PR is actually fixing. | Given several page components in this same PR already use component-scoped `max-md:`/`sm:` Tailwind variants, worth considering consolidating on that pattern rather than mixing both — the global-selector approach is a wider blast radius and higher long-term maintenance cost. |
| Major | `[role="dialog"]` and `[data-radix-popper-content-wrapper]` viewport-clamp rules in `index.css` are unscoped (apply globally, not gated to `.mobile-responsive-content`) at any width under 1024px — this will also clamp intentionally large dialogs (e.g. the full-screen dialog component, settings panels) at tablet widths. | Consider scoping these under `.mobile-responsive-content` too, or confirming this is intentional for all dialogs app-wide. |
| Major | Several new CSS rules key off structural/library-internal selectors (`nav[aria-label="Breadcrumb"] + div`, `[data-panel-group-direction="horizontal"]`, `[data-panel-resize-handle-enabled]`, `[role="tablist"]`) — these are one markup refactor or `react-resizable-panels`/Radix version bump away from silently breaking, and the panel-stacking rule isn't scoped to only the panels this PR intends to change. | Not asking for a rewrite, just flagging as a fragility risk worth a comment/test to catch regressions early. |
| Minor | `Dashboard.tsx`'s "Recent Runs" list is a `<button>`/`<div>` CSS-grid, not a `<table>`, but the sticky-first-column CSS rule targets `table > tbody > tr > :first-child` — so the claimed "sticky identity column while scrolling" doesn't actually apply to this particular list even though it got the `min-w-[720px]` + `overflow-x-auto` treatment. | Might be worth a quick manual check on a narrow viewport to confirm the row-identity-while-scrolling behavior actually holds there. |
| Minor | Coverage: `tests/e2e/mobile-responsive.spec.ts` covers drawer open/close and a root-scrollWidth bound check, which is good for the shell, but the PR's "full capability parity" claim (every table column, filter, and action reachable per the checklist in the description) isn't exercised by an automated test yet — it's currently verified by the screenshots/manual pass described in the PR body. | Consider one or two e2e specs that actually scroll a wide table at mobile width and assert a specific column/action is reachable, for a couple of the higher-traffic pages (benchmarks list, run inspector). |

Overall this is a solid pass at a real problem (several dense pages were genuinely broken on narrow viewports per the before/after screenshots), and none of the above invalidates the direction — mainly flagging the nav accessibility gap as worth a look given there's already an accessible primitive in the codebase, plus the global-CSS blast radius as something to keep an eye on as more pages get touched.

goyamegh (approved):
Tested on the ah-main tunnel (phone + desktop). Codex adversarial review posted earlier — drawer a11y noted as follow-up.

ashwin-pc:
Thanks for reviewing this after merge — both findings are being addressed in follow-up PR #445 rather than by amending merged code:

- **Accessible mobile navigation** (`5375a2b`): the custom off-canvas drawer is replaced with the repo's existing Radix `Sheet`, so focus trapping, Escape-to-close, modal semantics, background inertness, and scroll locking come from the primitive rather than bespoke code; the closed drawer is no longer keyboard-focusable. Covered by a11y tests (Escape closes, focus trapped while open, closed drawer not tabbable) and re-verified visually at 375×812.
- **Global responsive CSS blast radius** (`a8c9f78`, `a9ad8dc`): the widest-reach rules — dialog/popper overlays first — are migrated to component-scoped responsive utilities, with the global rules deleted as each is moved, plus mobile coverage for direct dialog consumers. Migration is incremental rather than a retrofit rewrite; any rule judged too risky to move in this pass is listed as remaining debt instead of being silently retained.

For context on how this happened: the retrofit was constrained to full capability parity with minimal divergence (no mobile-only components), and broad global rules were the shortest path to reflowing every page — including, as you spotted, rules reaching into third-party internals no page opted into. That was the predictable debt of the approach, and scoping it is the right correction.

PR #403, opened 2026-08-24
fix(storage): benchmark detail API must not return evaluation-run documents

goyamegh:
## What & why

Benchmarks and evaluation-runs share one storage index/directory, discriminated by `docType`. The benchmark **detail** op (`getById`) read by id with no `docType` check, so a CLI/SDK eval-run (`docType: 'evaluation-run'`, created via `POST /api/storage/evaluation-runs`) leaked into the benchmark detail surface: `/evaluations/benchmarks/<eval-run-id>/runs` rendered an eval-run as an empty benchmark ("No runs yet"). No benchmark doc is ever created for these — they are eval-runs, correctly shown on the Evaluations › Runs page, leaking into a benchmark surface because both doc types share one storage index/dir.

**List-path filtering is covered by #399 — this PR now only guards the single-benchmark/detail endpoint** (narrowed per maintainer decision to avoid overlapping with #399's independent, broader import-stabilization work, which already implements the same `getAll` exclusion on both backends).

## Changes

- `server/adapters/opensearch/StorageModule.ts`: `benchmarks.getById` returns `null` (→ route 404s) when the fetched doc's `docType` is `evaluation-run`.
- `server/adapters/file/StorageModule.ts`: same guard for the file-storage backend.
- `CHANGELOG.md`: `## [Unreleased]` → `### Fixed` entry, scoped to the detail endpoint.

## Testing

- **Unit**: regression cases in both `tests/unit/server/adapters/file/StorageModule.test.ts` and `tests/unit/server/adapters/opensearch/StorageModule.test.ts` covering `getById` (returns `null` for an eval-run id; still returns a real benchmark).
- **Integration**: [tests/integration/server/routes/storage/benchmarksExcludeEvalRuns.integration.test.ts](https://github.com/opensearch-project/agent-health/blob/main/tests/integration/server/routes/storage/benchmarksExcludeEvalRuns.integration.test.ts) seeds a real eval-run + a real benchmark and asserts `GET /api/storage/benchmarks/:evalRunId` 404s while a real benchmark detail still resolves; cleans up everything it creates. Also manually re-verified against a live server (non-default port) hitting real storage: seeded eval-run → `GET .../benchmarks/<evalRunId>` → 404; real benchmark → 200.
- `npm run build:all` + `npm run test:unit`: green — 257/257 suites, 4635 passed (5 skipped), 0 failed.
- e2e not re-run locally for this narrowing (CI covers it); unaffected surface (detail-route 404 vs list rendering).

Signed-off-by: Megha Goyal <goyamegh@amazon.com>


goyamegh:
## Verification proof

**Red → green story (the two bugs this fixes):**

Before the fix, `benchmarks.getAll()` used `match_all` and `benchmarks.getById()` had no `docType` filter — any `docType: 'evaluation-run'` doc (same index/dir as benchmarks) was returned as if it were a benchmark. After the fix, both ops filter it out.

- **Integration** — `tests/integration/server/routes/storage/benchmarksExcludeEvalRuns.integration.test.ts` (booted against a real running backend `npm run dev:server`): seeds a real `evaluation-run` doc + a real `benchmark` doc, then asserts:
  1. `GET /api/storage/benchmarks` → contains the real benchmark, does **not** contain the eval-run id, and no returned item carries `docType: 'evaluation-run'` (BUG 1 — list leak).
  2. `GET /api/storage/benchmarks/<eval-run-id>` → `404` (BUG 2 — detail leak; previously rendered as an empty "No runs yet" benchmark).
  3. `GET /api/storage/benchmarks/<real-benchmark-id>` → still resolves normally.
- **Unit** — added to both `tests/unit/server/adapters/file/StorageModule.test.ts` and `tests/unit/server/adapters/opensearch/StorageModule.test.ts`: `getAll` excludes an eval-run doc from results (file backend) / the OpenSearch query carries the `bool.must_not: [{term: {docType.keyword: 'evaluation-run'}}]` clause (opensearch backend); `getById` returns `null` for an eval-run id in both backends; a real benchmark is still returned by both ops.

Reverting the two guard lines locally (dropping the `must_not` filter and the `docType` check) reproduces the pre-fix behavior against these same tests — both the new unit assertions and the integration test fail without the fix, confirming they pin the regression.

**Full suite (local, `CI=1 AH_PORT=4301 AH_DEV_PORT=4300 BENCHMARK_RUN_RECOVERY_DISABLED=1 EVALUATION_RUN_RECOVERY_DISABLED=1`):**

```
npm run build:all               → PASS
npm run test:unit                → Test Suites: all passed. Tests: 5 skipped, 4610 passed, 4615 total
npm run test:integration         → Test Suites: 72 passed, 72 total. Tests: 753 passed, 753 total
npm run test:e2e (playwright)    → 407 passed, 2 flaky (retried green), 2 failed, 11 skipped
                                    (the 2 failures are both in tests/e2e/skills.spec.ts — a known
                                     pre-existing shared-box flake unrelated to this change / not
                                     touched by this diff)
npm audit --audit-level=high     → 31 pre-existing transitive findings (postcss, react-router,
                                     undici, vite, protobufjs) — none introduced by this change
```

Coverage of the changed lines: the new unit tests exercise every line touched in `server/adapters/file/StorageModule.ts` (`FileBenchmarkOperations.getAll`/`getById`) and `server/adapters/opensearch/StorageModule.ts` (`OpenSearchBenchmarkOperations.getAll`/`getById`) — confirmed via a scoped coverage run (`jest --coverage --collectCoverageFrom` on just those two files): none of the changed line ranges (301-317 file backend, 333-372 opensearch backend) appear in the reported "Uncovered Line #s".

goyamegh:
## codex_review (gpt-5.4) — adversarial pass on this diff

Ran per standing policy before posting for approval. Findings + disposition:

|---|---------|----------|----------|
| 1 | Fix assumes every eval-run doc already has `docType: 'evaluation-run'` — could miss legacy docs without it | blocker (claimed) | **Rejected** — verified: `docType: 'evaluation-run'` has been set unconditionally by `evaluationRuns.create()` in both backends since the `EvaluationRun` architecture was introduced (#205 / `357b1683`); no eval-run doc has ever existed without it. |
| 2 | `docType.keyword` term query assumes an ES `.keyword` subfield exists; could silently no-op on other mappings | blocker (claimed) | **Rejected** — this is the pre-existing, already-proven pattern (`OpenSearchEvaluationRunOperations.list()` already filters on `docType.keyword`); not a new assumption introduced by this diff, and it works against the real ephemeral OpenSearch service in CI (`integration-tests` job, 753/753 passed). |
| 3 | Comment mentions "experiments" also sharing the index — could leak between docTypes | major (claimed) | **Rejected** — "experiments" in the comment refers to the OpenSearch **index name** (`evals_experiments`), not a coexisting docType value; no other docType exists in this index. |
| 4 | `benchmarks.delete()` / `addRun()` / `updateRun()` / `deleteRun()` still operate on a raw id with no `docType` guard — an eval-run id could be deleted/mutated through the benchmark-scoped write routes even though the read routes (this PR) now 404 for it | **major — real** | **Not applied here** (pre-existing gap, not a regression introduced by this PR; keeping this PR scoped to the reported list/detail leak). Filed as [#404](https://github.com/opensearch-project/agent-health/issues/404) for a follow-up. |
| 5 | New integration test self-skips if the backend is unreachable — could mask a real regression | minor (claimed) | **Rejected** — this is the established repo-wide integration-test convention (`checkBackend`/`backendUp`, same as e.g. the `judgeModelId` suites), and CI evidence shows it ran for real (not skipped) against the ephemeral OpenSearch service in this PR's `integration-tests` job. |
| 6 | OpenSearch unit test asserts the exact `must_not` query shape (brittle to refactors) | nit | **Rejected** — intentional: this is a security-relevant filter, and an exact-match assertion is what actually catches an accidental loosening of it. |
| 7 | Type casts (`b as { docType?: string }`) weaken typing at the fix site | nit | **Rejected** — mirrors the existing codebase convention for this exact discriminator (`OpenSearchEvaluationRunOperations.getById` already does the same cast); a broader type refactor is out of scope for this fix. |

Net: 1 of 7 findings was real (write-path guard gap) and it's pre-existing, not introduced by this diff — tracked separately rather than expanding this PR's blast radius. No code changes made as a result of this review; CI (19/19 checks) is unaffected.

TackAdam (approved):

goyamegh:
Narrowed to the detail-path guard; list path defers to #399 (per maintainer decision). Re-verified: unit `npm run test:unit` 4635 passed/0 failed, `npm run build:all` green, and manually confirmed against a live server hitting real storage — seeded eval-run → `GET /api/storage/benchmarks/<evalRunId>` returns 404; real benchmark → 200. History rewritten to a single DCO-signed commit.

Issue #320, opened 2026-06-19
Trace-mode judge runs in multiple uncoordinated places and grades a tool-call-less trajectory (verdicts flip on refresh)


uhinze:
### Summary
For a `useTraces: true` agent, evaluation results are unstable: a run first renders as **passed** and, after a page refresh, renders as **failed** (with a different matcher label). Two root causes combine to produce this:

1. **The judge-after-traces logic is implemented in three separate places**, and for a single UI "Run Test" two of them run **concurrently in different runtimes** (Node server + browser). They call the judge independently, reach different (non-deterministic) verdicts, and persist to **different fields**, so which result the UI shows depends on write/read timing.
2. **The trajectory both judges grade contains no tool calls.** When no `buildTrajectory` hook is configured, trace spans are never converted into a judge-visible trajectory; the judge instead grades the AG-UI response (final assistant message only). For a trace-only agent the judge therefore "sees no tool was invoked" and fails — even though the tool calls are present in the traces.

These should be fixed together: the judge should grade a trajectory built from the traces, and it should run exactly once through a single shared path.

---

### Symptom
Same run, two answers:

| | Matcher label | Source field | Verdict | Badge |
|---|---|---|---|---|
| First render | `judge: expected outcomes` | `llmJudgeReasoning` (synthesized) | passed (lenient) | `pending` |
| After refresh | `judge: <expected outcome text>` | `matcherResults` | failed (strict) | `failed` |

The changing matcher *label* is the tell that two different code paths produced the two results.

---

### Root cause 1 — multiple, uncoordinated judge invocations

There are three judge-after-traces implementations:

| # | Location | Triggered by | Writes |
|---|----------|--------------|--------|
| A | `services/benchmarkRunner.ts` → `startTracePollingForReportWithModule()` (`onTracesFound`) | `runSingleUseCase()` (which `POST /api/evaluate` "Run Test" calls) and benchmark `executeRun()` | `matcherResults` (claim = expected outcomes), `passFailStatus`, `metrics`, `llmJudgeReasoning`, `llmJudgeResponse`; `metricsStatus: 'ready'` |
| B | `services/evaluationRunner.ts` → `waitForTracesAndJudge()` (`onTracesFound`) | `executeEvaluationRun()` (ad-hoc / scheduled `EvaluationRun` path) | same surface as A (near-verbatim duplicate) |
[...]

PR #413, opened 2026-08-25
fix(traces): headless trace-mode judging via sessionId/window correlation; stop pollers clobbering judged verdicts

goyamegh:
## Problem

`useTraces: true` (trace-gated judging) only worked when someone happened to have the run page open in a browser. Headless, every report timed out after 10 minutes and was marked `metricsStatus: error`:

1. **The server-side trace poller correlates by runId only** (`tracePoller.ts` → `fetchTracesByRunIds`). But Claude Code spans carry only `session.id` (no `agent_health.run.id` / `gen_ai.conversation.id`), pi spans carry neither, and REST-connector reports never get a runId at all — Strategy B can never find their spans.
2. **`MAX_POLL_CEILING = 60` silently clamped** per-agent `tracePolling.maxAttempts` overrides (e.g. the documented 180-attempt config for slow OTLP→API-Gateway→cluster ingestion) back down to 10 minutes.
3. **Trace polling clobbered already-judged verdicts**: the run-inspector's recovery fan-out starts polls for *every* pending report — including transiently-pending eager-path placeholders — and the poller's timeout patch overwrote real judge verdicts with `trace_timeout` 10 minutes later (observed clobbering a live 84-case benchmark's early verdicts on 2026-08-25).
4. **Boot recovery tombstoned healthy in-flight reports**: any pending report without a runId was unconditionally marked `error` on server boot — including reports created seconds earlier by a live run (or by a *sibling* agent-health server sharing the storage cluster).

## Fix

- `tracePoller.poll()` now fetches the report and unions Strategy B (runId) with the sessionId/service-window hints from `buildJudgeAgentsHints` (Strategies C/D) — the same correlation the run-page Traces tab already uses. `runId` is optional end-to-end.
- `MAX_POLL_CEILING` 60 → 240; explicit agent overrides are honored.
- Clobber guards at the choke point: the poller stops (resolving, not erroring) when a report reached a terminal `metricsStatus` through another path, and timeout/error patches are skipped for already-judged reports (`patchErrorIfStillPending`). The inspector fan-out additionally waits a 3-minute `minPendingAgeMs` grace period.
- All poll call sites (benchmarkRunner ×2, evaluationRunner, browserRecovery) drop the runId requirement; the evaluationRunner gate keeps (gains, relative to main) the `agentConfig.useTraces` guard.
- `traceRecoveryOnBoot`: new `TRACE_RECOVERY_MIN_AGE_MS` grace window (default 15 min) — young pending reports are presumed in-flight and left alone; no-runId reports past the window are **resumed** via hint-based polling instead of tombstoned.

## Verification

- `npx tsc --noEmit` clean; `npm run build:all` green.
- 516 unit tests green across the touched suites (`tests/unit/services/traces/` 409, `traceRecoveryOnBoot` 18 incl. 3 new grace-window/no-runId-resume tests, benchmarkRunner/evaluationRunner suites).
- Live verification on the ah-main deployment: three parallel 84-case EnterpriseRAG benchmark runs with `useTraces: true` across all three previously-broken span shapes — Claude Code (session.id-only), pi (service-window-only), and a REST agent (no runId) — details in the PR comment below.


goyamegh:
## Verification (live, headless — no browser open)

Adversarial review (Codex/gpt-5.4) on the initial commit surfaced 6 findings; all addressed in the follow-up commits:

| Finding | Fix |
|---|---|
| HIGH: benchmarkRunner upstream gates still required `runId` | gates relaxed at both call sites (`benchmarkRunner.ts`) |
| HIGH: terminal-state fast-stop could hang `waitForTracesAndJudge` forever | new `PollCallbacks.onStopped`, wired to resolve |
| HIGH: no claim state → sibling pollers double-judge | poller stamps `pending → calculating` before judging (best-effort claim; storage has no CAS — window narrowed, noted in comment) |
| HIGH: window fallback fail-open → judged foreign spans | **reproduced live** (a pi eval judged an unrelated pi session's spans); filter is now fail-closed on `session.id` / eval-span `traceId` |
| MED: `trace_timeout` could overwrite a more specific `error` | `patchErrorIfStillPending` skips `ready` **and** `error` |
| MED: no-correlator attempts issued unfiltered queries | fetch skipped; attempt budget advances with a specific timeout reason |

Smoke testing the fixes surfaced **three more latent bugs**, also fixed here:
1. `lib/config/loader.ts toAgentConfig()` silently **dropped `traceServiceName` and `tracePolling`** from user configs — window correlation fell back to protocol-default service names (colliding with other emitters on a shared cluster) and poll budgets reverted to 60×10s.
2. `BaseConnector.buildTraceparentEnv()` was **never called by any connector** — `traceContext.propagateEnv` was dead config; subprocess agents spawned without W3C trace context. Now wired into `SubprocessConnector`. pi's OTel SDK adopts TRACEPARENT (verified live), giving exact Strategy-A correlation.
3. The default span→trajectory conversion **replaced rich connector trajectories with content-less span stubs** (Claude Code message content lives in OTel logs, not span attributes) — the judge failed every case. Hook output still wins wholesale (#320 preserved); the default conversion now appends the connector trajectory's response steps when the span-built steps carry no response content.

**End-to-end proof** (2-case smoke × 3 agents, `useTraces: true`, headless):

| Agent (span shape) | Correlator used | Result |
|---|---|---|
| Claude Code (session.id only) | Strategy D | 2/2 passed, `ready`, spans on attempt 1 |
| pi (TRACEPARENT adopter) | Strategy A (eval traceId) | 2/2 passed, `ready`, spans on attempt 1 |
| REST/logos (no runId at all) | Strategy A via traceparent header | 2/2 passed, `ready`, spans on attempt 1 |

No foreign spans in any judged trajectory (verified by content inspection). Three full 84-case benchmark runs (one per span shape, running in parallel) kicked off on the same build; `npx tsc --noEmit` clean, `npm run build:all` green, 4642 unit tests pass (updated + new suites for the fail-closed filter, claim ordering, onStopped, error-preservation, and grace-window contracts).

TackAdam (approved):

PR #414, opened 2026-08-25
feat(runs): checkpoint-resume for interrupted evaluation runs + concurrent-safe run updates

goyamegh:
## Summary

Interrupted evaluation runs (server crash / deploy / cancellation) can now be **resumed in place**: `POST /api/storage/evaluation-runs/:id/resume` re-executes ONLY the test cases without a persisted report, preserving completed reports byte-for-byte. The per-test-case reports already persisted in storage act as the checkpoint — nothing new to store, nothing to lose.

> **Stacked on #413** — merge that first. This branch is based on it (the `useTraces` polling gate is required so non-trace agents in the new integration test don't take the 10-minute trace-poll detour). Only the last 4 commits are new here.

### What's included

1. **Resume endpoint + UI** — `POST /api/storage/evaluation-runs/:id/resume` (SSE, mirrors the create route); the run detail page shows a `Resume (N left)` button for interrupted runs with unfinished test cases. Same run id — history and stats stay coherent, `resumedAt` is stamped.
2. **Orphan EvaluationRun recovery on boot** (`evaluationRunRecoveryOnBoot.ts`, sister of `benchmarkRunRecoveryOnBoot.ts`) — stale `running` run docs are marked failed-with-note and become resumable. Previously they were stuck `running` forever with no way forward.
3. **Liveness heartbeat** — the executing server stamps `heartbeatAt` every 60s. Recovery and resume gate on heartbeat staleness (`EVALUATION_RUN_STALE_AFTER_MS`, default 1h), so **multiple agent-health servers sharing one storage cluster can never orphan-kill or double-execute each other's active runs**.
4. **Concurrent-safe run-doc updates (OpenSearch)** — found by verification: any run with `concurrency > 1` could abort within seconds on `version_conflict_engine_exception`. `updateResult` now sets `retry_on_conflict`, and `update()` is a partial doc-merge via the `_update` API instead of a read-modify-write full reindex (which could clobber concurrent per-test-case results).

### Verification (real 84-case benchmark, shared OpenSearch cluster)

- 84-case run at concurrency 4 → `kill -9` the server at **60/84** → run frozen `running`, heartbeat stopped
- resume rejected with 409 while heartbeat fresh; accepted once stale → re-executed **only the 24 unfinished**
- all **60 pre-crash reports verified untouched** (report ids + timestamps identical); final stats 84/84, `total` covers the full run
- second resume attempt → `400 Nothing to resume`
- cancel-then-resume flow exercised end-to-end through the UI button

### Tests

- Unit: `tests/unit/server/routes/storage/evaluationRunResume.test.ts` (resumable-id semantics, liveness/staleness) — 9 cases
- Integration: `tests/integration/server/routes/evaluationRunResume.integration.test.ts` (seed interrupted run → resume via API → preserved vs re-executed reports, full-size stats, 400/404 paths) — runs against a live backend with the built-in demo agent, no AWS needed
- e2e: `tests/e2e/resume-run.spec.ts` (button contract: visible with correct remaining count on interrupted runs, hidden when nothing to resume, POSTs to `/resume`)


goyamegh:
## Adversarial review (codex / gpt-5.4) — findings & resolutions

Ran a second-opinion review from a different model family on this diff before requesting approval. Findings and what was done (all fixes in `c4c77e85`):

| # | Sev | Finding | Resolution |
|---|---|---|---|
| 1 | HIGH | Stale `heartbeatAt` outranked fresh `resumedAt` right after a resume claim → just-resumed run looked orphaned (double-resume window) | ✅ `runLivenessAgeMs` = max(heartbeat, resumed, created); claim also stamps `heartbeatAt`. Regression unit test added |
| 2 | HIGH | No atomic claim — two servers could resume the same orphan concurrently | ✅ Claim token: write `resumeToken` nonce → re-read → abort via SSE error if another claimer won. (No cross-server CAS primitive in the storage interface; write-then-verify shrinks the race to ~one refresh cycle.) Same-process double resume already 409s — integration test added |
| 3 | HIGH | Partial source re-resolution flipped unresolvable test cases to eternally-`pending` on a "completed" run | ✅ Only test cases that will actually execute are reset; missing ones keep their failed-with-note entry and are surfaced as `missingCount` / `missingTestCaseIds` on the `started` event |
| 4 | MED | Boot recovery used offset pagination over the shrinking `status:running` result set → skipped stale runs beyond page 1 | ✅ Two-phase: scan all pages first, mutate after |
| 5 | MED | `update()` doc-merge is a semantic break for the PUT upsert path (omitted nested `results` keys would survive) | ✅ PUT route now does full-document replace via `create()`; internal partial updates keep doc-merge |
| 6 | MED | File backend had the same lost-update race (in-process, at `await` boundaries) | ✅ Per-run write serialization queue in the file adapter; the double-resume integration test runs against the file backend and exercises it |

## Verification proof

- Full unit suite: **258 suites / 4652 passed**
- Integration (against a live backend built from this branch, file storage, built-in demo agent — no credentials involved):
  - `resumes only unfinished test cases, preserves completed reports, then 400s when nothing is left` ✅ (5.1s)
  - `404s for an unknown run id` ✅
  - `409s a second resume while the first is still executing` ✅ (7.6s)
- Manual crash test on a real 84-case benchmark: `kill -9` at 60/84 → heartbeat froze → resume accepted only after staleness threshold → exactly 24 re-executed, all 60 pre-crash reports byte-identical, final stats 84/84.

goyamegh:
## Addendum: resumed runs no longer vanish from `benchmark.runs` (production bug, hit twice)

Two follow-up commits on this branch (`68b62993`, `24facd26`), independent of #417 / #418.

### Bug
`POST /api/storage/evaluation-runs` links a completed run into the benchmark's embedded `runs` array (`storage.benchmarks.addRun`) on its success path — that's what the benchmark detail page and the scoped comparison pool read directly, not the evaluation-run document. If a run's *original* create-route execution crashed before reaching that success branch, the run was never linked — and until this fix, a later `POST .../resume` that finished it successfully didn't link it either, since the resume completion path only updated the evaluation-run document. `GET .../evaluation-runs/:id` and the plain Evaluation Runs list looked completely fine (they read the evaluation-run doc directly); only benchmark-scoped views silently missed the run.

### Fix
The resume completion path now builds the same `BenchmarkRun` projection (`buildBenchmarkRunProjection`) and links it via a new `linkCompletedRunToBenchmark` helper: upserts by run id (`updateRun` if already linked, `addRun` otherwise) so re-resuming a run already linked never duplicates it.

### codex_review findings (applied)
- **Blocking** — a benchmark-linking failure (e.g. benchmark deleted mid-resume) rewrote the run's own `status` to `'failed'`, even though its test cases had genuinely completed and reports were already persisted. → Linking is now in its own try/catch, logged on failure, never allowed to corrupt the canonical run record. New integration test forces this exact path (benchmark id doesn't exist) and asserts the run still reports `completed`.
- **Medium** — `buildBenchmarkRunProjection` used truthy checks for optional fields, silently dropping a real `concurrency: 0` or an intentionally empty `description`/`evaluatorId`. → Switched to explicit `!= null` checks.
- **Documented, not silently claimed solved** — `linkCompletedRunToBenchmark` is read-then-branch-then-write, not an atomic storage upsert; two truly concurrent links of the same run id could both `addRun` and produce a duplicate a later `updateRun` can't repair. Added an explicit code comment describing the known, bounded race window (narrowed by the existing same-process/cross-process resume claim guards) plus a unit test that demonstrates it under controlled concurrency. Closing it fully needs an atomic upsert-by-id primitive in the storage adapters (OpenSearch + file) — a bigger, separate change touching both the create and resume paths uniformly, tracked as a follow-up rather than bundled into this bug fix.
- codex also flagged (medium/low, **not applied — rationale**): (a) the "shared helper" framing overclaims centralization since the create path (owned by unmerged #399) doesn't use it yet — code comments/CHANGELOG were reworded to be explicit that only the resume path uses it today; (b) throwing on a missing benchmark mirrors the create route's own pre-existing convention for the same domain concern, not a new decision — changing that shared convention is a deliberate design call better made once both paths are unified, out of scope here; (c) the denormalized-projection architecture itself (embedded `benchmark.runs` as a hand-maintained secondary view) is a pre-existing pattern in this codebase, not something a single bug-fix PR should redesign.

### Tests (red→green at every level, verified locally)
- Unit: `tests/unit/server/routes/storage/evaluationRunResume.test.ts` — `buildBenchmarkRunProjection` field coverage (including the `!= null` fix), `linkCompletedRunToBenchmark` add-vs-upsert-by-id, a repeated-call regression guard against real array semantics, and the documented concurrent-race test.
- Integration: `tests/integration/server/routes/evaluationRunResume.integration.test.ts` — seeds a benchmark-linked run with an empty `benchmark.runs` (simulating the pre-`addRun` crash), resumes to completion, asserts it appears exactly once, resumes again and asserts still no duplicate; a second test forces the benchmark-not-found path and asserts the run still resolves `completed`.
- E2E: `tests/e2e/resume-run-benchmark-link.spec.ts` — drives a real resume against the mock `demo` agent through the actual UI and asserts the run renders on `/benchmarks/:id/runs`.
- Verified by reverting just the route change and re-running each level (exact expected failure reproduced: e.g. integration test failed with `benchmark.runs` length 0 instead of 1) before restoring the fix.

### Verification
- `npm run test:unit` (this branch's full suite) — 4661 passed, 5 skipped, 0 failed.
- Full `test:integration` suite (73 suites / 760 tests) — run against an isolated, disposable OpenSearch 2.17.0 Docker container (not the shared cluster, not ports 4000/4001) — all green.
- `npm audit --audit-level=high` — pre-existing high-severity findings in `react-router`/`undici`/`vite`, unrelated to this diff (no `package.json` changes), noted per team convention, not a blocker.

Note: this PR (#414) is itself stacked on #413 and not yet mergeable to `origin/main` standalone; this addendum doesn't change that dependency.

PR #415, opened 2026-08-25
fix(benchmarks): version filter can no longer hide all runs; annotated version dropdown + escape-hatch empty state

goyamegh:
## Summary

Two fixes for how the benchmark runs page handles **versions** — motivated by a real incident where a benchmark with 8 runs rendered a completely empty runs panel ("No runs for v8") and looked exactly like data loss.

### 1. `fix`: stale run-version filter no longer hides all runs

The runs page persisted its version filter under **one global localStorage key** (`benchmark-runs:runVersionFilter`). A `v8` filter selected while viewing one benchmark leaked onto **every other benchmark** — including v1 benchmarks with plenty of runs — where it matched nothing and rendered a bogus "No runs for v8" empty state.

- filter is now persisted **per benchmark id**
- impossible values self-heal to "All Versions" via `effectiveRunVersionFilter()` (pure, unit-tested): a persisted version greater than the benchmark's `currentVersion` can never blank the page again

### 2. `feat`: make versions legible — annotated dropdown + escape-hatch empty state

Run-less versions read as "empty copies of the benchmark". Now:

- every dropdown option is annotated: `v8 (latest) · 32 runs`, `v3 · no runs` — empty versions explain themselves *before* selection
- filtering to a version with no runs shows `0 of 35 runs match v3 — Runs exist on other versions` with a one-click **Show all versions (35)** button instead of a dead end

Default stays **All Versions** (run rows carry their `vN` chips), and the user's explicit per-benchmark choice persists.

### Tests

- Unit: `tests/unit/lib/benchmarkVersionUtils.test.ts` — 5 new cases for `effectiveRunVersionFilter` (incl. the exact v8 regression)
- e2e: `tests/e2e/benchmark-version-filter.spec.ts` — seeds BOTH the legacy global key and a stale per-benchmark key before app load and asserts the "No runs for vN" state cannot reproduce (asserts the correct "No runs yet" renders instead)

Verified visually against a real benchmark with 8 versions (32 runs on v8, none on v1–v5): annotations render, escape hatch works, stale-filter injection can no longer blank the page.


goyamegh:
## Adversarial review (codex / gpt-5.4) — findings & resolutions

Second-opinion review from a different model family. Findings and what was done (fixes in `5d02b449`):

| # | Sev | Finding | Resolution |
|---|---|---|---|
| 1 | HIGH | `usePersistedState` never rehydrates when its key changes → per-benchmark keys alone don't fix the leak on client-side navigation (old benchmark's in-memory value survives and gets written under the new key) | ✅ **Confirmed real** — the hook now rehydrates on key change (render-phase derived-state reset). This was the actual root fix; the per-benchmark key was necessary but not sufficient |
| 2 | HIGH | e2e only covered initial mount, not cross-benchmark navigation | ✅ 4 new `renderHook` unit tests pin the exact navigation semantics (rehydrate on key change, default fallback, writes target the current key). Navigation e2e noted as follow-up — the hook tests cover the state machine directly |
| 3 | MED | Helper validated against `currentVersion` upper bound, not the actual version set (sparse/deleted versions slip through) | ✅ `effectiveRunVersionFilter` now takes the benchmark's real version list |
| 4 | MED | Self-heal was a render-time mask; corrupt value stayed in localStorage | ✅ Write-back effect repairs the persisted value once the benchmark is known |
| 5 | MED | Non-integer values (e.g. `1.5` from corrupted storage) passed validation | ✅ `Number.isInteger` + unit test |
| 6 | MED | Single-version benchmark + persisted `v1` + zero runs → filter-specific copy with no visible dropdown | ✅ Zero-run benchmarks always render "No runs yet" regardless of filter |
| 7 | LOW | `benchmarkId ?? 'unknown'` shared bucket during routing transitions | Accepted — harmless with rehydration in place (transitioning out of `'unknown'` now re-reads correctly) |
| 8 | LOW | Title overstates ("can no longer hide all runs" — user can still filter to an empty version) | Accepted wording — the empty state is now an explicit, recoverable filter state, never a silent dead end |

## Verification proof

- Full unit suite: **260 suites / 4722 passed** (incl. 4 new dynamic-key hook tests, 6 helper cases)
- Playwright: `stale persisted v8 filter self-heals instead of hiding all runs` ✅ — seeds both the legacy global key and a stale per-benchmark key before app load
- Manually verified against a benchmark with 8 versions (32 runs on v8, none on v1–v5): annotated dropdown renders, empty-version selection shows the escape hatch, injected stale filters cannot blank the page

TackAdam (approved):

PR #416, opened 2026-08-26
feat(comparison): agreement + category insights band

goyamegh:
## What

A deterministic **agreement + category insights band** on the comparison page, between the scoreboard and Table Compare. No LLM calls — pure arithmetic over the already-loaded comparison rows (`lib/comparisonInsights.ts`).

1. **Agreement chips** — `All/Both pass · All/Both fail · Split` counts across the selected runs (generalized to N runs from day one). Each chip filters Table Compare to its bucket; rows not covered by every run are counted separately and excluded from agreement math.
2. **Category × run pass-rate matrix** — collapsible (open by default), parsed from the bracketed tag imported benchmarks embed in test-case names (`qst_0011 [basic] …`), falling back to `topic:` labels. Categories below a minimum case count roll into `other`. Cells and headers click-to-filter.
3. **Shared-weakness callout** — fires when one category is the weakest (±5pp) for *every* run and genuinely weak (<75% mean): the "benchmark/corpus gap, not an agent choice" signal.

## Why

Comparing N agents on one benchmark, the page answered "who has the higher pass rate" but not *where they disagree* or *what's weak for everyone*. On a live 84-case RAG benchmark (pi vs logos, trace-judged), the band surfaces at a glance: `63 both-pass / 8 both-fail / 13 split`, and that `semantic` is the weakest category for **both** agents (60%/67%, 50% of both-fail cases) — i.e. the next investment is retrieval/chunking, not an agent swap. That story previously required scanning 84 rows by hand.

The math module is deliberately standalone: the upcoming N-run "What's actually different" deep-dive can feed the same deterministic stats to its LLM prompt as grounding context instead of re-deriving counts.

## Existing-filter interaction

An active agreement chip bypasses the `differences` default row filter (an agreement bucket *is* a difference statement — stacking both would hide all-pass/all-fail rows). Category filter composes with everything else.

## Testing

- Unit: `tests/unit/lib/comparisonInsights.test.ts` — partition math (incl. errored/missing verdicts, 3-run generalization), category extraction/rollup, shared-weakness detection (positive + both negative directions).
- E2E: `tests/e2e/comparison-insights-band.spec.ts` — seeded 2-run comparison; asserts chip counts, matrix render, chip→table filtering, collapse/expand.
- Verified live against a real 84-case benchmark comparison (screenshot in comment below).


goyamegh:
## Verification

**Live proof** (isolated recovery-disabled server, real 84-case EnterpriseRAG benchmark, pi vs logos trace-judged runs):
- Agreement chips: `✓ Both pass 63 · ✗ Both fail 8 · ⇄ Split 13 · 84 covered by every run` — matches hand-computed partition from the report docs exactly.
- Category matrix: `semantic ⚠` red for both runs (60% 9/15 / 67% 10/15); `info_not_found` 95 vs 80 and `constrained` 91 vs 100 visible at a glance.
- Shared-weakness callout fires with the honest claim; Split chip filters the table to exactly the 13 disagreement rows.

**Gates**: `tsc --noEmit` clean · `build:all` green · unit 4731 passed (16 for the new math module) · e2e `comparison-insights-band.spec.ts` green (seeded 2-run comparison: chip counts, matrix, chip→table filtering, collapse/expand) · integration 758/758 (26 CLI-suite failures reproduced only when a foreign server owns port 4001 — the foreign-server guard, not this change).

## Adversarial review (Codex/gpt-5.4) — 8 findings, all addressed in the follow-up commit

| Finding | Resolution |
|---|---|
| HIGH: verdict-less results (`status:'failed'`+null verdict, evaluator-errored) bucketed as fails | `hasVerdict` reworked: errored (#242) and verdict-less-completed → **uncovered**; run-level `status:'failed'` remains a legitimate fail verdict |
| HIGH: `other` rollup not filterable with the semantics it was computed with | breakdown exposes `members` (raw categories per column); filter matches on that set |
| HIGH: stale filters when run selection changes | both insights filters reset on any selection change |
| HIGH: silent `rowStatusFilter` bypass | removed — activating a chip **visibly** flips the row-status pill to `Show all`; filters compose normally |
| MED: band counts vs otherwise-filtered table | accepted + labeled: band is the run-level summary of **all shared cases** ("counts are for all shared cases" in the copy); chips *set* filters rather than mirror them — recomputing counts under active filters would make the numbers jump while used as controls |
| MED: clickable cells promise run×category precision | cells are display-only; column headers carry the filter |
| MED: synthetic `other`/`uncategorized` collide with real category names | sentinels renamed `(other)`/`(uncategorized)` — parentheses are outside the name-tag charset `[\w-]`, collision-impossible by construction (unit-tested) |
| MED: "weakest" claim false under 5pp tolerance | tolerance tightened to 1pp (tie-level); callout never fires when another category is meaningfully lower for any run |

TackAdam (approved):

PR #417, opened 2026-08-26
fix(evaluation-runner): trace-judged runs no longer inflate pass stats

goyamegh:
## Root cause

For a trace-judged (`useTraces`) test case, `waitForTracesAndJudge()` wrote the real judge verdict to storage but returned `void`. The caller in `services/evaluationRunner.ts` never saw the resolved judgment — it kept using its stale in-memory `savedReport` (still `metricsStatus: 'pending'`, no `passFailStatus`) — so `run.results[testCaseId]` was recorded as a bare `'completed'` result with **no verdict**. The run-completion stats loop then counted every bare `'completed'` result as **passed**, regardless of the real verdict. Net effect: a run with e.g. 66/84 real "passed" judgments displayed as **84/84 passed**.

`EvalRunDetailPage` (run detail view) trusted the denormalized `run.stats` directly with no defensive recompute — unlike the runs list, which already recomputes from `run.results` — so the detail page is where the bug was user-visible.

## Fix

- `waitForTracesAndJudge()` now **resolves with the judgment it just computed** (or `{ passFailStatus: null, metricsStatus: 'error' }` on a judge/trace-polling failure) instead of `void`; the caller uses that outcome directly instead of re-reading a stale object.
- `run.stats` is now computed via `lib/runStats.ts`'s `bucketRunResults()` — the same canonical verdict-based bucketing the runs list and comparison page already use — instead of a local loop that treated bare `'completed'` as `'passed'`. A `'completed'` result with no verdict is now bucketed `errored`, never a silent pass.
- `EvalRunDetailPage` now shares a new `getDisplayStats()` helper (`lib/runStats.ts`) with the runs list: prefer bucketing `run.results`, fall back to `run.stats` only when `results` is empty.

## Tests (red → green)

- **Unit**: `tests/unit/lib/runStats.test.ts` (new `getDisplayStats` cases, including "counts a completed result with no verdict as errored, not passed"), `tests/unit/services/evaluationRunner.test.ts` (stats assertions updated for the errored bucket).
- **Integration**: `tests/integration/services/evaluationRunner.integration.test.ts` adds a multi-test-case mixed-verdict regression (2 passed / 2 failed via trace judging) asserting `run.stats` matches the real split, and strengthens the existing single-case trace-polling test to assert `passFailStatus`/stats. Both fail red on the pre-fix code (asserted locally by reverting the fix commit).
- **e2e**: `tests/e2e/eval-run-detail-stats.spec.ts` (imports `./fixtures/test-fixtures`) seeds an evaluation-run doc with a deliberately-buggy `run.stats` (4 passed / 0 failed) but real mixed `run.results` verdicts (2 passed / 2 failed), and asserts the detail page renders the real numbers.

All new/changed tests were verified to fail against the pre-fix code and pass against the fix.

## Existing persisted data

The underlying `EvaluationReport` documents for already-completed trace-judged runs were judged correctly all along (the storage write in `waitForTracesAndJudge` always succeeded) — only the run-level `results`/`stats` denormalization was ever wrong. Runs created going forward self-heal via this fix. Already-persisted runs' displayed stats can be repaired losslessly (no data loss) by recomputing `run.results[testCaseId].passFailStatus` + `run.stats` from each report via `lib/runStats.ts`'s `calculateRunStats()` and writing the result back. Not shipped here as an automatic migration — it's a targeted write against a live storage cluster and should be run deliberately, as a follow-up.

## Pre-existing failures (not introduced by this change)

Verified failing identically on a clean `origin/main` checkout (and still failing with this fix reverted), so CI failures here are unrelated:
- `tests/e2e/skills.spec.ts:76` and `:99`
- `npm audit --audit-level=high`: pre-existing high-severity CVEs in `undici` and `vite` transitive deps.

## Scope

No API/schema changes. `services/evaluationRunner.ts`, `lib/runStats.ts`, `components/evals3/EvalRunDetailPage.tsx` plus tests/CHANGELOG.


goyamegh:
## Verification proof

**Red → green, reproduced locally** by swapping in the pre-fix (`eff3e36c`) versions of `services/evaluationRunner.ts`, `lib/runStats.ts`, and `components/evals3/EvalRunDetailPage.tsx` under the *current* (post-fix) test suite, then restoring the fix — confirmed via `git status --short` (clean) that the restore exactly matched this PR's HEAD.

### RED (pre-fix production code + this PR's tests)

```
FAIL tests/unit/lib/runStats.test.ts
  ● runStats › getDisplayStats › treats an explicit empty results object the same as absent results (falls back to stats)
    TypeError: (0 , runStats_1.getDisplayStats) is not a function
  ● runStats › getDisplayStats › falls through to all-zero when results is empty and stats.total is 0 (run not started)
    TypeError: (0 , runStats_1.getDisplayStats) is not a function
  ... (getDisplayStats doesn't exist pre-fix)

FAIL tests/unit/services/evaluationRunner.test.ts
FAIL tests/integration/services/evaluationRunner.integration.test.ts
  ● executeEvaluationRun › trace polling integration › REGRESSION (trace-judged stats inflation): run.stats reflects real mixed verdicts, not "every completed = passed"
    expect(result.stats).toEqual({ passed: 2, failed: 2, pending: 0, errored: 0, total: 4 })
    - all 4 test cases counted as passed:  { passed: 4, failed: 0, pending: 0, total: 4 }

Test Suites: 3 failed, 1 passed, 4 total
Tests:       14 failed, 56 passed, 70 total
```

### GREEN (this PR's code + this PR's tests)

```
PASS tests/unit/lib/runStats.test.ts
  runStats
    getDisplayStats
      ✓ prefers results-derived verdicts over a stale/inflated run.stats blob
      ✓ falls back to run.stats when results is empty (run has not started / legacy data)
      ✓ returns all-zero stats when neither results nor stats are present
      ✓ treats an explicit empty results object the same as absent results (falls back to stats)
      ✓ falls through to all-zero when results is empty and stats.total is 0 (run not started)
      ✓ falls through to all-zero when results is empty and stats is entirely absent
      ✓ defaults missing individual stats fields to 0 (partial/legacy stats blob)

PASS tests/unit/services/evaluationRunner.test.ts
PASS tests/integration/services/evaluationRunner.integration.test.ts
  executeEvaluationRun
    trace polling integration
      ✓ triggers trace polling when report has metricsStatus pending
      ✓ REGRESSION (trace-judged stats inflation): run.stats reflects real mixed verdicts, not "every completed = passed"
      ✓ judge failure during trace-judged polling is bucketed errored, not passed (does not crash the run)
      ✓ trace polling failure (onError) is bucketed errored, not passed

Test Suites: 4 passed, 4 total
Tests:       70 passed, 70 total
```

### CI (this PR, commit `36228527`)

All 17 checks green, including `codecov/patch` (85.71% of diff hit, target 59.39%) and `codecov/project` (70.78%, +11.38% vs. `eff3e36c`), `DCO`, `lint-and-typecheck`, `build-and-test` on Node 18/20/22, `integration-tests`, `e2e-tests`, `security-scan`, `license-check`, `check-changelog`, `link-check`.

### e2e spec (not run locally in this comment — Playwright requires a live server/OpenSearch; verified green in CI's `e2e-tests` check)

`tests/e2e/eval-run-detail-stats.spec.ts` seeds an evaluation-run doc with a buggy denormalized `run.stats` (`{passed: 4, failed: 0}`) but real mixed `run.results` verdicts (2 passed / 2 failed), and asserts the rendered detail page shows `2`/`2` — the numbers a regression back to trusting `run.stats` directly would break.

No screenshots or Amazon-internal data in this proof — this is generic open-source harness code exercised with synthetic fixtures.

TackAdam (approved):

goyamegh:
Pushed a follow-up commit (8e202f2) after a cross-PR review pass ahead of the release push:

- **Rebase check**: verified against current origin/main (eff3e36c) — this branch is already based directly on that commit, so `git merge origin/main` is a no-op today. #413 (trace-correlation strategies, open, not yet merged) will touch the same `if (!hasDeterministicEval && savedReport.metricsStatus === 'pending' && savedReport.runId)` gate in services/evaluationRunner.ts (dropping the runId requirement, gating on `agentConfig?.useTraces` instead) — this PR's `judgeOutcome` capture wraps that block but doesn't modify those 3 condition lines itself, so no logic conflict is expected when #413 lands; whoever integrates second should just confirm `judgeOutcome = await waitForTracesAndJudge(...)` survives on top of #413's new condition.
- **Consolidated `getDisplayStats()` into `computeRunStats()`** in lib/runStats.ts: the sibling `ui-fixwave-1` branch ("fix(ui): evals3 papercuts") independently extracts a `computeRunStats()` of the same shape from EvalRunsPage/BenchmarkRunsPage's local duplicated helpers into this same file. Renamed this PR's helper to match (superset shape: passed/failed/errored/pending/total) so only one canonical stats function lands in lib/runStats.ts regardless of merge order — flagging here so that PR's owning session isn't surprised by an export-name collision on rebase.
- Added a producer/consumer agreement regression test (tests/unit/lib/runStats.test.ts) pinning that the run-completion writer's `bucketRunResults` output and `computeRunStats`'s own recompute never diverge for the same seeded mixed-verdict run — the double-correction risk flagged during review.

Re-ran full build:all + test:unit (4724 passed) + test:integration (761 passed) locally on the updated branch; all green. DCO signoff intact on all 4 commits.

PR #418, opened 2026-08-26
fix(storage): stop EvaluationRun.results from blowing up evals_experiments field-count mapping

goyamegh:
## Root cause

Real incident, 2026-08-26: a 400-test-case `EvaluationRun` crashed mid-run at 243/400 test cases with `Limit of total fields [5000] has been exceeded`. Ops mitigation raised `index.mapping.total_fields.limit` on the shared cluster, but that only defers the ceiling for any fresh/other deployment.

`EvaluationRun` (`docType: 'evaluation-run'`) is a **top-level** document in the `evals_experiments` index (same index as `Benchmark`, discriminated by `docType` — see `OpenSearchEvaluationRunOperations`). Its `results: Record<testCaseId, {...}>` map is a *different* field than the legacy nested `runs.results` embedded inside `Benchmark.runs`, which already got `enabled: false` protection back in 2026-03 (#… commit `326e31e7`). The new top-level `results` field was never added to `indexMappings.ts`, so OpenSearch dynamically mapped `results.<testCaseId>.reportId` / `.status` / their `.keyword` sub-fields as new fields — shared across **every** `EvaluationRun` document ever created in the index — until the field-count ceiling was hit.

Confirmed against a read-only mapping dump of the live shared cluster: the top-level `results` field already has 800+ dynamically-mapped sub-properties (one set per historical `testCaseId`), while the nested `runs.results` correctly stays `{type: 'object', enabled: false}`.

## Fix

`server/constants/indexMappings.ts`: add `results: { type: 'object', enabled: false }` at the top level of the `benchmarks` (`evals_experiments`) index mapping, mirroring the existing nested protection.

- `enabled: false` stores the subtree in `_source` (reads/writes unaffected — including `OpenSearchEvaluationRunOperations.updateResult()`'s painless partial update, `ctx._source.results.put(...)`) but never parses it into the mapping.
- Confirmed no query anywhere filters/sorts/aggregates on `results.*` subfields (`OpenSearchEvaluationRunOperations.list()` only queries `docType.keyword` / `benchmarkId.keyword` / `agentKey.keyword` / `status.keyword` / `trigger.keyword` / nested `testCaseSnapshots.id.keyword` — none of those are touched by this change).
- Deliberately minimal diff: did **not** add explicit mappings for `docType`/`benchmarkId`/`agentKey`/`status`/`trigger` even though they're also dynamically inferred — those are already inferred as `text` + `.keyword` multi-field, exactly matching what `list()`'s `<field>.keyword` term queries expect; declaring them explicitly risks a type mismatch that would break those queries, for no benefit (they're bounded scalars, not a growth vector).

### Backward compatibility (verified against a disposable OpenSearch 2.17.0 container, not the shared cluster)

- **Fresh deployments / not-yet-poisoned indexes**: `ensureIndexes()`'s `putMapping` call applies cleanly. Verified: created a fresh index, wrote a doc with 500 synthetic `results` entries — mapped field count stayed flat (3 fields) instead of growing to 1500+.
- **Already-poisoned indexes (e.g. the shared cluster)**: OpenSearch rejects an `enabled` change on a field that already has real dynamically-mapped sub-properties (`mapper_exception: the [enabled] parameter can't be updated for the object mapping [results]`). `updateExistingIndex()`'s `putMapping` call is already wrapped in try/catch (best-effort — see existing code), so this fails safely: caught, surfaced as a warning, **no crash, no data loss, no reindex triggered**. It does mean the shared cluster's `results` field keeps growing until an explicit reindex (out of scope here — ops already mitigated via the raised `total_fields.limit`). This fix protects every fresh/other deployment immediately and any existing index that hasn't hit this bug yet.
- Left the shared cluster's `index.mapping.total_fields.limit` untouched, per standing directive — this PR doesn't touch cluster settings, only the mapping template applied at index-creation/update time.

## Tests (red → green)

- **Unit** — [`tests/unit/server/constants/indexMappings.test.ts`](tests/unit/server/constants/indexMappings.test.ts): asserts the top-level `results` field is `{ type: 'object', enabled: false }`, and that `docType`/`testCaseSnapshots` are deliberately left dynamic (still queried via `term`/nested queries elsewhere).
- **Integration** (real OpenSearch, not mocked) — [`tests/integration/services/storage/evaluationRunMappingGrowth.integration.test.ts`](tests/integration/services/storage/evaluationRunMappingGrowth.integration.test.ts): boots the real `ensureIndexesWithValidation()` + `OpenSearchStorageModule` code path against a live cluster (`TEST_OPENSEARCH_ENDPOINT`, defaults to the same `localhost:9200` CI's `integration-tests` job provisions), synthesizes a run with **500** test-case results (more than the 400 that crashed in prod), and:
  - confirms all 500 results persist and read back correctly,
  - confirms the index's total mapped-field count doesn't grow (within a small slack tolerant of other integration test files sharing the same CI index),
  - confirms `results`'s mapping stays exactly `{ type: 'object', enabled: false }`.
  - Also exercises the real per-test-case `updateResult()` painless-update path incrementally.
  - **Verified red→green locally**: reverting just the `indexMappings.ts` change reproduces the exact incident mechanism against a disposable cluster — 500 results added exactly 2500 new mapped fields (47 → 2547) — confirming this test would have caught the regression.
  - Deliberately non-destructive to the shared CI index (never deletes/truncates `evals_experiments`) since other integration test files run against the same OpenSearch service container, possibly in parallel Jest workers.

## Verification

- `npm run test:unit` — 4717 passed, 5 skipped, 0 failed.
- `npm run test:integration` — 760 passed (74 suites), 0 failed, run against a real disposable OpenSearch 2.17.0 container + the real backend server on a non-default port (never touched ports 4000/4001 or the shared cluster).
- `npm run test:e2e` — 8 pre-existing/flaky failures reproduced identically on unmodified `origin/main` (verified by re-running the same failing specs after `git stash`ing this change) — unrelated areas (Skills evaluator page, benchmark sort, test-case-runs back button), not touched by this diff.
- `npm audit --audit-level=high` — pre-existing high-severity findings in `react-router`/`undici`/`vite`, unrelated to this change (no `package.json`/`package-lock.json` changes in this diff); noted per team convention, not a blocker.

Independent of #417 — no merge-order dependency.


goyamegh:
## Verification proof

**CI (final commit `ffa3e63`):** all checks green — `build-and-test` (Node 18/20/22), `lint-and-typecheck`, `integration-tests`, `e2e-tests`, `release-rehearsal`, `security-scan`, `Mend Security Check`, `license-check`, `link-check`, `check-changelog`, `DCO`, `Code-Diff-Analyzer`/`Code-Diff-Reviewer`, `codecov/patch`, `codecov/project`. (`publish-badges` shows `skipped`, expected off the default branch.)

**Local verification (in addition to CI), against a disposable OpenSearch 2.17.0 container — never the shared/production cluster:**

```
$ npm run build:all                 # green
$ npm run test:unit                 # Test Suites: 259 passed | Tests: 5 skipped, 4718 passed, 0 failed
$ npm run test:integration          # (real backend + real OpenSearch) Test Suites: 74 passed | Tests: 760 passed, 0 failed
$ npm run test:e2e                  # 8 failures — reproduced identically on unmodified origin/main
                                     # (Skills evaluator page, benchmark sort, test-case-runs back button —
                                     # unrelated areas, not touched by this diff)
$ npm audit --audit-level=high      # pre-existing react-router/undici/vite findings, no package.json/
                                     # package-lock.json changes in this diff
```

**Red → green regression proof (both fixes in this PR), reverting only the relevant file and re-running against a fresh disposable cluster:**

1. `results` field-count growth (the primary fix):
   ```
   # with server/constants/indexMappings.ts reverted:
   ✕ synthesizes a run with 500 test-case results ... does not grow the mapping
     Expected: <= 15   Received: 2500   (baseline 47 -> 2547 mapped fields)

   # with the fix restored:
   ✓ synthesizes a run with 500 test-case results ... does not grow the mapping
   ✓ incrementally written results (real updateResult() painless path) ... does not grow the mapping
   ```

2. `evaluationRuns.list({ testCaseId })` nested-vs-object mapping mismatch (found via `codex_review` while verifying #1):
   ```
   # with server/adapters/opensearch/StorageModule.ts reverted:
   ✕ does not throw a query_shard_exception
     ResponseError: search_phase_execution_exception: [query_shard_exception]
     Reason: failed to create query: [nested] nested object under path
     [testCaseSnapshots] is not of nested type

   # with the fix restored:
   ✓ does not throw a query_shard_exception (nested-vs-object mapping mismatch)
   ✓ returns only runs whose testCaseSnapshots contain the requested testCaseId
   ```

**Backward-compat check (empirical, against a disposable OpenSearch 2.17.0 container seeded to reproduce the shared cluster's already-poisoned state — the shared cluster itself was never touched):**

```
$ curl -X PUT .../poisoned_index/_mapping -d '{"properties":{"results":{"type":"object","enabled":false}}}'
{"error":{"root_cause":[{"type":"mapper_exception",
  "reason":"the [enabled] parameter can't be updated for the object mapping [results]"}],
  "status":500}}
```
Confirms the documented behavior: `ensureIndexes()`'s existing try/catch around `putMapping` swallows this as a warning (no crash, no data loss, no forced reindex) on an already-poisoned index (e.g. the shared cluster), while a fresh or not-yet-poisoned index gets the fix applied cleanly.

Independent of #417 (different bug, different files) — no merge-order dependency.

TackAdam (approved):

PR #420, opened 2026-08-26
Readable test-case definitions in run views

### LINKED ISSUE #419: Run views show the test-case definition as a raw JSON blob
## Symptom

Reviewing how an agent performed requires being able to read the test it ran. Today, run views dump the entire test-case definition as a raw JSON blob, which makes the prompt, expected outcomes, context, and metadata unnecessarily difficult to scan.

The current rendering is in `components/evals3/CollapsibleTestCaseDefinition.tsx:58`, where the component displays `JSON.stringify(testCase, null, 2)`.

> "Shouldn't it be ready to read what the test is about when I try to see how the agent performed in that test for that run?"

## Expected behavior

Render the definition in the same reader-oriented form as the test-case detail page:

- prompt in a readable text block
- expected outcomes as a checklist
- context as individual items
- category and difficulty as chips

Keep the serialized object available for debugging, but demote it behind a **View raw JSON** toggle instead of making raw JSON the primary presentation.

ashwin-pc:
Fixes #419.

When reviewing how an agent performed in a run, the test-case definition is the rubric the reviewer scores against — but run views rendered it as `JSON.stringify(testCase, null, 2)` (CollapsibleTestCaseDefinition.tsx). Motivating user question: *"Shouldn't it be ready to read what the test is about when I try to see how the agent performed in that test for that run?"*

## Changes
- Extracted the reader-oriented definition rendering (description/provenance, INPUT prompt, EXPECTED outcomes checklist, CONTEXT items, category/difficulty chips) into a shared component, reused by both the test-case detail page and every run view that shows a definition.
- Raw JSON demoted to an explicit **JSON** toggle for debugging; SDK-type cases keep their existing handling.
- Responsive at phone and desktop widths.

## Before / after
| Before | After |
|---|---|
|  |  |

Mobile: 

Component tests assert the readable fields render and raw JSON appears only behind the toggle.


goyamegh:

Reviewed the full diff (6 files) against the PR's stated goal (replace raw-JSON test-case rendering in run views with a reader-oriented shared component, demoting JSON behind a toggle). Verified the two most material findings against the current tree before including them.

|---|---|---|
| Blocker (worth resolving before merge) | `TestCaseDefinition.tsx`'s SDK short-circuit (`if (testCase.sourceFile) return <source-file-pointer-view>`) assumes code-authored test cases have no declarative fields — but `services/sourceResolver.ts`'s `resolveCodeImport()` sets `sourceFile` *alongside* `initialPrompt`, `expectedOutcomes`, `context`, and `description` on the same document (forwarded specifically so server-side evaluators can grade code-authored tests, per the comment near sourceResolver.ts's `expectedOutcomes` forwarding). So for any SDK test case that has those fields populated, this component will hide real rubric/prompt data behind just a "source file pointer" note instead of showing it. | Worth checking: should the branch be `sourceFile && !initialPrompt` (or similar), or render both the source-file pointer *and* the declarative fields when present? |
| Major | `labelValue()` reimplements category/difficulty label parsing locally (lowercase-prefix match, no value validation) instead of reusing the existing canonical `parseLabels()` in `lib/labels.ts` (which validates difficulty against `DIFFICULTY_VALUES` exactly). Since it's preferred over `testCase.category`/`testCase.difficulty`, this could show a different chip than other UI surfaces that read those normalized fields directly for the same test case. | Consider reusing `parseLabels()` for consistency, unless there's a reason the local, more permissive parsing is intentional here. |
| Major | Test coverage: the only new test is a unit test (jsdom + mocked `react-markdown`) covering the readable-fields render, raw-JSON lazy-toggle, and the SDK-fallback branch. Given the project's own testing convention calls for Playwright coverage on UI-visible bug fixes (and this touches 3 different run/test-case views), there's no e2e test asserting the fix actually renders correctly in the run inspector / standalone run page / test-case detail page in a real browser. | Consider adding at least one Playwright spec exercising the readable definition + raw-JSON toggle on one of the affected pages, per the repo's stated testing bar for UI-visible fixes. |
| Minor | `formatContextValue()` runs `JSON.parse` on every context value and reformats on success — a context value that's just a plain string like `"123"` or `"true"` (which are also valid JSON) would get its original literal quoting/formatting altered in the "readable" view. | Probably fine in practice, but worth a quick check that this doesn't unexpectedly reformat legitimate plain-string context values. |
| Nit | The raw-JSON view is lazily *mounted* (not rendered into the DOM until toggled), but `JSON.stringify(testCase, null, 2)` itself still runs eagerly on every render regardless of whether the toggle is open — so the "deliberately not mounted until asked for" comment is only half true for the serialization cost (likely negligible in practice, just noting the discrepancy). | No action needed unless test-case objects get very large. |

Nothing here undermines the direction of the change (it's a nice readability win, and the shared-component extraction is reasonable for 3 call sites) — the SDK-case data-hiding issue is the one I'd want addressed before merge since it could silently regress rubric visibility for a subset of code-authored tests.

ashwin-pc:
Both findings addressed — the first was a genuine over-correction on my side:

- **Mixed SDK/declarative rendering** (`e45614a`): the `sourceFile` short-circuit is gone. Cases that carry both a source pointer *and* populated declarative fields now render the pointer **and** the prompt/expected outcomes/context; pointer-only remains the fallback when no declarative fields exist. You were right that the guard was hiding grading inputs — it was added to prevent an empty rubric for code-authored cases and ended up creating a hidden one.
- **Label normalization** (`3eb0673`): the permissive local `labelValue()` helper is replaced by the canonical `parseLabels()`, so category/difficulty chips can't diverge from canonical parsing.

One coordination note for whoever merges: this branch now overlaps #439 (context item dispositions). It carries the definition-page disposition bridge, the shared `ContextDispositionGroups` renderer, and the `ContextItemDisposition` typing. If #439 merges first, this branch rebases and keeps the shared renderer; if this merges first, #439 should drop or reconcile the overlapping portions. Happy to sequence whichever way you prefer.

goyamegh (approved):

PR #421, opened 2026-08-26
fix(storage): SigV4 credentials now refresh mid-process, no restart after ada rotation

goyamegh:
## What

A long-running agent-health server's SigV4-signed OpenSearch storage/observability client silently wedges once the resolved AWS credentials rotate (e.g. hourly `ada credentials update`): every OpenSearch op starts failing with opensearch-js's opaque, bodiless `Response Error` (a 403 under the hood), while the cluster is green and a fresh CLI call against the same credentials works. **This fixes the operational "server must be restarted after `ada` rotation" failure mode.**

## Root cause

`server/services/opensearchClientFactory.ts` already wired `AwsSigv4Signer`'s `getCredentials` to call `fromNodeProviderChain()` fresh on every invocation — so the hook itself was never "one-shot". The actual bug is in when opensearch-js's own transport decides to call that hook again:

- opensearch-js's `AwsSigv4Signer` transport caches the last-resolved credentials object in memory and only re-invokes `getCredentials()` once it decides those cached credentials are expired, via `needsRefresh()` / `.expired` / `.expireTime` / `.expiration`.
- Session/profile credentials resolved from `~/.aws/credentials` — exactly how `ada credentials update` writes them — carry **none** of those fields (verified directly: `fromNodeProviderChain({profile:'default'})()` on a real `ada`-rotated profile returns `{ accessKeyId, secretAccessKey, sessionToken }` with no `expiration`).
- Without an `expiration`, opensearch-js's expiry check never fires again after the very first successful request, so the signer freezes on the first-ever-resolved access key for the life of the process — even though our own `getCredentials()` would happily return fresh credentials if only opensearch-js called it again.

## Fix

`resolveSigv4Credentials()` now attaches a synthetic 15-minute `expiration` to the resolved credentials whenever the provider chain doesn't supply one itself (real STS temporary-credential `expiration` values, when present, are left untouched). This makes opensearch-js's own expiry check periodically force a re-resolve, which re-reads whatever `ada` (or any other rotation mechanism) last wrote to disk/environment — no restart required.

This is the single shared client-construction helper (`createOpenSearchClient`) used by **both** the storage and observability/traces OpenSearch clients (`server/services/storageInitializer.ts`, `server/middleware/storageClient.ts`, `server/services/observabilityClient.ts`, `server/adapters/index.ts`, `server/routes/storage/admin.ts` all route through it), so one fix in one file covers both construction sites.

Also, minimal error-surface improvement: `describeOpenSearchError()` / `opensearchErrorStatusCode()` append the HTTP status code to storage error messages/logs (e.g. `Response Error (HTTP 403)`) at the storage-init, health-check, and test-connection call sites, so an auth failure (expired/rotated SigV4 credentials) is distinguishable from a cluster-side 5xx without a broader error-handling rewrite.

## Tests (red→green)

- **Unit** — `tests/unit/server/services/opensearchClientFactory.test.ts`:
  - `getCredentials` is wired as a function (a refreshing provider hook), never a resolved credentials object.
  - `resolveSigv4Credentials()` creates a fresh provider chain and calls it on every invocation (no memoized/one-shot credentials).
  - Attaches a synthetic near-term expiration when the resolved credentials carry none; **honors** a real expiration when the provider does supply one (doesn't clobber genuine STS temporary credentials).
  - **Regression proof**: simulates an `ada`-style rotation — a mocked provider chain returns a stale key on the first call and a rotated key on the second — and asserts the second `getCredentials()` call surfaces the fresh key.
  - `opensearchErrorStatusCode` / `describeOpenSearchError` status-code formatting (403 vs 503, message-less errors, etc.)
  - `tests/unit/server/services/storageInitializer.test.ts`, `tests/unit/server/adapters/index.test.ts`, `tests/unit/server/adapters/opensearch/StorageModule.test.ts` — status-code-aware error messages at each surfacing site (storage init, health check, test-connection).
  - Confirmed **red** on pre-fix code (`git stash` of the 4 source files → 15 test failures, incl. `TypeError: ... is not a function` for the new exports) and **green** after the fix.
- **Integration** — not added. The existing `storageClient` integration harness authenticates via basic auth against a live backend *process* started from env vars; it has no seam to inject a mock AWS credential provider into that already-running process, and provisioning a real SigV4-reachable OpenSearch domain in CI is out of scope here. The unit tests above already exercise the exact function opensearch-js calls at request time (via a mocked `fromNodeProviderChain`), with full red→green proof of the rotation scenario.
- Not UI-visible (no new UI surface) — the two response-shape changes (status-code-suffixed error strings) are covered by the boundary tests above (`testStorageConnection`/`testObservabilityConnection`/`health()`), which is what the Settings UI actually consumes.

`npm run build:all`: green. `npm run test:unit`: 259 suites / 4730 tests green (198/198 in the four touched suites, +15 new/changed assertions).

## Independence from other open PRs

Verified no overlap with #417 (trace-judge-stats), #418 (eval-results-mapping), or #414 (evaluation-runs checkpoint-resume) — none touch `opensearchClientFactory.ts`, credential/signer code, or storage error formatting. This PR only touches: `server/services/opensearchClientFactory.ts`, `server/services/storageInitializer.ts`, `server/adapters/index.ts`, `server/adapters/opensearch/StorageModule.ts`, their tests, and `CHANGELOG.md`.


goyamegh:
## Verification proof

**Red→green (pre-fix vs post-fix), touched source files only:**
```
$ git stash push -- server/services/opensearchClientFactory.ts server/services/storageInitializer.ts \
    server/adapters/index.ts server/adapters/opensearch/StorageModule.ts
$ npx jest tests/unit/server/services/opensearchClientFactory.test.ts \
    tests/unit/server/services/opensearchClientFactory.sigv4Refresh.contract.test.ts \
    tests/unit/server/services/storageInitializer.test.ts \
    tests/unit/server/adapters/index.test.ts \
    tests/unit/server/adapters/opensearch/StorageModule.test.ts
...
Test Suites: 5 failed, 5 total
Tests:       15 failed, 187 passed, 202 total
● ... TypeError: (0 , opensearchClientFactory_1.opensearchErrorStatusCode) is not a function
● ... TypeError: (0 , opensearchClientFactory_1.describeOpenSearchError) is not a function
...

$ git stash pop
$ npx jest tests/unit/server/services/opensearchClientFactory.test.ts \
    tests/unit/server/services/opensearchClientFactory.sigv4Refresh.contract.test.ts \
    tests/unit/server/services/storageInitializer.test.ts \
    tests/unit/server/adapters/index.test.ts \
    tests/unit/server/adapters/opensearch/StorageModule.test.ts
...
Test Suites: 5 passed, 5 total
Tests:       202 passed, 202 total
```

**Contract test against the REAL, unmocked opensearch-js `Client` + `AwsSigv4Signer` transport** (`tests/unit/server/services/opensearchClientFactory.sigv4Refresh.contract.test.ts`) — only `fromNodeProviderChain` is mocked, everything else is the real dependency:
```
SigV4 credential refresh — real opensearch-js transport contract
  ✓ documents the bug: with NO expiration on the resolved credentials, the real
    opensearch-js transport never calls getCredentials() again, even across a 1h+ gap
  ✓ proves the fix: resolveSigv4Credentials()'s synthetic expiration makes the SAME
    real transport re-invoke getCredentials() once the refresh window elapses
  ✓ createOpenSearchClient() end-to-end: the wired getCredentials hook is honored
    by the real signer/transport
```

**Root-cause verification (manual, against a real `ada`-rotated profile on this machine):**
```
$ node -e "
const { fromNodeProviderChain } = require('@aws-sdk/credential-providers');
fromNodeProviderChain({ profile: 'default' })().then(c => console.log(Object.keys(c), c.expiration));
"
[ 'accessKeyId', 'secretAccessKey', 'sessionToken', '\$source' ] undefined
```
Confirms the root cause directly: profile-based credentials as `ada` writes them to `~/.aws/credentials` carry no `expiration`, which is exactly the field opensearch-js's `AwsSigv4Signer` transport relies on to decide whether to call `getCredentials()` again.

**Full build + unit suite:**
```
$ npm run build:all                 # green (client, server, cli, lib)
$ npm run test:unit
Test Suites: 260 passed, 260 total
Tests:       5 skipped, 4734 passed, 4739 total
```

**CI on this PR (commit 5b00d7b4):** build-and-test (Node 18/20/22), lint-and-typecheck, security-scan, license-check, release-rehearsal, integration-tests, check-changelog, DCO, codecov/patch, codecov/project — all ✅ green. e2e-tests running (Playwright; unrelated to this backend-only change, no new UI surface).

**Operational note:** this fixes the "server must be restarted after `ada` rotation" failure mode reported 2026-08-26 — a long-running server's storage/traces OpenSearch client would wedge on every op with an opaque 403 ("Response Error") once the SigV4 credentials it started with rotated, requiring a manual restart to recover. It now self-heals within ~5 minutes of a rotation instead.

**codex_review (adversarial second opinion, gpt-5.4):** ran on the diff pre- and post-fix-up. Findings and disposition:
- HIGH — tests only exercised our own helper, not the real opensearch-js `AwsSigv4Signer` transport contract → **applied**: added `opensearchClientFactory.sigv4Refresh.contract.test.ts` against the real, unmocked `Client`/`AwsSigv4Signer`.
- HIGH — CHANGELOG/PR wording implied restart-free zero-downtime rotation, when the fix is actually a bounded self-heal window → **applied**: reworded to "self-heals within ~5 minutes", reduced the window from 15→5 min.
- MED — mutating the credentials object returned by the provider risks poisoning a memoized/shared instance → **applied**: `resolveSigv4Credentials()` now returns a spread copy instead of mutating in place.
- MED — `StorageModule.health()`'s `describeOpenSearchError()` call had no `|| fallback`, so a message-less error rendered as `error: ''` on the exact operator-facing health endpoint → **applied**: falls back to `'Unknown error'` (matching the other 3 call sites), with a regression test.
- MED — only `.expiration` is checked, not opensearch-js's other (AWS SDK v2-only) freshness signals → **rejected**: `fromNodeProviderChain` is AWS SDK v3 and only ever produces `.expiration`; the v2-only fields (`needsRefresh()`/`.expired`/`.expireTime`) can't appear on its output, documented in the updated code comment instead of adding dead branches.
- MED — hardcoded refresh window forces periodic re-resolution for all non-expiring SigV4 sources, not just the `ada` case → **rejected, with mitigation**: this is an intentional generalization (there's no reliable way to detect "this credential source needs a synthetic expiration" other than "it didn't supply one"), and the window was reduced 15→5 min to shrink the tradeoff; making it configurable would be premature flexibility for a single call site.
- LOW — only 4 of ~33 `StorageModule.ts` catch blocks got status-code-aware messages → **rejected, as designed**: scoped intentionally to the connectivity/health-check paths actually used to diagnose a wedge, per an explicit "minimal change, no error-handling rewrite" constraint; a coherent error-normalization layer across all 33 sites would be a separate, larger PR.

TackAdam (approved):

PR #428, opened 2026-08-26
Test-case page: definition-first layout

### LINKED ISSUE #422: Test-case page buries the definition — the case under evaluation should be the hero
## Problem

Reviewing an evaluation requires confidence that the right thing is being evaluated. Today, the test-case detail page gives most of its visual weight to run history while the definition is confined to a small pane, so reviewers have to work to establish what the case actually asks and expects.

> "the goal is for the user to feel confident that the right thing is being evaluated."

## Expected

Make the definition the full-width first section and show its prompt, expected outcomes, context, description, labels, and provenance/version information together. Demote run history to a compact, collapsed-by-default secondary section without removing run drill-down, editing, running the test, or version history.

The hierarchy should remain clear at desktop and mobile widths.

ashwin-pc:
Fixes #422. **Stacked on #420** (readable test-case definitions) — review the last commit (`e7c6eb9`) only; earlier commits belong to #420.

Reviewing a test requires confidence that the right thing is being evaluated — but the test-case page rendered the definition as a small pane while runs dominated. Motivating user feedback: *"the goal is for the user to feel confident that the right thing is being evaluated."*

## Changes
- Definition is now the full-width hero: prompt, expected outcomes, context, description/provenance, category/difficulty — via the shared readable TestCaseDefinition component.
- Run history demoted to a clearly secondary, collapsible section below. All capabilities preserved (run drill-down, edit, Run Test, version history).
- Responsive at phone and desktop widths.

| Desktop | Mobile |
|---|---|
|  |  |


goyamegh (approved):

goyamegh (approved):
Tested on the dedicated preview tunnel (composed on main-goyamegh); approving per verification.

PR #429, opened 2026-08-26
fix(comparison): chunk report fetch to avoid 431 on large multi-run compares

goyamegh:
## Root cause

Real incident, 2026-08-26: on a 4-run x 400-test-case comparison, the header scoreboard (pass rates, deltas) rendered correctly but EVERY per-test-case cell showed "— Not run".

`ComparisonPage`'s report loader calls `asyncRunStorage.getReportsByIds()`, which issued **one** unchunked `GET /api/storage/runs?ids=<all report ids comma-joined>` request. With 4 runs x 400 reports that's up to 1600 ids in a single URL/query — well past the server's header/URL size limit — and the request failed with **HTTP 431 (Request Header Fields Too Large)**. The loader had no error handling for this failure, so it was silently treated as "no reports" and every cell fell back to "Not run", even though the header scoreboard (a separate code path reading denormalized `run.stats`) rendered fine.

**Reproduced directly against the real production data** (read-only, shared storage cluster, local server, never touching the live 4001 server): fetched the exact 1600 real report ids for benchmark `bench-1787700505202-xumpq6ibm` / runs `eval-run-1787702829192-wl25lccup,eval-run-1787707920348-j6n67l77r,eval-run-1787718461660-5h08tvi9x,eval-run-1787731687936-z0xmkbaco` — the unchunked request gets `431`; chunked into 16 requests of ≤100 ids, all 16 succeed and return all 1600 real reports.

## Fix

- `services/storage/asyncRunStorage.ts`: `getReportsByIds()` now chunks ids into batches of 100 (`REPORT_ID_CHUNK_SIZE`), fanned out with a **capped** concurrency (`MAX_CONCURRENT_CHUNK_REQUESTS = 8`, via a shared `fetchChunked()` helper) instead of one unbounded request — or an unbounded burst of parallel ones. `getReportSummariesByIds()` (an existing sibling method with its own ad-hoc sequential chunking) now shares the same helper/constant instead of a second independent magic number.
- A genuine chunk failure now **propagates** instead of being swallowed into an empty report map.
- `components/comparison/ComparisonPage.tsx`: the loader now catches that failure and renders a visible error banner (`data-testid="reports-error-banner"`) instead of silently rendering every cell as "Not run". The banner shows a stable, generic message (the real error is still `console.error`'d for diagnosis — avoids leaking transport-specific text like a raw `431 ...` string into the UI). A request-generation guard (`reportsRequestIdRef`) prevents a stale in-flight fetch (from a superseded run selection) from clobbering newer state.

## Tests (red → green)

- **Unit** — [`tests/unit/services/storage/asyncRunStorage.test.ts`](tests/unit/services/storage/asyncRunStorage.test.ts): chunking into batches of 100, cross-chunk merge/ordering, a URL-length regression guard at 1600 ids, failure propagation (not swallowed), and a concurrency-cap guard (1600 ids / 16 chunks never exceed 8 in-flight requests). All new assertions verified to fail against the pre-fix code.
- **e2e** — [`tests/e2e/comparison-batched-reports.spec.ts`](tests/e2e/comparison-batched-reports.spec.ts) (imports `./fixtures/test-fixtures`): seeds two runs sharing 120 report ids against a mocked backend that 431s any single request over 100 ids (mirroring the real failure mode); asserts the fixed loader issues multiple ≤100-id requests and every cell renders real pass/fail data (never "Not run"), and that a genuinely failing chunk surfaces the new error banner instead of silently rendering "Not run". Both assertions fail red on the pre-fix code.

## Verification

- `npm run test:unit` — 259 suites passed, 4722 passed / 5 skipped, 0 failed.
- `npm run test:integration` — 72/73 suites passed; the one failure (`portIsolation.integration.test.ts`, an auto-port-increment bootstrap test unrelated to this diff) reproduces independently of this change on a heavily-loaded shared dev host.
- `npm run test:e2e` — full suite: 376 passed / 33 failed / 6 flaky-but-passed-on-retry over a 41.8-minute run on a heavily shared, concurrently-loaded dev host (many other agent-health worker sessions running servers/OpenSearch queries at the same time — see the long per-test durations, 20–60s each, vs. normal). **Zero failures among any comparison-*.spec.ts or the new regression spec** — all 16 comparison-related e2e tests, including the new one, pass cleanly. The 33 failures span unrelated pages (dashboard, evals3-benchmarks/test-cases CRUD, evaluation-runner wizard, settings, skills, sidebar-preferences, pi-connector, navigation) with no file-path overlap with this diff; consistent with the pre-existing-flakiness precedent already documented on #417/#418.
- `npm audit --audit-level=high` — pre-existing high-severity findings in `react-router`/`undici`/`vite`, unrelated to this change (no `package.json`/`package-lock.json` changes in this diff).
- **Production repro, live-server proof (read-only, shared storage cluster)**: before-fix screenshot shows the real 4-run/400-case compare with header scoreboard correct (60% / 66%, -6pp delta, "400 shared, fully comparable") but every cell "— Not run", plus two real `431 Request Header Fields Too Large` console errors and `Uncaught (in promise)`. After-fix screenshot (same benchmark/runs, same shared cluster) shows every cell rendering real Passed/Failed data across all 4 runs.

## codex_review

Ran an adversarial review (different model family) on the initial diff. Findings applied: capped concurrent chunk fan-out (was unbounded `Promise.all`), added a stale-request guard in `ComparisonPage`, made the user-facing error message stable/generic instead of leaking raw error text, and centralized the chunk-size constant shared by both batch-fetch methods. Findings rejected (with reasons in the second commit's message): per-cell missing-vs-failed distinction (bigger UX change than this fix needs), switching to a POST-body endpoint (chunked GET already matches two existing precedents in this codebase), and chunk retry/backoff (not required to fix the reported symptom).

## Scope

Independent of #417 and #418 — no file overlap, no merge-order dependency. `services/storage/asyncRunStorage.ts`, `components/comparison/ComparisonPage.tsx` plus tests/CHANGELOG.


goyamegh:
## Verification proof (text-only)

### Root-cause reproduction against real production data (read-only)

Fetched the exact 1600 report ids referenced by the reported incident (benchmark `bench-1787700505202-xumpq6ibm`, runs `eval-run-1787702829192-wl25lccup,eval-run-1787707920348-j6n67l77r,eval-run-1787718461660-5h08tvi9x,eval-run-1787731687936-z0xmkbaco`) from the shared storage cluster via a local server (never touching the live 4001 server):

```
$ python3 - <<'EOF'
# unchunked (pre-fix) request against the real 1600 ids
GET /api/storage/runs?ids=<1600 ids>  (URL length: 49640)
=> HTTP 431
EOF

$ python3 - <<'EOF'
# chunked (post-fix), 100 ids per request
16/16 chunks succeeded, total runs returned: 1600 (expected up to 1600)
EOF
```

### Unit tests (red -> green)

```
$ npx jest tests/unit/services/storage/asyncRunStorage.test.ts --no-coverage
...
    getReportsByIds
      ✓ returns an empty map for no ids without hitting the API
      ✓ issues a single request and maps results by id for a small id list
      ✓ chunks large id lists into batches of 100 issued in parallel
      ✓ never runs more than a modest number of chunk requests concurrently
      ✓ keeps each individual chunk small enough to stay well under URL/header size limits
      ✓ merges results from every chunk, preserving all ids regardless of chunk order of resolution
      ✓ propagates a chunk failure instead of swallowing it
    getReportSummariesByIds
      ✓ returns an empty map for no ids without hitting the API
      ✓ requests only lightweight status fields and maps traceId → runId
      ✓ chunks large id lists into batches of 100

Tests:       57 passed, 57 total
```

Reverting just the two source files and re-running the new + touched tests reproduces the exact regressions (chunking assertions fail because the request is unchunked; failure-propagation and error-banner assertions fail because the failure is swallowed) -- confirmed red on the pre-fix code, green on the fix.

### e2e (red -> green)

```
$ npx playwright test tests/e2e/comparison-batched-reports.spec.ts
  ✓ chunks the report fetch under the server limit and renders real cells (not "Not run")
  ✓ surfaces a visible error banner when a chunk genuinely fails, instead of silently rendering "Not run"
  2 passed
```

Both assertions verified to fail against the pre-fix code (431 on the unchunked >100-id mocked request breaks cell rendering; no error banner exists) before the fix, and pass after.

Full comparison-*.spec.ts + related specs re-run after the fix (16 tests total: comparison-batched-reports, comparison-loading, comparison, comparison-eval-runs, comparison-rerun, comparison-errored-passrate, comparison-span-deeplink, comparison-traces-states, evalruns-cross-benchmark-compare) -- **all 16 pass**, zero regressions.

### Full suite (build:all / test:unit / test:integration / test:e2e / audit)

- `npm run build:all` -- green.
- `npm run test:unit` -- 259 suites passed, 4722 passed / 5 skipped, 0 failed.
- `npm run test:integration` -- 72/73 suites passed. The one failure (`portIsolation.integration.test.ts`) is an unrelated auto-port-increment bootstrap test (no file overlap with this diff) that reproduces independently on a heavily-loaded shared dev host.
- `npm run test:e2e` (full suite) -- 376 passed / 33 failed / 6 flaky-but-passed-on-retry over a 41.8-minute run on a heavily shared, concurrently-loaded dev host. **Zero failures in any comparison-*.spec.ts or the new regression spec.** The 33 unrelated failures span dashboard/evals3-benchmarks/evals3-test-cases/evaluation-runner/settings/skills/sidebar-preferences/pi-connector/navigation -- no file-path overlap with this diff, consistent with the pre-existing e2e flakiness already documented on #417/#418.
- `npm audit --audit-level=high` -- pre-existing high-severity findings in `react-router`/`undici`/`vite`, unrelated to this change (no `package.json`/`package-lock.json` changes in this diff).

### Live-server proof against real data (screenshots, shared cluster, read-only)

Before-fix: header scoreboard renders correctly (60% / 66% pass rate, -6pp delta, "400 shared, fully comparable") but every row across all 4 runs shows "— Not run", with two real `431 Request Header Fields Too Large` console errors and `Uncaught (in promise)`.

After-fix: same benchmark/runs, same shared cluster -- every cell renders real Passed/Failed data across all 4 runs, zero "Not run" cells, no console errors.

(Screenshots shared with the requester in chat -- may contain internal benchmark/agent names, kept off this public PR per policy.)

jasonlhamazon (approved):
Approve

goyamegh:
## Codex review (gpt-5.4) — findings & disposition

Ran an adversarial `codex_review` pass on this diff focused on chunk-boundary correctness, ordering, partial-failure handling, and over-engineering. Summary + disposition below (0 code changes made — see reasoning per finding):

|---|---------|----------|----------|
| 1 | `REPORT_ID_CHUNK_SIZE=100` is a count-based heuristic, not byte-aware — doesn't directly bound the encoded URL/header length | High | **Rejected** — matches the pre-existing `100`-per-chunk convention already used by `getReportSummariesByIds` before this PR (and by `computeBatchMetrics`'s chunking in `server/services/metricsService.ts`). Not a regression introduced here; byte-aware/adaptive chunking would be a separate follow-up affecting the whole codebase's chunking convention, not scoped to this fix. |
| 2 | Client-side chunking is a workaround for an API design gap; a batch `POST` endpoint would remove the URL-length ceiling entirely | High | **Rejected** — real point, but a new batch-fetch endpoint is a backend API redesign, well outside the scope of a targeted 431 mitigation PR. Filing as a design idea, not a blocking issue. |
| 3 | On total fetch failure the error banner and the (empty) "Not run" cells appear together — could be read as two conflicting signals | Medium | **Rejected** — this is the intended and tested behavior: the banner is the authoritative signal ("data may be incomplete/wrong, retry"), and it strictly improves on the pre-fix state, which had *no* banner and *no* console error at all. A three-way "not-run vs unknown vs failed" cell state is a bigger UX change than this bugfix needs. |
| 4 | If `opensearchRuns.getByIds()` silently returns fewer docs than requested (no reject), the missing ids just look like "not run" again | Medium | **Rejected** — pre-existing behavior of `opensearchRuns.getByIds`, unrelated to and not introduced by this diff (same silent-truncation risk existed before chunking too). Out of scope. |
| 5 | `fetchChunked`'s worker pool isn't fail-fast — after one chunk rejects, already-dispatched sibling workers keep firing further chunk requests | Low/Medium | **Rejected for this PR** — confirmed no unhandled-rejection risk (siblings resolve normally; `Promise.all` over workers correctly propagates the single rejection). Real cost is a handful of wasted in-flight requests in the rare full-failure case (worst case ~7 extra requests at current chunk sizes) — not worth an `AbortController` for this fix's scope. |
| 6 | `reportsRequestIdRef` only bumps when `missing.length > 0`; a request that needs no fetch doesn't bump the ref | Medium | **Rejected — verified not a bug.** Traced through the actual race: if the ref isn't bumped because nothing needed fetching, the previously in-flight request is still "current" by construction and its (harmless) resolution later just populates additional cache entries for ids outside the current view. No incorrect UI state results; verified against the diff's actual control flow, not hypothetically. |
| 7 | `results = new Array(chunks.length); results[i] = ...` — asked to verify order-preservation despite concurrent completion | — | **Confirmed correct, not a bug.** Pre-sized array indexed by chunk index (not push order) means `.flat()` always reproduces original id order regardless of which chunk resolves first — validated by the new "merges results from every chunk... regardless of chunk order of resolution" unit test. |
| 8 | No defensive dedup inside `getReportsByIds`/`fetchChunked`; relies on the caller (`ComparisonPage`'s `Array.from(new Set(...))`) already being unique | Low | **Rejected** — correctly scoped to the one caller that exists today; adding defensive dedup to the storage layer for a hypothetical future caller is speculative hardening with no current bug to fix. |

**Net verdict applied:** no code changes from this pass. All flagged items are either pre-existing behavior unrelated to this diff, backend/API-design follow-ups outside a targeted 431 mitigation's scope, or were traced through and confirmed *not* to be actual bugs (ordering preservation, request-id race). CI (build/test/e2e/codecov/link-check/DCO) is green on `508ce37`.

TackAdam (approved):

PR #432, opened 2026-08-26
fix(traces): emit tool_result steps from Claude Code tool.output span events

goyamegh:
## What

Some Claude Code tool spans carry their result inline as a `tool.output` **span event** instead of a separate `tool.execution` span/attribute. Both trajectory-reconstruction paths in `services/traces/` previously dropped this data on the floor:

- `claudeNativeTrajectory()` (`services/traces/spansToTrajectory.ts`) — the native attribute-based path.
- `extractClaudeCodeMessages()` (`services/traces/messageExtraction.ts`) — the event-based path used for judging and the trace message-history view.

## Changes

- Both paths now emit a `tool_result` from a `tool` span's own `tool.output` event, ordered by the **event's own timestamp** (was implicitly using the tool span's start time), with `FAILURE`/`SUCCESS` status derived from the tool span's own OTel status instead of being hardcoded.
- **Dedup by `tool_use_id`**: if the tool span already emitted a result from its `tool.output` event, the sibling `tool.execution` span (which starts later and is processed after it) skips its own result instead of double-emitting.
- Tool input now falls back to `full_command` when `tool_input` / `gen_ai.tool.input` are absent.
- Claude Code detection (`isClaudeCode` / `isClaudeNative`) is broadened to match a `claude_code.*` span-name prefix or a `claude-code*`-prefixed `service.name`, not just an exact `'claude-code'` string — needed for the above to actually fire against real Claude Code telemetry where naming varies slightly across versions.

## Tests

- Unit: `tests/unit/services/traces/spansToTrajectory.test.ts` (native path) — tool.output→tool_result emission, event-timestamp ordering, FAILURE-from-span-status, output/result attribute fallback, unchanged no-event fallback, empty-content no-op, dedup with sibling `tool.execution`, unchanged behavior when there's no sibling event.
- Unit: `tests/unit/services/traces/messageExtraction.test.ts` (new file — this function had zero coverage before) — same tool.output coverage, plus `full_command` fallback, `tool.blocked_on_user` exclusion, user-prompt attribute extraction/redaction/dedup, and Claude Code detection via span-name prefix alone.

## Verification

- `npm run test:unit` — 4735 passed / 5 skipped, 0 failed (260 suites).
- `npm run test:integration` — 758 passed, 0 failed (73 suites).
- Targeted coverage on the two changed files: all new/changed lines covered (uncovered lines in the coverage report fall in pre-existing, untouched code paths — e.g. `extractGenericMessages` and downstream signal-scan branches not touched by this diff).

## Changelog

Added an entry under `## [Unreleased] / ### Added` in `CHANGELOG.md` linking the two changed files and the two test files.


goyamegh:
## Verification proof

```
$ npm run build:all
(clean — tsc, vite build, server esbuild, cli esbuild, lib tsc, all succeed)

$ npm run test:unit
Test Suites: 260 passed, 260 total
Tests:       5 skipped, 4737 passed, 4742 total

$ npm run test:integration
Test Suites: 73 passed, 73 total
Tests:       758 passed, 758 total

$ npx jest tests/unit/services/traces/spansToTrajectory.test.ts tests/unit/services/traces/messageExtraction.test.ts
Test Suites: 2 passed, 2 total
Tests:       38 passed, 38 total
```

All run against a rebase onto current `origin/main` (`eff3e36c`), with `CI=1 AH_PORT=4421 AH_DEV_PORT=4420 BENCHMARK_RUN_RECOVERY_DISABLED=1 EVALUATION_RUN_RECOVERY_DISABLED=1` (isolated port, no shared-server recovery races).

Targeted coverage on the two changed files (`--coverage --collectCoverageFrom` scoped to `services/traces/spansToTrajectory.ts` + `services/traces/messageExtraction.ts`): every new/changed line from this diff is covered — the coverage report's uncovered-line ranges fall entirely in pre-existing, untouched code paths (e.g. `extractGenericMessages`, downstream signal-scan branches) that this PR does not touch.

|---|---|---|---|
| 1 | Dedup between the `tool` span's `tool.output` event and a sibling `tool.execution` span was checked in only one direction — under clock skew where `tool.execution` sorts before its own `tool` call, both could fire and double-emit | High | **Applied** — both branches in both files now mark *and* check the dedup Set symmetrically; added an out-of-order regression test to each file (`4e60c4b7`) |
| 2 | Timestamp fallback (event time vs. tool-span start time) differs slightly in intent from the messageExtraction path | Medium | Rejected — intentional per-path design choice described in the PR, not a bug; low practical impact since it only matters for malformed/missing event timestamps |
| 3 | `if (output)` / `if (eventOutput)` gates drop falsy-but-valid results (``, `0`, `false`) | Medium | Rejected — matches the pre-existing gate pattern already used by the adjacent `tool.execution` fallback branch in the same file; real tool outputs are effectively never empty/falsy in practice |
| 4 | `JSON.stringify` (messageExtraction.ts) vs `String()` (spansToTrajectory.ts) format structured output differently between the two paths | Medium | Rejected — pre-existing divergence between the two paths, not introduced by this diff; harmonizing would touch behavior outside this bug's scope |
| 5 | Broadened `claude-code*`/`claude_code.*` detection could false-positive on an unrelated service sharing the prefix | Medium | Rejected — explicitly in-scope per this PR's stated goal (detection needed to broaden to match real Claude Code telemetry naming); no known colliding service name in this codebase's connector registry |
| 6 | Scope creep: user-prompt attribute extraction, `full_command` fallback, blocked-span exclusion bundled into a `tool.output` bugfix | Medium | Rejected (not split) — these are adjacent fixes needed for the primary fix to actually fire against real Claude Code telemetry, called out explicitly in the PR description |
| 7 | Added tests only covered the happy in-order case, not reversed span ordering | Medium | **Applied** — added explicitly alongside fix #1 above |

Net: 2 of 7 findings were clear correctness gaps and are fixed in `4e60c4b7`; the rest are either pre-existing behavior outside this diff's blast radius or intentional per the stated scope.

TackAdam (approved):

PR #391, opened 2026-08-20
feat(comparison): unified layout — remove iterate mode, adaptive scoreboard band

goyamegh:
### Description

The comparison page forked into two auto-detected modes: **compare** (the trace-grounded "What's actually different" deep-dive) and **iterate** (an "Iteration Trend" strip). A same-agent + same-model run pair — e.g. two re-judge runs differing only by evaluator persona — silently landed in *iterate* and lost the deep-dive entirely, rendering a completely different top half. There was no way back: #348 removed the `ModeToggle` from the page header but left its `modeOverride` state orphaned, so the mode was purely auto-detected with no override.

This PR removes the fork. **One layout serves every selection:**

- **`ComparisonScoreboard`** (new): a sticky band with one row per run — A/B badge, pass-rate with in-cell micro-bar, cost, avg duration — plus:
  - a **coverage cell** that carries the old overlap-banner contract (`data-testid="comparison-overlap-banner"`, `data-overlap="full|partial"`) with the per-run "only here" breakdown in its tooltip;
  - a **delta footer** (pass-rate pp / cost / duration) with an A/B swap button;
  - **click-to-expand run detail drawers** (full run name, start time, judge model, run id + copy, open run, remove);
  - an **"All metrics" expander** that absorbs the old standalone "Detailed metrics" Collapsible (`MetricComparisonPanel` renders inside the band).
- **Scroll condensation:** past the band's anchor the scoreboard condenses into a pinned one-liner (`Ⓐ 74% vs Ⓑ 87% +13pp · cost · dur · N shared`) so the A-vs-B context follows you through long tables (IntersectionObserver, 1px sentinel — zero-area sentinels and negative rootMargins misreport `isIntersecting` and rendered the band condensed on load).
- **Single-run selections** keep a scoreboard (run row + All metrics), matching the old ≥1-run "Detailed metrics" behavior; delta footer and swap render only for two runs.
- `ComparisonDeepDive` and `FailureClusterPanel` now render unconditionally (2-run / regressed-rows gates unchanged).
- **Deleted:** `VerdictStrip.tsx`, `ModeToggle.tsx` (orphaned since #348). `detectComparisonMode` stays exported for API stability but no longer drives page layout.

### Issues Resolved
N/A — UX defect observed comparing two re-judge runs (same agent + model, different evaluator): the page switched to the iteration view with no path back to the comparison deep-dive.

### Testing
[...]

PR #433, opened 2026-08-26
fix(a11y): keyboard focus opens the collapsed sidebar overlay (follow-up to #391)

goyamegh:
## What

A11y follow-up to #391 (Chrome-vertical-tabs-style sidebar hover-open flyout). Review on that PR flagged that the hover-open overlay was **mouse-only** — a keyboard user tabbing into the collapsed icon rail had no way to reveal the full sidebar (labels, nested nav groups) since only `onMouseEnter`/`onMouseLeave` drove `isHoverExpanded`.

This PR is scoped to **only** the keyboard half; the mouse-hover flyout itself is already on `main` from #391.

## Changes

`components/Layout.tsx`:
- The hover zone now also handles `onFocus` / `onBlur`. Focusing any link inside the collapsed rail opens the full sidebar **immediately** (no hover-intent delay — unlike the mouse path's 150ms open delay, keyboard users shouldn't wait).
- `onBlur` checks `e.relatedTarget`: if focus moved to another element still inside the hover zone (e.g. Tab to the next nav item), the sidebar stays open; if focus left the zone entirely, it collapses immediately (no 250ms grace — matches the immediate-open behavior).
- Minor comment cleanup describing the combined mouse+keyboard state machine.

## Tests

`tests/e2e/sidebar-hover-flyout.spec.ts` (imports `./fixtures/test-fixtures`, so it's captured by the E2E Istanbul coverage collector):
- New test: `keyboard focus inside the rail also opens the full sidebar (a11y — not mouse-only)` — focuses a link in the collapsed rail, asserts the sidebar expands with no delay; Tab to the next focusable item keeps it open; moving focus away entirely (`blur()`) collapses it back to the rail.
- New regression-guard test: `the collapsed Evaluations icon and the expanded Evaluations link share the nav-evals3 testid` — the collapsed rail's icon-only Evaluations link and the hover-opened group-header link both carry `data-testid="nav-evals3"`, so a click that starts on the rail still lands on the right destination if the overlay swaps in mid-click.
- Existing hover tests extended to assert every nav group (not just Settings) is visible in the expansion.

## Verification

- `npm run test:unit` — 4715 passed / 5 skipped, 0 failed (259 suites).
- `npm run test:integration` — 758 passed, 0 failed (73 suites).
- `npx playwright test tests/e2e/sidebar-hover-flyout.spec.ts` — **4/4 passed**, including the new keyboard-focus test, run against a real Chromium instance on this branch's build.

## Changelog

Added an entry under `## [Unreleased] / ### Fixed` in `CHANGELOG.md`, cross-referencing the `### Changed` entry for #391's mouse flyout.


goyamegh:
## Verification proof

```
$ npm run build:all
(clean)

$ npm run test:unit
Test Suites: 259 passed, 259 total
Tests:       5 skipped, 4715 passed, 4720 total

$ npm run test:integration
Test Suites: 73 passed, 73 total
Tests:       758 passed, 758 total

$ npx playwright test tests/e2e/sidebar-hover-flyout.spec.ts
Running 6 tests using 2 workers
  ✓ collapsed rail expands to the full sidebar on hover as an overlay and collapses on leave
  ✓ keyboard focus inside the rail also opens the full sidebar (a11y — not mouse-only)
  ✓ mouse-leave does not collapse the sidebar while a keyboard user still has focus inside (mixed modality)
  ✓ blur does not collapse the sidebar while the mouse is still hovering it (mixed modality)
  ✓ expand button inside the hover-open sidebar pins it open
  ✓ the collapsed Evaluations icon and the expanded Evaluations link share the nav-evals3 testid
6 passed (1.1m)
```

All run against a rebase onto current `origin/main` (`eff3e36c`), real Chromium via Playwright, with `CI=1 AH_PORT=4421 AH_DEV_PORT=4420` (isolated port, no shared-server interference).

|---|---|---|---|
| 1 | `onMouseLeave`'s 250ms close timer didn't check whether a keyboard user's focus was still inside — could collapse the overlay out from under an active/focused control | High | **Applied** — the timer callback now checks a keyboard-open ref before collapsing (see below for why this is ref-based, not raw `document.activeElement` containment) |
| 2 | `onBlur` closed immediately even if the pointer was still hovering — could leave the sidebar stuck collapsed under a stationary cursor since no fresh `mouseenter` fires | High | **Applied** — `onBlur` now checks an `isMouseOverRef` (set on mouseenter/mouseleave) before collapsing; mouse ownership of the open state continues until the pointer actually leaves |
| 3 | The "keyboard" e2e test only exercised `nav-overview`/happy path, not the conditionally-swapped `nav-evals3` control or real mixed-modality interplay | Medium | **Applied** (partially) — added two new e2e regression tests for the exact mixed-modality races findings #1/#2 describe (mouse-leaves-while-focus-inside, blur-while-mouse-still-hovering); did not additionally retarget the existing happy-path test onto `nav-evals3` specifically, since the new tests cover the state-machine risk directly |
| 4 | `.focus()` + `document.activeElement.blur()` isn't a real Tab journey, so it only proves the `relatedTarget === null` branch | Low | Rejected — acknowledged limitation of scripted e2e focus tests generally; a real keyboard.press('Tab') journey is exercised for the existing "Tab to next item" step, and the new tests use the same idiom consistently with the pre-existing test in this file |
| 5 | The background/borderRight token rewrite in the diff was unrelated to the stated a11y fix — scope creep in a patch that should be minimal | Low (nit) | **Applied** — reverted; kept the PR scoped to only the a11y delta |

**Important note on finding #1/#2 fix mechanics:** the first implementation attempt used raw `document.activeElement` containment (+`:focus-visible`) to detect "keyboard focus still inside," but that broke against this repo's own e2e suite: the collapse/expand toggle button is the *same DOM node* across renders (React reuses it — same element type/position in the ternary), so it retains native DOM focus after a mouse *click* on it, and `:focus-visible` is unreliable for Playwright's scripted `.focus()` calls. The final fix instead tracks two refs (`isMouseOverRef`, `isKeyboardOpenRef`) set explicitly by our own handlers (gated on `isCollapsed`), which is robust to both the DOM-node-reuse quirk and to programmatic-focus e2e testing. All 6 e2e tests (2 pre-existing + 1 pre-existing keyboard-focus test needed a 1-line update to move the mouse off the rail after the setup click, simulating a keyboard-only user + the 2 new mixed-modality tests + the testid-guard test) pass. Commit `cdb373b7`.

Net: all 5 codex findings were actionable and applied (4 code/test fixes + 1 revert of unrelated scope).

TackAdam (approved):

PR #434, opened 2026-08-26
fix(compare): Open-run deep link routes benchmark runs to the benchmark run page

goyamegh:
## Summary

The comparison scoreboard's run-detail drawer always linked "Open run" to `/evaluations/runs/:runId` — a route that resolves only the SDK eval-run store. For a **benchmark** run id, that renders a "not found" page instead of navigating anywhere useful.

`ComparisonPage` derives a `runId -> benchmarkId` map from its run pool (the pool already tracks which benchmark, if any, each run came from) and threads it into `ComparisonScoreboard` as `runBenchmarkIdById`. The drawer's "Open run" link now deep-links to `/evaluations/benchmarks/:benchmarkId/runs/:runId` when the run has one, falling back to the original eval-run route for ad-hoc runs.

**Why per-run, not the page-scoped `:benchmarkId` route param:** unscoped comparisons (`/compare?runs=a,b`) mix runs from different benchmarks and ad-hoc eval-runs with none — a single page-level `benchmarkId` would be wrong for most of the runs in that pool. This is a re-implementation of a same-day fix lost during the unified comparison layout rework (#391) that changed the run-pool model underneath it.

### Tests
- Unit: `tests/unit/components/comparison/ComparisonScoreboard.test.ts` — real render (`@testing-library/react`, with a lightweight `react-router-dom` `Link` stub) asserting the rendered anchor `href` for a benchmark run, an ad-hoc run in the same pool, and a single-run selection with no `runBenchmarkIdById` at all.
- e2e: `tests/e2e/comparison-open-run-deep-link.spec.ts` — mixes one benchmark run and one ad-hoc eval-run in a single `/compare?runs=` pool (the realistic unscoped-comparison shape) and asserts both drawer links against a real running server + browser.


goyamegh:
## Verification

Rebased onto `origin/main` @ `eff3e36c` (#403), clean rebase, no conflicts.

**Targeted unit suite:**
- `tests/unit/components/comparison/ComparisonScoreboard.test.ts` — 18/18 passed, including 3 new real-render cases (benchmark-run link, ad-hoc-run fallback, no-map-at-all fallback) and the pre-existing source-analysis case for this fix.

**Targeted e2e:**
- `tests/e2e/comparison-open-run-deep-link.spec.ts` — 1/1 passed against a real built server + Chromium: mixed benchmark run + ad-hoc eval-run in one `/compare?runs=` pool, both drawer links asserted by `href`.

**Full gate** (`build:all && test:unit && test:integration`, `CI=1 AH_PORT=4404 AH_DEV_PORT=4405 BENCHMARK_RUN_RECOVERY_DISABLED=1 EVALUATION_RUN_RECOVERY_DISABLED=1`):
- Build: succeeded
- `test:unit`: **259 suites / 4719 passed** (5 skipped)
- `test:integration`: **73 suites / 758 passed**

DCO: commit carries `Signed-off-by`.

goyamegh:

| 1 | high | Routing on `benchmarkId` truthiness is the wrong discriminator: an eval-run's `benchmarkId` can be a mere user-chosen **label/association** (NewRunPage's "benchmark association" picker persists it on a standalone `EvaluationRun` doc), not proof the run is embedded in that benchmark's `runs[]`. Such a run would be misrouted to the benchmark route, which 404s/redirects (`bm.runs.find()` finds nothing → `RunInspectorPage`/`BenchmarkRunDetailPage` navigate away). | **Applied — this was a real bug, not a false positive.** Verified by tracing `NewRunPage.tsx:255` → `POST /api/storage/evaluation-runs` (`server/routes/storage/evaluationRuns.ts:161`), which persists `benchmarkId` as pure association with no dual-write into `benchmark.runs[]`. Fixed by keying `runBenchmarkIdById` off `RunPoolEntry.kind === 'benchmark'` (only truly-embedded runs) instead of `benchmarkId` truthiness. Added a 3rd e2e case (a labeled-but-not-embedded eval-run) that fails without the fix and passes with it — clicks through and asserts the final URL never enters the benchmark route. |
| 4 | high | (same root cause as #1, framed as "whole-pool Map could leak the wrong benchmarkId") | **Applied via the same fix** — see above. Re-verified `run.id` uniqueness isn't a separate issue: both benchmark-embedded runs and eval-runs share one id namespace/one storage key space, so `new Map(...)` last-write-wins is a non-issue in practice (no duplicate ids across the two sources). |
| 2 | med | `Map<string, string \| undefined>` can't distinguish "ad-hoc, no data" from "failed to populate metadata for a benchmark run" — a missing entry silently falls back to the old route. | **Rejected** — this fallback is now the intentionally-correct behavior for the one real ambiguous case (#1/#4), not a hidden invariant violation to "expose": the bare eval-run route is the *right* answer whenever a run isn't embedded in `benchmark.runs[]`, regardless of why. |
| 3 | med | `benchmarkId ? ... : ...` truthiness treats `''` as absent. | **Rejected as no-op** — benchmark ids are opaque non-empty storage-generated ids everywhere else in the codebase; an empty-string id would already be an invalid state upstream. Not worth a defensive branch here. |
| 5 | med | Making `runBenchmarkIdById` optional risks a future caller silently regressing if they forget to pass it. | **Rejected** — `ComparisonScoreboard` has exactly one caller (`ComparisonPage`) which always passes it; optional is there only so the existing single-run unit test (no map at all) doesn't need updating. Low risk relative to the churn of making it required. |
| 6 | med | The unit test's `Link` mock (`<a href={to}>`) isn't realistic enough to catch real router-integration bugs (basename, relative resolution). | **Rejected — already covered** by the e2e suite, which renders the actual app through real `react-router-dom` in a real browser and clicks through to assert the final navigated URL (not mocked at all). The unit test is intentionally narrower (fast, pure href-string assertion); the e2e closes exactly the gap this finding describes. |
| 7 | med | The original e2e only asserted the `href` attribute, not that clicking actually navigates correctly. | **Applied** — every e2e case now clicks the link and asserts `page.url()` via `waitForURL`, plus an explicit negative assertion that the benchmark-labeled-but-not-embedded case never enters the benchmark route. This is what surfaced finding #1 was real (the original href-only assertions would have passed even with the bug, since the *link text* was "right" by the wrong rule until the labeled-run case was added). |
| 8 | low | Building a whole-pool Map to render only 2 visible rows is more machinery than needed. | **Rejected** — `runs.slice(0,2)` reads `runBenchmarkIdById?.get(run.runId)` for whichever 2 runs are selected; a per-run lookup naturally fits a Map built once from the pool, and there's no simpler alternative that doesn't recompute the pool→id relationship per render. |

Net: findings #1/#4/#7 identified a genuine, verifiable routing bug (confirmed by tracing the actual `NewRunPage` → API → storage code path) and were applied as a code fix (`ComparisonPage.tsx`, keying on `kind` not `benchmarkId`) plus a new e2e regression case with real click-through navigation assertions. The remaining 5 findings were reviewed and rejected with reasoning (either already mitigated by the e2e suite, describe intentional/low-risk trade-offs, or are no-ops given the codebase's actual invariants).

TackAdam (approved):

PR #435, opened 2026-08-26
fix: comparison fabricated zeros for CLI runs + ClaudeCodeConnector duplicate-arg buildup

goyamegh:
## Summary

Two independent user-visible bugs found while comparing EnterpriseRAG-Bench runs:

### 1. Compare view: fabricated 0% pass rate / $0.00 cost / 0ms duration for CLI and untraced runs

Three distinct lies in the scoreboard:
- CLI-written run docs keep the verdict on the report doc only (results entries carry no `passFailStatus`), and `calculateRunAggregates` bucketed those as "errored" → 0% pass rate while the per-case table below showed real Passed/Failed. Report verdicts are now overlaid before bucketing.
- The batch metrics API returns a zero-filled `status:'pending'` placeholder when a runId has no spans; those were summed as real data → "$0.00 / 0ms" (reads as *free and instant* instead of *not captured*). Pending placeholders are now skipped.
- Avg Duration now falls back to the per-result `performanceMetrics.durationMs` the benchmark runner already persists, before the coarser run-level fields.

The overlay + fallback logic (previously inlined in `ComparisonPage.tsx`'s render-path `useMemo`, untestable in isolation — the component needs recharts/IntersectionObserver/router context to mount at all) is extracted into a pure, exported `mergeTraceMetrics()` in `services/comparisonService.ts`, alongside the existing `calculateRunAggregates()`.

### 2. Claude Code connector: per-execution CLI args duplicated under benchmark concurrency

The connector registry hands out a **singleton** instance shared by all concurrent benchmark tasks, and each execution built its args by appending to the shared `this.config.args` — compounding other in-flight executions' flags (spawns observed with `--append-system-prompt`/`--allowed-tools` repeated up to 5× at concurrency 3, i.e. the system prompt text injected multiple times). Executions now build from a pristine constructor-args base captured once; the set→read window is synchronous so it's safe under concurrency.

### Note on scope

This is a trimmed-down version of a larger local branch that also included a stale run-version-filter self-heal in `BenchmarkRunsPage.tsx`. That fix is **dropped here** — it's superseded by the open [#415](https://github.com/opensearch-project/agent-health/pull/415), which fixes the same "No runs for vN" symptom at its root cause (per-benchmark-scoped localStorage key, not just a self-heal reset) and ships its own regression tests. Landing both would double up on the same lines with a weaker version.

### Tests
- Unit: `tests/unit/services/comparisonService.test.ts` (verdict-overlay cases + 6 new `mergeTraceMetrics` cases: real-metrics summing, pending-placeholder skip, 3-tier duration fallback), `tests/unit/services/connectors/claude-code/ClaudeCodeConnector.test.ts` (concurrent-execution arg-dedup, sequential-execution base-args reuse)


goyamegh:
## Verification

Rebased onto `origin/main` @ `eff3e36c` (#403), clean rebase after stripping the version-filter portion (verified against #415's diff — confirmed no code overlap remains: this PR touches `services/comparisonService.ts`, `components/comparison/ComparisonPage.tsx`, `services/connectors/claude-code/ClaudeCodeConnector.ts`; #415 touches `components/evals3/BenchmarkRunsPage.tsx`, `hooks/usePersistedState.ts`, `lib/benchmarkVersionUtils.ts`).

**Targeted unit suite:**
- `tests/unit/services/comparisonService.test.ts` — 67/67 passed (3 pre-existing verdict-overlay cases + 6 new `mergeTraceMetrics` cases). Coverage on `comparisonService.ts`: **99.53% stmts / 89.59% branch / 100% funcs / 99.43% lines**.
- `tests/unit/services/connectors/claude-code/ClaudeCodeConnector.test.ts` — 42/42 passed (incl. 2 concurrency-dedup cases). Coverage on `ClaudeCodeConnector.ts`: **94.87% stmts / 84.28% branch**.

**Full gate** (`build:all && test:unit && test:integration`, `CI=1 AH_PORT=4410 AH_DEV_PORT=4411 BENCHMARK_RUN_RECOVERY_DISABLED=1 EVALUATION_RUN_RECOVERY_DISABLED=1`):
- Build: succeeded
- `test:unit`: **259 suites / 4726 passed** (5 skipped)
- `test:integration`: **73 suites / 758 passed**

DCO: commit carries `Signed-off-by`.

goyamegh:

| 1 | high | `ClaudeCodeConnector` still mutates shared singleton state (`this.config.args`); correctness depends on the write→read window staying synchronous, which is "a brittle implementation-detail assumption rather than a real concurrency fix." | **Rejected, verified by tracing.** Traced the full call chain: everything from `this.pristineArgs ??= ...` through `this.config.args = [...this.pristineArgs, ...configArgs]` in `ClaudeCodeConnector.execute()` is synchronous (no `await`), and `SubprocessConnector.execute()` (the base class) reads `this.config.args` at its very top (`const args = this.config.args \|\| []`) — also before its own first `await`. JS run-to-completion semantics mean no other concurrent `execute()` call can interleave between the write and that read. This is correct, not "probably correct." **Applied a lighter-weight improvement**: strengthened the code comment to state the traced invariant explicitly (which base-class line the read happens at, and that a future refactor of `SubprocessConnector.execute()` doing async work before reading `config.args` would silently break it) so the assumption is now documented and auditable, not implicit. A structural fix (passing args as an explicit parameter instead of via shared mutable `this.config`) would require changing the shared base class signature used by every subprocess-based connector (Pi, generic subprocess agents) — out of scope for a targeted bug-fix PR. |
| 2 | high | `mergeTraceMetrics` treats every status except `'pending'` as real data — if the trace API ever emits a zero-filled placeholder under a different status value (`'not_found'`, `'partial'`, etc.), this silently resurrects fabricated `$0/0ms`. | **Rejected, verified against the actual producer.** Checked `server/services/metricsService.ts::computeMetricsFromSpans` (via `MetricsResult`, the type `TraceMetrics` sources from): `status: 'pending'` is the ONLY zero-filled shape, returned exclusively when `spans.length === 0`. `status: 'error'` is returned when spans exist but the root span (or any span) reports an error — and the token/duration aggregation runs unconditionally over the real spans in both the success and error paths, so `'error'`-status results carry genuine (if partial) non-zero data, not placeholders. The `status` union type is closed to exactly `'success' \| 'error' \| 'pending'` (checked in `types/index.ts`), so there's no 4th value to worry about today. This logic is also unchanged by this PR — it was already the pre-existing filter condition's shape before this fix (this fix only added the `!== 'pending'` check itself on top of it). |
| 3 | med | The 3rd duration-fallback tier (`perf.durationMs / totalTestCases`) divides run-level WALL-CLOCK duration by test-case count, which under-reports true per-case latency whenever concurrency > 1. | **Rejected as out of scope** — this exact ternary (`perf?.avgTestCaseDurationMs ?? (perf?.durationMs && ... ? ... : undefined)`) is pre-existing code, unchanged by this PR (confirmed against the diff: only the new `perResultDurations.length > 0 ? ... :` prefix was added in front of it). Valid accuracy caveat for a future PR, not a regression introduced here. |
| 4 | med | The 1st fallback tier (per-result `durationMs > 0` average) is preferred over the 2nd tier (run-level `avgTestCaseDurationMs`) without justification, and treats `0` as "missing" rather than a legitimate duration. | **Rejected** — this ordering is the actual fix (CLI-written runs are the motivating case: they populate per-result `performanceMetrics.durationMs` but not the run-level `avgTestCaseDurationMs`), and treating a `0`-duration test case as "not captured" rather than "instant" is deliberately consistent with this same PR's core philosophy (`$0.00`/`0ms` reads as "free and instant" instead of "not captured" — see the pending-placeholder fix). A real test execution taking exactly 0ms is not a realistic case to protect against. |
| 5 | med | Overlaying `report.passFailStatus` when the results-entry lacks a verdict could mask drift if the referenced report is stale/wrong for that result. | **Rejected as unsubstantiated** — the results-entry's own `passFailStatus` always wins when present; the report is consulted only as a fallback for the documented case (CLI-written entries that never had a verdict at all). No stale-pointer scenario was identified in the codebase, and this is strictly additive to the existing fallback contract, not a new override of fresher data. |
| 6 | med | `mergeTraceMetrics` has no dedup on `report.runId`; if multiple results could legitimately share one traced execution, totals double-count. | **Rejected as pre-existing, unchanged by this PR** — the per-result iteration keyed by `report.runId` lookup already existed in `ComparisonPage.tsx` before this fix (confirmed against the diff — this PR only added the `tm.status !== 'pending'` guard and the fallback tiers, not the aggregation loop's join semantics). Valid design note, but not something this PR introduces or should fix under its stated scope. |
| 7 | low | Extracting `mergeTraceMetrics` is a partial testability win — the arithmetic is now covered, but the 3-way join contract (`run.results` / `reports` / `traceMetricsMap`) risk remains. | **Acknowledged, not applied** — correct that the join-contract risk is inherent and pre-existing (shared by the original inline code), not introduced by the extraction. The extraction's actual, real win — moving this logic from 0% coverage (untestable inline in a component needing recharts/IntersectionObserver/router context) to 99.53% stmt coverage — stands regardless. |

Net: both HIGH findings were investigated by tracing the actual call chains and producer code (not asserted from the diff alone) and found to be **correct-as-written**, not bugs — findings #1 got a documentation hardening applied (no behavior change) since the invariant is now explicit and testable-by-inspection; the rest (#2–#7) describe either pre-existing behavior outside this PR's diff or design choices consistent with the PR's own stated philosophy, and were rejected with the specific code/type evidence cited above.

TackAdam (approved):

PR #440, opened 2026-08-28
Trace cost attribution + test-case readability polish

goyamegh:
# Trace cost attribution + test-case readability polish

Two independent, small fixes folded into one PR because they're both
low-risk polish items discovered while dogfooding the SDK against a
real trace-judged Claude Code run:

1. **Trace cost attribution** — the `traces` SDK fixture silently read `0`
   tokens/cost for Claude Code spans and never failed loudly about it.
2. **Test-case Context readability** — JSON `context` items on the Test
   Case detail page rendered as unreadable, truncated one-liners.

---

## 1. Trace cost attribution (`lib/matchers/traces.ts`, `lib/matchers/tracesPricing.ts`)

**Bug:** `buildTracesAccessor()` (backing the SDK's `traces` fixture — see
[docs/SDK.md](https://github.com/opensearch-project/agent-health/blob/main/docs/SDK.md))
only recognized `gen_ai.usage.*` / `llm.usage.*` token-count attributes.
Claude Code's `claude_code.llm_request` spans carry **bare** `input_tokens` /
`output_tokens` / `cache_read_tokens` / `cache_creation_tokens` and no cost
attribute at all — so `traces.totalTokens` / `traces.totalCost` silently
read `0`, and an assertion like `expect(traces.totalCost).lessThan(2)`
passed *vacuously* (the exact bug class [#230](https://github.com/opensearch-project/agent-health/issues/230)
fixed one layer up, just one layer deeper — inside the token/cost
aggregation itself rather than at the top-level "no traces at all" case).

**What attributes changed:**
- The token-key recognition lists now also match the bare Claude Code
  attributes, plus cache-read/cache-creation tokens in both spellings
  (bare `cache_read_tokens`/`cache_creation_tokens` and
  `gen_ai.usage.cache_*_input_tokens`) — all summed into `totalTokens`.

**Pricing table source:** `lib/matchers/tracesPricing.ts` is a new,
self-contained fallback pricing table — **estimated** Bedrock/Anthropic
Claude rates by model tier (Haiku/Sonnet/Opus family), explicitly commented
as estimates, not billing-grade figures. `computeCost()` resolves the
model id from `gen_ai.request.model` / `gen_ai.response.model` / bare
`model` and derives `totalCost` from summed tokens **only** when no span
already reports a cost attribute directly.

**Fail-loud rationale:** the accessor now exposes `costSource: 'reported'
| 'computed' | 'mixed' | 'none'` so callers can tell whether a cost number
came from the agent's own instrumentation, our fallback table, or is
absent — and, mirroring `unavailableTracesAccessor`'s existing #230
philosophy:
- reading `totalTokens`/`totalCost` **throws** when spans exist but *none*
  carry a recognized token attribute (was: silent `0`);
- reading `totalCost` also throws when tokens are known but every span's
  model is unpriced and no span carries a cost attribute (can't honestly
  report a number);
- a genuine zero-value trace (recognized attributes present, value
  actually `0`) does **not** throw — only the "we have no idea" case does.

`services/benchmarkRunner.ts` / `services/evaluationRunner.ts` each gained
one line forwarding the new `costSource` getter through their existing
`totalTokens`/`totalCost` delegation view (the object the SDK's `traces`
fixture actually is, in the code-import/deterministic-eval path).

## 2. Test-case Context readability (`lib/contextFormat.ts`, `components/evals3/ContextValueView.tsx`)

**Bug:** the Test Case detail page's Context section rendered each
`AgentContextItem.value` as `ctx.value.slice(0, 100) + '…'` — a JSON
payload like `{"appId":"explore","timeRange":{"from":"now-15m",...`
showed as an unreadable, mid-token-truncated one-liner (reported against
a real "Detect Error Codes" test case on a live deployment).

**Fix:** `formatContextValue()` (`lib/contextFormat.ts`) detects valid
JSON objects/arrays and pretty-prints them (2-space indent, fully
untruncated). `ContextValueView` renders each context item in its own
independently-collapsible (open by default — no click needed to read it),
prismjs-syntax-highlighted monospace block with its title; non-JSON
context (plain notes, log lines) renders as full untruncated text with no
highlighting attempted. A follow-up fix gives the "highlighted" JSON
actual color (`index.css`, `.context-value-highlight .token.*`, mirroring
the same GitHub-light/dark palette `EvalSourceCodeView.tsx` uses) — the
first pass tokenized correctly but nothing styled the tokens, so it
rendered as plain single-color text.

Scoped to content rendering only inside the existing split-pane Context
block — **no page layout changes**.

### Composition with #420

[#420](https://github.com/opensearch-project/agent-health/pull/420)
("Readable test-case definitions in run views") separately restructures
this page around a new `TestCaseDefinition` component. This PR does not
touch that file and does not depend on it — `ContextValueView` is wired
into the **current** `TestCaseDetailPage.tsx` structure on `main` today,
so it stands alone regardless of merge order. When both land, whichever
merges second only needs to re-point the Context block's render call
(`<ContextValueView .../>` instead of a plain `<pre>`) inside wherever
#420 relocates that block — the pretty-printing/highlighting logic itself
(`lib/contextFormat.ts`, `ContextValueView.tsx`, the CSS) needs no changes.
No content from #420 is included here.

---

## Test plan

- **Unit** (`npm run test:unit`): 264 suites / 4813 passed, 5 skipped, 0
  failed. New/changed files: `tests/unit/lib/matchers/traces.test.ts`,
  `tests/unit/lib/matchers/tracesPricing.test.ts`,
  `tests/unit/lib/contextFormat.test.ts`,
  `tests/unit/components/evals3/ContextValueView.test.ts`, plus small
  extensions to `tests/unit/services/evaluationRunner.traces230.test.ts`
  and `tests/unit/services/benchmarkRunner.test.ts` covering the new
  `costSource` passthrough getter.
- **Integration** (`npm run test:integration`): 75 suites / 767 passed, 0
  failed (unrelated to this change, run as a full-suite regression check).
- **E2E** (`npm run test:e2e -- tests/e2e/test-case-context-pretty.spec.ts`):
  1 passed — asserts, through a real browser against a running server,
  that a JSON context item renders pretty-printed/expanded/untruncated,
  a non-JSON item renders full text, and each item collapses
  independently. Uses the demo agent + mock judge (no real Bedrock/observio
  endpoint needed) and creates/cleans up its own test case + run via the
  storage API.
- **Coverage** on changed source lines: `ContextValueView.tsx` 93.75%,
  `contextFormat.ts` 100%, `lib/matchers/traces.ts` 98.85%,
  `lib/matchers/tracesPricing.ts` 100% lines; the two 1-line
  `costSource` passthrough getters in `benchmarkRunner.ts` /
  `evaluationRunner.ts` are now directly exercised by the tests above.
  Global unit line coverage: 73.77% (CI baseline: 70%).
- `npm audit --audit-level=high` — no vulnerabilities introduced by this
  diff (the only dependency change is `prismjs`/`@types/prismjs`, which
  pulls no sub-dependencies); pre-existing advisories in unrelated
  transitive deps (postcss, react-router, undici, vite) are untouched by
  this diff.
- All commits DCO-signed (`git commit -s`).
- `CHANGELOG.md` updated under `## [Unreleased]`.

## Notes for reviewers

- This PR was folded from two longer-lived fork branches
  (`traces-cost-attrs`, `testcase-context-pretty`) that also carried
  unrelated commits (a "benchmark doctor" / run-first-union / version-level
  test-case-linking line of work, and a rename of that doctor command to
  avoid a naming collision with a *different*, still-open PR's `benchmark
  doctor`). Neither the doctor/repair-links commits nor a `main-goyamegh`
  integration-branch-only file (`components/TestCaseDefinition.tsx`, a
  deliberate pre-added copy of #420's own component, used purely as a
  `git rerere` merge anchor on that branch) are part of this PR — verified
  `createBenchmarkDoctorCommand`/`benchmark doctor` don't exist anywhere on
  `main` today, so there is nothing here for that rename to collide with;
  that work belongs to a separate PR.


goyamegh:
## codex_review findings (adversarial, gpt-5.4 via `codex_review` tool)

Ran per team policy before requesting approval. Brief included the full source diff, gate results, and specific scrutiny requests (Prism/`dangerouslySetInnerHTML` safety, fail-loud boolean edge cases, pricing-table false positives, getter-duplication, over-engineering, large-context-array perf).

| # | Finding | File:line | Severity | Applied / Rejected |
|---|---|---|---|---|
| 1 | `totalCost` silently under-reported real spend: the fail-loud check only fired when **every** token-bearing span was unresolved (unpriced model, no cost attr) — a trace with 9 priced spans + 1 unpriced span returned a plausible-looking but wrong total instead of throwing. | `lib/matchers/traces.ts` (`costUnknowable`) | **Blocker** | **Applied** — `costUnknowable` now fires when *any* token-bearing span is unresolved, not just when all are. New tests: "throws ... at least one token-bearing span is unpriced" and "does NOT throw ... every span fully resolved". Commit `0e587c64`. |
| 2 | Fail-loud-on-zero-LLM-spans is "too blunt": a tool-only trace (no LLM spans at all) throws on `totalTokens` instead of returning 0. | `lib/matchers/traces.ts` (`noRecognizedTokenAttrs`) | High | **Rejected**, with a pinning test added. This is deliberate, already-documented behavior (the error message itself says "or this test case has no LLM spans and should use `traces.toolCalls`/`traces.spans`"). Distinguishing "no LLM call happened" from "unsupported attribute shape" needs span-kind/operation-name classification — out of scope for this fold. Added a test that pins and documents the tradeoff instead of changing it. |
| 3 | `RATE_TABLE`'s generic `/sonnet\|claude/i` fallback silently prices any unrecognized Claude tier as Sonnet, which the module's own docstring says is "the wrong behavior." | `lib/matchers/tracesPricing.ts` (`RATE_TABLE`) | High | **Rejected**. This is a deliberate, already-*tested* tradeoff — `tests/unit/lib/matchers/tracesPricing.test.ts`: "falls back to sonnet-class rates for a generic 'claude' id with no tier keyword" explicitly pins this exact fallback. Changing it would contradict that existing test / the original author's intent. The docstring's complaint is about defaulting for *unrelated model families* (GPT/Gemini/etc.), not a same-vendor same-order-of-magnitude default. |
| 4 | Bare (un-namespaced) `input_tokens`/`output_tokens`/`model` key matching has no span-kind/namespace guard — an unrelated span could coincidentally collide. | `lib/matchers/traces.ts` (`PROMPT_TOKEN_KEYS` etc.) | Medium | **Deferred, acknowledged**. Real but narrow risk (only matters if another span in the *same correlated trace* happens to reuse these exact generic names for something unrelated); fixing well requires span-kind/operation-name gating, a bigger design change than this fold should make unilaterally to inherited, already-tested code. |
| 5 | Prism + `dangerouslySetInnerHTML` + custom CSS + default-open highlighting is more dependency/render surface than the truncation bug warrants. | `components/evals3/ContextValueView.tsx` | Medium | **Rejected (kept as-is)**. Verified no realistic XSS bypass (codex agrees: "I do not see a concrete XSS bypass from Prism's JSON highlighter"; `Prism.util.encode()` HTML-escapes leaf token content before `dangerouslySetInnerHTML`, and the non-highlighted plain-text path never uses `dangerouslySetInnerHTML` at all). Bundle cost is small (prism core + json component ≈17KB gzip, ~1.4% of the app's existing 1.25MB gzip bundle; zero transitive sub-dependencies). Default-open matches the reported bug's actual ask ("readable on arrival", not "readable after another click"). |
| 6 | Default-open highlighting on every item could be a perf footgun for very large/many context items. | `components/evals3/ContextValueView.tsx` | Medium | **Deferred, acknowledged**. Real test cases observed in practice have a handful of small context items; virtualizing/lazy-highlighting would undermine the "readable on arrival" fix goal. Flagged as a follow-up if it proves to be a real-world problem. |
| 7 | Two hand-maintained pricing tables (`lib/matchers/tracesPricing.ts` vs `server/services/codingAgents/pricing.ts`) risk drifting. | `lib/matchers/tracesPricing.ts` | Medium | **Deferred** (codex itself categorized this as defer). The two tables are intentionally separate — different fallback *policies* (fail-loud here vs. silent-Sonnet-default there) — but consolidating the underlying rate *data* while keeping the differing policies would be a reasonable future refactor. |

Codex's own summary: *"I do not see a concrete XSS bypass... I also do not see residual #420 / TestCaseDefinition / benchmarkDoctor / repair-links references"* — confirms the exclusion work (no #420 mirror content, no version-links/benchmark-doctor bleed-through) is clean.

**Net effect of finding #1's fix**: `totalCost` is now strictly *more* likely to throw (any unresolved span, not just all-unresolved) rather than ever silently returning a partially-wrong number — consistent with the whole feature's stated fail-loud philosophy. Pushed as commit `0e587c64` (fixes + 3 new pinning/regression tests), on top of the original 5 commits.

---

## Text proof (gates re-run after the hardening fix)

```
$ npm run build:all
✓ tsc && vite build — clean
✓ build:server, build:cli, build:lib — clean

$ CI=1 AH_PORT=4661 AH_DEV_PORT=4660 BENCHMARK_RUN_RECOVERY_DISABLED=1 EVALUATION_RUN_RECOVERY_DISABLED=1 npm run test:unit -- --coverage
Test Suites: 264 passed, 264 total
Tests:       5 skipped, 4816 passed, 4821 total
Statements   : 72.9%  (15118/20737)
Branches     : 62.72% (8232/13124)
Functions    : 69.17% (2273/3286)
Lines        : 73.76% (13941/18899)   <- CI baseline: 70%

$ CI=1 AH_PORT=4661 AH_DEV_PORT=4660 BENCHMARK_RUN_RECOVERY_DISABLED=1 EVALUATION_RUN_RECOVERY_DISABLED=1 npm run test:integration
Test Suites: 75 passed, 75 total
Tests:       767 passed, 767 total

$ CI=1 AH_PORT=4661 AH_DEV_PORT=4660 BENCHMARK_RUN_RECOVERY_DISABLED=1 EVALUATION_RUN_RECOVERY_DISABLED=1 npm run test:e2e -- tests/e2e/test-case-context-pretty.spec.ts
Running 1 test using 1 worker
✓ 1 [chromium] › tests/e2e/test-case-context-pretty.spec.ts:110:3 › Test Case Detail — Context section pretty-printing › JSON context renders pretty-printed and expanded by default; non-JSON stays full text (4.2s)
1 passed (54.2s)

31 vulnerabilities (2, 15 moderate, 14 high) — all pre-existing in unrelated
transitive deps (postcss, react-router, undici, vite, protobufjs); none
introduced by this diff (only new dep is prismjs/@types/prismjs, 0 sub-deps,
verified via `npm ls prismjs`).

$ git log origin/main..HEAD --format='%H' | while read h; do git show -s --format='%B' "$h" | grep -q "Signed-off-by:" && echo OK || echo MISSING; done
OK  OK  OK  OK  OK  OK   (6/6 commits DCO-signed)
```

Per-feature coverage on changed source (post-hardening): `ContextValueView.tsx` 93.75%, `contextFormat.ts` 100%, `lib/matchers/traces.ts` 98%+, `lib/matchers/tracesPricing.ts` 100% lines.

Fold provenance recap: cherry-picked `d5921b1` (cost attribution) from `traces-cost-attrs` and `6d2c384a`+`82f7097d` (context pretty-print + highlight color) from `testcase-context-pretty`; deliberately excluded `57d2253d`/`562fa96e`/`b711aee1` (benchmark-doctor/run-first-union/repair-links rename — verified `createBenchmarkDoctorCommand` doesn't exist anywhere on `main`, so nothing for that rename to collide with; belongs to a separate version-level-test-case-links PR) and `7536bcb0` (a pre-added mirror of #420's own `TestCaseDefinition.tsx`, used only as a `git rerere` anchor on an unrelated integration branch).

TackAdam (approved):

PR #437, opened 2026-08-27
feat: content-addressed benchmark images, benchmark doctor, and quick-mode dedup — shared, comparable benchmarks via CLI

goyamegh:
## Why: making shared, comparable benchmarks a first-class thing

This PR ships the piece the release depends on: **content-addressed benchmark images**, a **`benchmark doctor`** command to migrate/clean up the storage you already have, and two prerequisite fixes that make quick-mode runs and the images/doctor rollout safe.

**The story, end to end:**

1. **Content-addressed images (`3f861e24`)** — every evaluation run is stamped with an `imageDigest`: a SHA-256 over the test-case *contents* (not storage ids) + eval conditions (`evaluatorId`, `judgeModelId`). Two teammates on two machines who run "the same benchmark" now provably converge on **one** `BenchmarkImage` doc — inherent dedup — and any runs sharing a digest ran under **identical conditions**, so they're comparable by construction (only the agent varies). This is the comparability guarantee the release is selling.
2. **`benchmark doctor` (`887b91b1`)** — for everyone who *isn't* starting fresh: dry-run by default, detects content-duplicate benchmarks (merges into a canonical, re-points eval-runs, deletes husks — **never touches runs/reports**) and timestamped debris (`quick-<ts>`, `*-<epoch-ms>` benchmarks with no runs anywhere, >24h old, never sample data). `--migrate-images` converts the survivors into tagged, content-addressed images. This is the migration path from "my messy local benchmarks list" to the shared-image model.
3. **CLI quick mode stops polluting storage (`f85b4948`)** — bare `agent-health benchmark` used to mint a fresh `quick-<timestamp>` Benchmark doc *every run*. Fixed to run ad-hoc (no Benchmark doc), which is both a bugfix and what makes `doctor`'s debris count actually converge to zero over time instead of one step behind the next `quick-` mint. `findBenchmark` also hardens name-keyed reuse (ambiguous name match → null, not a guess).
4. **`d8e79d08`** — classic (non-trace) runs no longer inherit the placeholder's `pending` metricsStatus, so a 3s demo run doesn't trace-poll for 10 minutes and misreport as `errored`. Unrelated bug, but it was blocking clean assertions in the doctor/dedup integration tests, so it's bundled here rather than split.
5. **`1c306c8b`** — e2e pin for benchmarks-page eval-run isolation, CLI docs for `doctor`/images/quick-mode, CHANGELOG.
6. **`a8f35e9d`** — hardens the ad-hoc dedup integration test against parallel-suite races.
[...]

PR #442, opened 2026-08-28
Code-SDK & CLI: .eval.ts support, always-recorded metrics, expect.soft, version-level case links, external-cwd DX

goyamegh:
## Summary

Folds 4 fork branches into one PR covering the code-test SDK's `.eval.ts` support, always-recorded objective metrics, `expect.soft()`, benchmark version-level test-case links, and CLI external-cwd DX guards.

**Folds 4 branches** (fetched from `goyamegh/dashboards-traces`, cherry-picked/rebuilt as a clean series onto `origin/main`):
- `goyamegh/evalts-registry-fix`
- `goyamegh/sdk-always-record`
- `goyamegh/version-level-case-links` (cherry-picked its one clean fix commit `5c4740c2`; the rest of that branch is integration-merge noise not relevant here)
- `goyamegh/cli-dx-guards`

No branch conflicted on production files; the only overlap across all four was `CHANGELOG.md`'s `[Unreleased]` section, consolidated into one entry in the last commit.

## 1. `.eval.ts` / `.eval.mjs` test cases now register into the shared registry

**Bug:** `.eval.ts` / `.eval.mjs` files went through a plain native `import()`, so the bare `@opensearch-project/agent-health` specifier resolved through Node's normal module resolution. When the host process (e.g. the dev CLI running from TS source) resolved that specifier to a *different physical file* than the loader's own copy (e.g. a project-local `node_modules` symlink → the compiled `lib/dist`), the two were separate ES module instances, each with its own registry `Map` — a `.eval.ts` file's `test()` calls landed in a registry the loader never read from, surfacing as `"Module ... has no test cases"` even though `.eval.ts` is documented as supported.

**Fix:** `test()`/`describe()`/hook registration now lives behind `globalThis[Symbol.for('agent-health.test-registry.v1')]` instead of plain module-level variables, so every module instance — dist vs source, CJS vs ESM, symlinked vs installed — reads/writes the same registry. No loader/transform changes; native Node type-stripping already compiles typical `.eval.ts` syntax without `tsx`.

**Tests:** unit (`define.registrySharing.test.ts`, simulates two module instances via `jest.resetModules()`); integration (`evaltsRegistrySharing.integration.test.ts`, spawns the real CLI binary against a `.eval.ts` fixture from a project with its own symlinked dependency — the real-world repro shape).

## 2. SDK: always-recorded objective actuals + `notReached` marker

**Bug:** chai's `expect()` is fail-fast — the first failing assertion throws, so any code textually after it (a later `traces.totalCost` read, a `judge()` call) never executes. `durationMs`/`agentDurationMs` already survived a mid-body throw, but `totalTokens`/`totalCostUsd` did not, so a failing gate silently dropped cost/token figures a measurement harness needs from every run regardless of which gate failed.

**Fix:** `stampObjectiveActuals()` stamps `totalTokens`/`totalCostUsd` onto `report.performanceMetrics` from the runner's own `TracesAccessor` the instant `agent.run()` resolves — independent of whether the test body ever reached a matcher that read them. `appendNotReachedMarker()` appends a distinctly-flagged synthetic `MatcherResult` (`notReached: true`) for the tail of a body that never ran, so `MatcherResultsPanel` shows a clear "not reached" row, excluded from pass/fail counts and the pass-rate denominator.

**Tests:** unit (`sdkAlwaysRecord.test.ts`, `MatcherResultsPanel.test.ts`, `benchmarkRunner.test.ts`); integration (`sdkAlwaysRecord.integration.test.ts`, red→green: a 2-matcher eval file where matcher #1 fails still carries `durationMs`/`totalTokens`/`totalCostUsd` on the persisted report, `undefined` pre-fix).

## 3. SDK: `expect.soft(...)`

`expect.soft(value)` is the identical chai assertion surface (every built-in BDD matcher + the SDK's `haveCalledTool`/`haveStepsOfType`/`haveOutputMatching`/`haveCompletedWithin`/`toPass`) except a failing assertion records the `MatcherResult` and returns instead of throwing — so later matchers/`judge()`/`evaluate()` calls in the same body still run. This is the mechanism that rescues a `judge()` call placed *after* a failing assertion; item 2's always-record fix can only recover runner-owned actuals, not a call the body's own source order never reached. Overall test verdict is unchanged: still the AND of every recorded gate. Closes the last open item on the SDK roadmap (`docs/SDK.md`).

**Tests:** unit (`expect.test.ts` — records-without-throwing, body continues after a soft failure, a hard `expect()` after a soft failure still bails on its own failure, custom matchers work in soft mode, no cross-call flag leakage).

## 4. Benchmark version-level test-case links

**Bug:** A CLI-created benchmark (`agent-health benchmark -f foo.eval.js -n "My Benchmark"`) linked resolved test-case ids into `benchmark.testCaseIds` (top level) at run-creation time, but never into the **current version's own** `testCaseIds` entry — the array the benchmark page's test-case panel actually reads (`getSelectedVersionData` → `getVersionTestCases`). The panel rendered "No test cases in this version" despite correct top-level links (dogfood finding: `bench-1787782179901-c1h0eld64`).

**Fix:** `linkTestCaseIdsToBenchmark()` now unions ids into BOTH the top-level array and the matching `versions[]` entry, in place, no version bump. New `POST /api/storage/benchmarks/:id/link-test-case-ids` exposes the same repair for the CLI. `agent-health benchmark repair-links` (built additively, named to avoid colliding with #437's separate `benchmark doctor` image/dedup command) gained a second check, `computeVersionLinkRepairPlan`, that backfills stale benchmarks created before this fix (dry-run by default, `--apply` calls the new endpoint).

**Tests:** unit (`benchmarkPromotion.test.ts`, `benchmarkDoctor.test.ts`); integration (`evaluationRunBenchmarkLink.integration.test.ts` — a CLI-shaped run against a shell benchmark shows the id at both levels, no bump; new endpoint's 404/validation); e2e (`benchmark-version-link-panel.spec.ts`, red→green against the pre-fix top-level-only shape — the exact coverage gap that let this ship, since prior specs only ever seeded `testCaseIds` directly at creation).

## 5. CLI DX guards (A5+A6) + external-cwd docs

- **A5.1** — `startServer`'s port-in-use error now names `AH_PORT` once all fallback ports (4001–4010) are exhausted, instead of a bare `EADDRINUSE`.
- **A5.2/A5.3** — `loadUserConfig`'s config-loading error messages now guide users to `package.json {"type":"module"}` (module resolution failures) or `TSX_TSCONFIG_PATH` (TypeScript parsing failures).
- **A6** — `GET /api/storage/config/status` surfaces a `warnings[]` field (and the server logs a startup warning) when a config file exists but declares zero agents — catches the "silent config data loss" scenario where an auto-updated config stub blanks the agents list.
- **Docs** — new "Running from Outside the Repo" section in `docs/CLI.md`: troubleshooting checklist + a complete external-invocation example.

**Test-quality note:** this branch's own tests were hardened as part of the fold (separate commit, no production code changes) — one integration test file had been committed with literal `\n` escape sequences instead of real newlines (mechanically repaired, byte-for-byte, in the feature commit) and its assertions rewritten to hit the real running server instead of a `mockConfigStatus` literal; the two `loadUserConfig` error branches gained real-fixture unit tests (a real temp-dir `agent-health.config.mjs` + the real, unmocked `loadConfig()`) since they previously had none reaching the actual function.

## Seams checked while composing

- `version-level-case-links` and `cli-dx-guards` both touch CLI wiring, but on disjoint files (`cli/commands/benchmark.ts` + `cli/utils/benchmarkDoctor.ts` vs. `cli/utils/startServer.ts` + `lib/config/loader.ts` + `server/services/configService.ts`) — no merge conflicts, nothing to de-duplicate.
- `benchmark repair-links` is registered **exactly once** (`cli/commands/benchmark.ts`); the branch's own commit message already named it `repair-links` (not `doctor`) specifically to avoid colliding with PR #437's `benchmark doctor` (image/dedup) command.
- The `refactor(cli): rename benchmark doctor to repair-links` commit that exists on the `traces-cost-attrs` fork branch was deliberately **not** pulled in here — it belongs to a different fold PR that still calls the command `doctor` at the point that rename applies; this branch's `repair-links` command was authored with that name from the start, so no rename is needed or included.

## Test plan

```bash
npm run build:all
npm run test:unit
npm run test:integration
```

New/changed test files (all listed above) pass; the new e2e spec (`tests/e2e/benchmark-version-link-panel.spec.ts`) imports `./fixtures/test-fixtures` per repo convention. Full gate output attached as a PR comment.

## Commits

7 commits, each DCO-signed (`git commit -s`):
1. `fix(sdk): .eval.ts/.eval.mjs test cases now register into the shared registry`
2. `fix: SDK code-tests always record objective actuals + notReached marker on a mid-body throw`
3. `feat: expect.soft() — non-throwing assertions for the code-test SDK`
4. `fix(server,cli): link resolved test case ids into a benchmark's CURRENT version, not just the top level`
5. `fix: CLI DX improvements — port-in-use error, config loading help, zero-agents warning (A5+A6)`
6. `test(cli-dx): real-fixture + real-HTTP coverage for the A5/A6 guards`
7. `docs(changelog): consolidate Unreleased entries for the SDK/CLI fold`


goyamegh:
## codex_review (adversarial review, gpt-5.4)

Ran an adversarial review of the full diff against `origin/main` before requesting approval, per team convention. Findings and what I did with each:

| # | Severity | Finding | Action |
|---|----------|---------|--------|
| 1 | HIGH | `benchmarkDoctor.ts`'s two planners are not version-aware — `computeBenchmarkRepairPlan` unions test-case ids from every linked run into the top level regardless of which benchmark version was current when that run happened, and `computeVersionLinkRepairPlan` then offers to copy that top level into the CURRENT version. For a benchmark with more than one version, `--apply` could silently mix an older version's test cases into the current one. | **Fixed.** Added a `needsManualReview` flag (`versions.length > 1`); `repair-links --apply` now skips flagged plans and prints an explicit manual-review pointer, dry-run still reports them. New unit tests. |
| 2 | HIGH | Linking test-case ids into a benchmark happens *before* the evaluation-run document is created (`server/routes/storage/evaluationRuns.ts`), with no rollback if run creation fails afterward. | **Accepted, documented.** Pre-existing best-effort tradeoff (already explicitly commented in the code as "must not abort the run"); reordering risks destabilizing a well-tested runner path during a fold. Flagged for a follow-up PR. |
| 3 | HIGH | `lib/testCases/define.ts`'s registry-sharing fix globalizes `activeFile`/`describeStack`, not just the registries — theoretical risk if two eval files ever loaded concurrently. | **Investigated, documented.** Verified both real call sites (`services/sourceResolver.ts`, `cli/commands/benchmark.ts`) load files strictly sequentially (`for` + `await`, never `Promise.all`) — the design's safety depends on that invariant, which holds today. Added a code comment stating the invariant explicitly for future maintainers. |
| 4 | MED | `appendNotReachedMarker`'s message asserted "a prior assertion threw" as the cause of any mid-body throw — not something the runner actually knows (could be a bug in test-body helper code, a thrown non-Error, etc.). | **Fixed.** Reworded to describe only what's observed; hedged the `expect.soft(...)` suggestion to the case it actually applies to. |
| 5 | MED | Pre-existing divergence between `evaluationRunner.ts`'s and `benchmarkRunner.ts`'s `agentFailed` computation now feeds into the new always-record behavior in two slightly different ways. | **Accepted, documented.** Confirmed both files' `agentFailed` logic is unmodified by this PR (pre-existing on `origin/main` in both). Flagged as a separate cross-cutting inconsistency worth its own follow-up. |
| 6 | MED | `expect.soft`'s patched `Assertion.prototype.assert` may not survive multi-step chains correctly (e.g. `.to.have.property('x').that.equals(5)`). | **Verified empirically — real.** A missing-property chain in soft mode does record a derivative, misleading second failure (`expected undefined to equal 5`) that's a symptom, not independent signal. Documented as a known limitation in `docs/SDK.md` and a code comment in `lib/matchers/expect.ts`, with a pinning regression test so it can't silently change unnoticed. Deep-fixing chai's internal chain semantics was judged too invasive for this fold. |
| 7 | MED | `lib/config/loader.ts`'s TypeScript-parsing hint matched a bare `message.includes('TS')` — a 2-letter substring that could false-positive on unrelated messages (e.g. anything mentioning "HTTPS"). | **Fixed.** Removed; the `'tsconfig'` check in the same condition already covers the real-world case with zero loss of legitimate coverage. |
| 8 | LOW | `configService.ts`'s zero-agents `console.warn` fired on every `getConfigStatus()` call — and the endpoint is polled by the UI on page loads, not once at startup as the CHANGELOG claimed — so a persistent zero-agents config would spam the server log. | **Fixed.** Added a warn-once guard (mirrors the existing `emitExperimentalWarningOnce` pattern already in this codebase); the returned `warnings[]` field itself stays live on every call. Corrected the CHANGELOG's inaccurate "on startup" claim. Also discovered while fixing this: `getConfigStatus()` **is** reachable from Jest via a relative-import path that misses the `moduleNameMapper` redirect (the redirect's original `import.meta.url` reason no longer applies to this file) — added a real unit test (`configServiceZeroAgents.test.ts`) calling the actual function instead of the re-derived-literal-logic style the rest of the A6 tests were stuck with. |
| 9 | LOW | The `repair-links` CLI command / endpoint / planner / client-method surface looks like a lot of permanent surface for what's essentially a one-off historical backfill. | **Noted, not actioned.** Reasonable design-philosophy disagreement (discoverable end-user CLI command vs. a maintainer-only migration script) rather than a bug — leaving as authored. |

**Bonus catch while re-running the full suite after applying the above:** found one genuine regression against a **pre-existing** `origin/main` test — `tests/integration/cli/benchmarkCodeSdk.integration.test.ts`'s issue #245 regression test counted every `pass:false` matcher as a real failure, and the always-record fix's new `notReached` marker (also `pass:false`) pushed that count from 1 to 2. Fixed by excluding `notReached` entries, mirroring the exclusion `computeSdkMatcherSessionMetrics`'s own gate filter already applies. This is exactly the class of issue a fold is supposed to surface — the "sdk-always-record" branch and this particular `origin/main` test had never run against each other before this PR.

All fixes are in the last commit (`fix: codex_review hardening — ...`), separate from the folded feature commits so the attribution/history stays clean. Full text proof (build/test output) in the next comment.

goyamegh:
## Verification / test proof

All gates run locally against this branch's final commit (`ebf66778`), rebuilt from a fresh `npm install`.

### `npm run build:all` — clean

```
> tsc && vite build            ✓ built in ~20s (no tsc errors)
> esbuild server/...            ✓ server/dist/index.js, server/dist/app.js
> esbuild cli/...                ✓ cli/dist/index.js (434.1kb)
> tsc -p tsconfig.lib.json && tsc-alias ...   ✓ (no errors)
```

Full `npx tsc --noEmit -p tsconfig.json` (whole-project typecheck) also passes with zero errors.

### `npm run test:unit` — 268/268 suites, 4847/4852 tests passed (5 skipped), 0 failed

```
Test Suites: 268 passed, 268 total
Tests:       5 skipped, 4847 passed, 4852 total
Snapshots:   0 total
Time:        39.878 s
```

### `npm run test:integration` — 85/85 suites, 799/799 tests passed, 0 failed (clean run)

```
Test Suites: 85 passed, 85 total
Tests:       799 passed, 799 total
Snapshots:   0 total
Time:        57.716 s
```

Ran against the real, shared OpenSearch storage cluster (`BENCHMARK_RUN_RECOVERY_DISABLED=1 EVALUATION_RUN_RECOVERY_DISABLED=1`, `AH_PORT=4641` — never touched 4000/4001).

**Note on flakiness during this session:** this branch was built and tested on a shared box running many other concurrent orchestrated CI-style jobs (other in-flight branches' build/test/e2e runs). Two full `test:integration` re-runs during heavy concurrent load produced 8–12 failing suites, but every failure traced to one of two unrelated causes, neither a regression in this diff:
1. **Client-side test timeouts** (`AbortError` from a hardcoded 20s `AbortController` in test helpers) on tests whose underlying operation legitimately took longer under system-wide CPU contention — non-deterministic *which* test hit it each time, and the exact files affected (`testCasesBulkUpsert.integration.test.ts`, `benchmarkExecuteEvaluator.integration.test.ts`) are untouched by this PR and already exist on `origin/main`.
2. **A `jest --testPathPatterns='tests/integration'` run briefly reported a `benchmarkDoctorRepair.integration.test.ts` file that doesn't exist in this branch or on `origin/main`** — root-caused to jest's default `cacheDirectory` (`/tmp/jest_<5-char-suffix>`) colliding across several *other* concurrently-running worktrees on the same box that *do* still have that file (from an earlier, superseded commit of the same upstream feature branch). Confirmed by re-running with an explicit unique `--cacheDirectory`: the phantom file disappeared. Not a defect in this PR; flagging in case it's useful for the shared multi-worktree CI setup.

Every test file this PR's diff specifically touches was independently re-verified clean and deterministic in isolation (not just as part of the full suite): `evaluationRunBenchmarkLink.integration.test.ts` (5/5), `evaltsRegistrySharing.integration.test.ts`, `benchmarkCodeSdk.integration.test.ts` (18/18 — see codex_review finding + fix above), `benchmarkDoctor.test.ts` (18/18), `configServiceZeroAgents.test.ts` (6/6), `expect.test.ts` (24/24).

### `npm audit --audit-level=high`

31 pre-existing vulnerabilities (react-router, undici via a devDependency, vite) — **zero new**. `package.json`/`package-lock.json` are unchanged by this PR (`git diff origin/main..HEAD -- package.json package-lock.json` is empty).

### DCO

All 8 commits carry `Signed-off-by: Megha Goyal <goyamegh@amazon.com>`.

### e2e fixture convention

The one new e2e spec, `tests/e2e/benchmark-version-link-panel.spec.ts`, imports `./fixtures/test-fixtures` per repo convention (confirmed by inspection).

goyamegh:
## Verification proof (repair wave 2026-08-31, head b40cf2bf)

**CI**: all required checks green on this head (build 18/20/22, unit, integration-tests, e2e-tests, codecov patch+project, lint, security-scan, release-rehearsal, DCO, changelog, license, link-check). Code-Diff-Analyzer red is the advisory AI bot.

**Local verification** (Node 20.10.0, matching the CI integration-tests pin):
- `npm run build:all` green; `npx tsc --noEmit --skipLibCheck` clean
- unit: 294 suites, 5071 passed / 5 skipped / 0 failed
- integration: 79 suites, 783/783 (2 consecutive clean full runs)
- DCO: 13/13 commits signed off; CHANGELOG updated under Unreleased

**Root-cause fix for the previously failing `integration-tests` job**: `.eval.ts` loading relied on native `import()` TypeScript support (absent on Node 20) — now unified on an esbuild-transpile → synthetic-CJS path for all Node versions (one loader, no ESM cache staleness), with esbuild declared as a runtime dependency.

**Adversarial review (OpenAI Codex, 2 rounds) — all actionable findings addressed**:
- BLOCKER: unguarded read-modify-write in `linkTestCaseIdsToBenchmark` → replaced with an atomic server-side Painless scripted `_update` (union into top-level `testCaseIds` + current version entry, legacy-doc v1 synthesis in-script, `retry_on_conflict`), file backend serialized via write mutex. Validated against a real OpenSearch node incl. a 20-way fully-concurrent link burst with zero id loss.
- HIGH: multi-file `.eval.ts` imports → `wrappedRequire` recursively transpiles relative `.ts` helpers (per-load memoized); ESM-only imports get an actionable error; two-file fixture + real-subprocess tests added.
- HIGH: `.eval.mjs` reload staleness → cache-busting import URL + reload test.
- Rejected (documented): globalThis registry version-skew (v1-keyed symbol, accepted residual); `expect.soft` chain derivative-failure limitation (documented in docs/SDK.md).

ps48 (approved):

PR #443, opened 2026-08-28
Run report v2: verdict-first report, failure insights, trends band, and report-page performance

goyamegh:
## What / Why

This PR folds 4 fork branches (6 logical features) into one series to cut review overhead — they were built and tunnel-verified together as one coherent "run report v2" surface, and land more coherently as one PR than six drive-by ones touching the same files.

### 1. Legacy run-report page: verdict-first redesign (`components/RunDetailsPage.tsx`, `components/RunSummaryBand.tsx`)
The bare `/benchmarks/:benchmarkId/runs/:runId` route required clicking a "Summary" card to see anything besides an empty pane. It now renders an always-on **RunSummaryBand** (agent + model, judge + evaluator, started/duration/concurrency, judged cost, verdict counts) with the full test-case list directly underneath — no redirect, no click-through. Selecting a case `push`es `?testCase=<id>` (was a history `replace`, breaking back/forward); a deep link with `?testCase=<id>` preselects and scrolls to that case. `RunSummaryPanel.tsx` (old click-through donut-chart panel) is removed.

### 2. Failure-insights pane (`components/RunInsightsPane.tsx`, `lib/runInsights.ts`)
With nothing selected, the right pane now shows verdict overview, per-category pass/fail bars, and a **deterministic (no-LLM) "why runs failed" theme clustering** of failing cases' judge reasoning, with click-to-filter by theme, slowest/costliest case shortcuts, and a collapsible Details section.

### 3. Agent Trends band + legend drawer (`lib/agentTrends.ts`, `components/charts/AgentTrendsEChart.tsx`, `components/dashboard/AgentTrendsBand.tsx`, `components/dashboard/AgentTrendsLegendDrawer.tsx`)
Replaces the Recharts Performance Trends card (an always-visible legend unreadable past ~4 agents) with an ECharts multi-agent trend chart (rolling average + per-run scatter, metric toggle, benchmark/time-range scoping). The per-agent legend moves into an on-demand drawer; data points are labeled directly, with each agent's latest point always pinned. Includes two correctness fixes: proper chart unmount, and partially-populated trace-cost sums no longer presented as complete.

### 4. Run-report lazy loading (`components/RunDetailsPage.tsx`)
The legacy run-report page fetched every full report body sequentially and resolved test cases via the entire unscoped test-case corpus (**168MB observed in repro**) just to filter to the run's cases — a large run rendered a silent, permanent blank pane. Now uses one lightweight status batch + one id-scoped test-case batch; the full report body is fetched on-demand only when a case is selected.

### 5. Run-selector popover anchoring (`components/ui/popover.tsx`, `components/comparison/ComparisonSearch.tsx`)
The shared `Popover` primitive used hand-rolled `position: absolute` CSS inside a `position: sticky` toolbar, which could render the run-selector panel floating over the scoreboard instead of anchored beneath its trigger. Rebuilt on Radix's `@radix-ui/react-popover` (portaled, Floating UI-positioned) — same public API, no call-site changes elsewhere.

### 6. Judge and Evaluator columns (`components/evals3/EvalRunsPage.tsx`, `lib/utils.ts`, `services/storage/asyncBenchmarkStorage.ts`)
The Evaluation Runs table now shows which judge model and which evaluator graded each run. Also fixes a pre-existing bug where `asyncBenchmarkStorage`'s field whitelists silently dropped `judgeModelId`/`evaluatorId` for benchmark-embedded runs, so the columns would have shown blank for most historical runs.

## Test plan

- `npm run test:unit` — 266 suites / 4866 tests passed.
- `npm run test:integration` — 75 suites / 767 tests passed (against a real OpenSearch cluster).
- e2e: all 6 new/changed specs (`run-report-redesign`, `run-report-insights`, `run-report-lazy-load`, `agent-trends-band`, `comparison-search-popover-anchor`, `evaluation-runs-judge-evaluator-columns` — 15 tests total) pass; each imports `test`/`expect` from `./fixtures/test-fixtures` for coverage collection.
- `npm audit --audit-level=high` — 14 high / 15 moderate / 2 low / 0 critical, identical to the `main` baseline measured on the same commit this branch is based on (delta: zero net-new vulnerabilities from the one added dependency, `@radix-ui/react-popover`).
- Unit coverage on the files this PR changes: 82.6% statements / 74.5% branches / 80.1% functions / 83.3% lines (several are 90–100%; the lower-covered files are page-level components exercised by the e2e specs above rather than by `.test.ts` unit tests, per this repo's convention).
- All 6 commits carry DCO signoff; `CHANGELOG.md` has one consolidated `## [Unreleased]` entry set covering exactly these six features.

## Breaking changes

None. Same routes, same public component APIs (`Popover`/`PopoverTrigger`/`PopoverContent` unchanged), no config/schema migrations.

## Why one PR

This folds what were 4 separate fork branches (`tunnel-ui-stack` — itself `run-report-redesign` + `run-report-insights` + `trends-legend-drawer` — plus `run-report-lazy-load`, `fix-run-selector-anchor`, `evalruns-judge-columns`) into one clean, dependency-ordered commit series (6 commits) with a single consolidated CHANGELOG entry, to cut review overhead versus raising six overlapping PRs that all touch `RunDetailsPage.tsx`/`Dashboard.tsx`/the runs table.


goyamegh:
## codex_review (adversarial second opinion — GPT-5.4)

Ran two focused `codex_review` passes over the full diff (batch A: run-report redesign + failure-insights pane + lazy-load; batch B: Agent Trends band + popover anchoring fix + Judge/Evaluator columns), asked to hunt both correctness risks and over-engineering. Findings, and what I did with each:

### Batch A — run-report page (redesign / insights / lazy-load)

| # | Sev | Finding | Applied / Rejected |
|---|-----|---------|---------------------|
| 1 | HIGH | `loadRunData` depends on `searchParams`, so selecting a case (which updates `?testCase=`) re-triggers the full run bootstrap fetch, partially undermining the lazy-load fix. | **Rejected (mischaracterized).** Verified this `searchParams` dependency + `setSearchParams` on selection is byte-identical to `origin/main`'s pre-existing code (`git show origin/main:components/RunDetailsPage.tsx` — same deps array, same call pattern). Pre-existing behavior, not introduced by this PR. Worth a separate follow-up issue, not a blocker here. |
| 2 | HIGH | The "deterministic" failure-clustering claim is overstated: it clusters the first 100 `Object.entries()`-ordered failures, so ordering changes could change the sample. | **Acknowledged, deferred.** Object key insertion order is stable in practice (derived from the benchmark's stored `testCaseIds` order), so this is a low-probability edge case, not an observed bug. Filed as a robustness follow-up. |
| 3 | HIGH | Short (<3 words) or empty judge-reasoning strings become singleton "themes" instead of clustering. | **Rejected (by design).** The 3-word minimum is an explicit, commented tradeoff in `lib/runInsights.ts` (`MIN_WORDS_FOR_SHINGLING`) — clustering very short reasonings risked merging unrelated failures. Accepted limitation for v1. |
| 4 | HIGH | Union-find transitive closure can bridge genuinely distinct failures if they share different phrases via a chain (A~B~C). | **Acknowledged, deferred.** Real property of the chosen heuristic; validated against a real 64-case production run (57/64 correctly merged, 7/64 correctly separate) per the source comment. A precision/recall tradeoff for a v1, LLM-free heuristic — noted for follow-up, not a hard bug. |
| 5 | HIGH | The "N of M failing cases" capped note's fallback expression (`\|\| failingReportIds.length`) can misstate counts. | **Acknowledged (minor).** The current JSX only renders the note when `themes.length > 0`, which prevents the worst case codex described from actually reaching the UI, but the `\|\|` fallback is genuinely sloppy defensive code. Filed as a cheap follow-up cleanup. |
| 6 | MED | `fetchBatchMetrics` is called 3x (twice in `RunInsightsPane`, once in `RunDetailsPage`) for overlapping `runIds`. | **Acknowledged, deferred.** Real redundancy, not a correctness bug — filed as a follow-up consolidation. |
| 7 | MED | The redesign drops the old per-row `RunScore` (numeric score) from the test-case list; only status + duration remain per row. | **Acknowledged.** Intentional design tradeoff from the original redesign (avg score still available in the Details fold-out) — accepted as shipped, flagging for product follow-up if reviewers want it back. |
| 8 | MED | `RunInsightsPane`'s "why runs failed" only clusters `status === 'failed'`, excluding `errored`. | **Rejected.** `errored` means the evaluator/judge itself didn't run (no `llmJudgeReasoning` produced), so including it would only add empty-reasoning singleton noise, not signal. Errored counts are still visible in the Overview verdict chips above. |
| 9, 10 | MED/LOW | Manually-joined `useEffect` deps (`.join(',')`) and in-place `searchParams` mutation are brittle patterns. | **Acknowledged, deferred.** Both match pre-existing conventions elsewhere in this codebase; real but not novel-to-this-PR risk. |

### Batch B — Agent Trends band / popover / Judge+Evaluator columns

| # | Sev | Finding | Applied / Rejected |
|---|-----|---------|---------------------|
| 1, 2 | HIGH/MED | `rollingAverage()` carries a prior non-null value forward across `null` (incomplete-run) samples, so the **rolling-average line** can plot a smoothed value at a timestamp whose actual cost/tokens was `null`/unknown. | **Verified, acknowledged, deferred.** Confirmed by hand-tracing `rollingAverage([10, null, 20])` → `[10, 10, 15]`. Importantly, this only affects the smoothed **line**; the **scatter dots** (the actual per-run values, which is what the "no fabricated/understated cost" fix targets) correctly stay excluded for incomplete runs via the separate `valued` filter in `buildAgentTrendSeries`. So the PR's stated correctness goal (never show a fabricated/partial per-run total) is intact; the smoothing artifact on the trend line is a real but secondary issue, filed as a follow-up. |
| 3 | MED | `AgentTrendsEChart`'s imperative echarts `click` handler effect omits `onSelectRun` from its dependency array (stale-closure risk). | **Acknowledged, deferred.** Confirmed the omission; today's only caller (`goToRun`) has no captured state that changes between renders, so it's not a live bug, but it is a footgun for future callers. Filed as a follow-up (ref pattern instead of effect-scoped closure). |
| 4 | MED | The popover rewrite (hand-rolled → `@radix-ui/react-popover`) is a bigger behavioral surface (portal, collision detection, focus handling) than "same API, anchoring fix" suggests. | **Acknowledged (framing), not reverted.** Fair critique of how I described it. Mitigated by: it's a well-established, heavily-used Radix primitive (already a transitive dependency via Select/DropdownMenu), zero new `npm audit` vulnerabilities, and e2e regression coverage across 4 specs (the new anchor spec + 3 pre-existing specs with nested Selects) all pass. |
| 5 | MED | `(stored as any).judgeModelId`/`.evaluatorId` casts in `asyncBenchmarkStorage.ts#toBenchmarkRun` are redundant — `StorageBenchmarkRunConfig` already declares both fields as optional. | **Applied.** Removed both casts (now plain `stored.judgeModelId`/`stored.evaluatorId`), fixed the interface comment that incorrectly justified the cast. Verified: `tsc --noEmit` clean, `asyncBenchmarkStorage.test.ts` (56 tests, including the exact regression test for this mapper) still green. Folded into the "judge columns" commit. |

**Not applied as code changes in this PR:** everything above except #5 in batch B — all are either mischaracterized-as-regression (verified pre-existing), accepted v1 design tradeoffs with explicit rationale in source comments, or real-but-secondary follow-ups that don't change this PR's correctness for the scenarios its own tests cover. Happy to open follow-up issues for the deferred items if maintainers want them tracked individually.

goyamegh:
## Verification proof

**Build** — `npm run build:all` (tsc + vite build + esbuild server/cli + tsc lib): green.

**Unit tests** — `npm run test:unit`:
```
Test Suites: 266 passed, 266 total
Tests:       5 skipped, 4866 passed, 4871 total
Time:        41-47s
```

**Integration tests** — `npm run test:integration` (against a real OpenSearch cluster):
```
Test Suites: 75 passed, 75 total
Tests:       767 passed, 767 total
Time:        ~57s
```

**e2e** — all 15 tests across the 6 new/changed specs (`run-report-redesign.spec.ts`, `run-report-insights.spec.ts`, `run-report-lazy-load.spec.ts`, `agent-trends-band.spec.ts`, `comparison-search-popover-anchor.spec.ts`, `evaluation-runs-judge-evaluator-columns.spec.ts`) pass. Each spec imports `test`/`expect` from `./fixtures/test-fixtures` (verified via `grep`) so Istanbul coverage is collected in CI. Sample run (workers=1, no internal contention):
```
✓ run-report-lazy-load.spec.ts › renders case list promptly, no unscoped fetch (7.8s)
✓ run-report-lazy-load.spec.ts › explicit loading indicator (2.8s)
✓ run-report-lazy-load.spec.ts › inline error + Retry (2.0s)
✓ run-report-redesign.spec.ts › bare URL renders summary band + case list directly (7.9s)
✓ run-report-redesign.spec.ts › click updates URL to ?testCase=<id> (2.3s)
✓ run-report-redesign.spec.ts › deep-link preselects and scrolls (1.9s)
6 passed (10.1m)
---
✓ agent-trends-band.spec.ts × 4 (mobile drawer, chart render + metric toggle, empty-state placeholder, checkbox hides series)
✓ comparison-search-popover-anchor.spec.ts › anchors beneath trigger, click-outside closes (1 flaky→pass on retry: sub-pixel gap assertion, 12.35px vs a 12px threshold — timing/rendering artifact, not a logic bug)
✓ evaluation-runs-judge-evaluator-columns.spec.ts › renders Judge/Evaluator columns with fallback
✓ run-report-insights.spec.ts × 3 (split view + dominant theme, theme click filters list, Overview clears selection+filter)
8 passed, 1 flaky (3.8m)
```

**Security** — `npm audit --audit-level=high`: `14 high / 15 moderate / 2 low / 0 critical` — **identical** to the `origin/main` baseline (verified by running the same audit against a clean `origin/main` checkout at the same base commit). The one new dependency this PR adds (`@radix-ui/react-popover`) introduces zero net-new vulnerabilities of any severity.

**Coverage on changed files** (unit-only, local approximation — CI's Codecov patch-coverage additionally merges integration + e2e Istanbul coverage, which isn't reproducible outside CI):
```
Statements: 82.61% (1117/1352)   Branches: 74.53% (682/915)
Functions:  80.06% (241/301)     Lines:    83.33% (950/1140)
```
Several files (new, self-contained logic) are 90-100% unit-covered (`RunSummaryBand.tsx` 100%, `lib/runInsights.ts` 97%, `lib/agentTrends.ts` 97%, `services/storage/asyncRunStorage.ts` 91%). The page-level container components (`Dashboard.tsx`, `EvalRunsPage.tsx`, `components/ui/popover.tsx`, `AgentTrendsBand.tsx`) have little/no dedicated `.test.ts` coverage but are exercised end-to-end by the e2e specs above per this repo's convention (unit tests target logic-heavy `lib/*` + isolated components; page-level integration is asserted via e2e).

**DCO** — all 6 commits carry `Signed-off-by: Megha Goyal <goyamegh@amazon.com>`.

Ran under isolated ports (`AH_PORT=4631`/`AH_DEV_PORT=4630`) with recovery sweeps disabled; e2e test data (test cases/benchmarks) was created against and cleaned up from the shared OpenSearch storage cluster after verification (confirmed 0 leftover `e2e-*`-named documents post-cleanup).

goyamegh:
Trends band upgraded in `d189d62f`: replaced the multi-agent time-series (colliding on-point labels, misleading interpolation across sparse runs) with a **ranked dot plot of the last-active benchmark** per owner feedback — one row per agent ranked by latest score, large dot = latest run, faded dots = history on the same benchmark, numbers on hover only, click-through to the run report. Data-layer fixes included: zero-evaluable runs plot as null (no fake 0.0%), deterministic per-agent palette, gap-broken sparklines in the History drawer. 11 unit specs + typecheck + build green.

goyamegh:
Added a fix folded into this PR at 46cece78 for the "benchmark runs list shows 0✓ 0✗ N⚠️ on every row" issue reported against a Feb 2026 pulsar benchmark (`exp-1765401828206-yq9ychdhu`, v8).

**Diagnosis (live server, read-only):** fetched several of those runs' reports directly — the judge verdicts DO exist (`passFailStatus: 'passed'/'failed'`, `metricsStatus: 'ready'`) and match the run's own denormalized `run.stats` exactly (verified: 4 reports `passed` + 3 `failed` == `run.stats = {passed:4, failed:3}`). The bug is in `computeRunStats`/`bucketRunResults` (`lib/runStats.ts`): it reads verdicts only from `run.results[testCaseId].passFailStatus`, a field that `evaluationRunner.ts` only started denormalizing onto `results` after these runs were created. Legacy `results[testCaseId]` entries only ever carried `{ reportId, status: 'completed' }`, so every case bucketed as `errored` even though real verdicts exist.

**Fix:** `computeRunStats` now detects that specific legacy signature (every result bucketed `errored`, 0 passed/0 failed) and falls back to the run's own `run.stats` when it independently carries real (non-zero) passed/failed evidence, instead of asserting the whole run errored. Genuinely all-errored runs, and runs with a real mix of verdicts + judge errors, are unaffected (covered by dedicated negative tests).

**Tests:** `tests/unit/lib/runStats.test.ts` — new case reproduces the exact legacy doc shape (verified red without the fix, green with it), plus two negative cases guarding against masking real errors.

Gates run locally on this head: `npm run build:all` ✅, `npm run test:unit` (268 suites / 4901 passed) ✅, targeted integration (`benchmark-stats-refresh`, `storage/routes/benchmarks`) ✅. DCO signoff present.

ps48 (approved):

PR #431, opened 2026-08-26
feat(test-cases): render code-SDK eval source as an IDE-style view

goyamegh:
## What

The Test Case detail page renders a code-SDK (`benchmark -f *.eval.js`/`.eval.ts`)
test case's **full source file** as a read-only, IDE-style code view — Prism.js
syntax highlighting, non-scrolling line-number gutter, a sticky filename header
(language badge, line count, copy-to-clipboard), collapsed by default (chevron
toggle; highlighting runs lazily on first expand). It replaces the old
"Source File" path row + bare sha256 line (the new header shows the full
relative path, with the sha256 in its tooltip). Pre-existing rows imported
before this feature shows a graceful "source not captured at import"
placeholder instead of a blank/broken view.

## Why

A code-SDK test case (`sourceFile` set) previously only surfaced its file
*path* on the detail page — the `.eval.ts`/`.eval.js` body that actually
defines the test (matchers, hooks, `evaluate()`) was invisible from the UI.
Anyone auditing/debugging a benchmark had to go find the file on disk.

## Persistence design decision

The loader (`lib/testCases/loader.ts`) now captures the eval file's raw text
**once at import time** (`LoadResult.fileSource`, read before either the
CJS-wrapper or ESM dynamic-import execution path) and both import producers —
the CLI's `benchmark -f` file mode and the server-side `resolveCodeImport`
code-import job type — persist it on the `TestCase` as `sourceCode` +
`sourceFileName` + `sourceLanguage` (extension-based `detectSourceLanguage()`
in `lib/utils.ts`, isomorphic).

`computeTestCaseHash()` now folds the **whole file's text** into the hash
(previously only the options fields it tracked). Editing only the
`evaluate()` body, a helper, or even a comment — with no options-field
change — now still invalidates the version and refreshes the stored
`sourceCode` on the next import. (One-time consequence: every pre-existing
code-SDK row gets a hash change, and therefore a `sourceCode` backfill, on
its next `benchmark -f` run.)

**Summary-vs-full, lazy fetch:** `sourceCode` can be many KB, and list
endpoints (`fields=summary`) already strip `context`/`expectedOutcomes` for
the same reason — `sourceCode` is stripped there too. The two pages that
bulk-load *all* of a run's test cases just to paint a list —
`RunInspectorPage.tsx` and `BenchmarkRunDetailPage.tsx` — now request that
same summary shape (`getByIds(ids, { summary: true })`) and lazily fetch the
[...]

PR #444, opened 2026-08-28
Server & storage hardening: evals index field-limit protection, fresh SPA HTML, scoped test-case fetches, category matrix fix

goyamegh:
# Server & storage hardening: evals index field-limit protection, fresh SPA HTML, scoped test-case fetches, category matrix fix

Folds 4 independently-developed fixes into one series (4 commits, one per
feature) so they land together as a themed "server & storage hardening"
pass. Each feature already had its own worktree/tests; this PR is the
integration point.

## 1. Storage: `evals_runs` field-count protection (the incident this PR is really about)

**What happened:** on a real cluster, benchmark runs against code-QA test
cases (many distinct custom evaluator metrics) started **executing
successfully but failing to persist**. The judge would run, produce a
verdict, and then the storage write would error with:

```
illegal_argument_exception: Limit of total fields [5000] has been exceeded
```

**Root cause:** `evals_runs`' report-level `metrics` object and the SDK's
per-matcher `matcherResults[].judgeMetrics` object are both open-ended
`Record<string, number>` shapes, driven by whatever `evaluator.scoringConfig.metrics`
names a custom evaluator uses. Every distinct metric/dimension name ever
written minted a **new mapped field, shared across the whole index** —
OpenSearch's default dynamic mapping has no way to know these are "just
data." A benchmark exercising enough distinct evaluators eventually pushes
the index past its total-field-count ceiling. The run's *execution*
succeeded; only the *write* failed — so failures showed up as silent data
loss (a completed run with nothing persisted), not a crash.

This mirrors [#418](https://github.com/opensearch-project/agent-health/pull/418)'s
fix for `evals_experiments`, which did not cover this index.

**Fix** (`server/constants/indexMappings.ts`): both fields now declare
`dynamic: false`, with the 4 legacy metric names kept explicit/typed/queryable.
Every other custom metric name still round-trips via `_source` — it's just
not separately mapped, so it can never blow the field-count budget. Query-audited:
no OpenSearch-side query anywhere filters/sorts/aggregates on a non-legacy
`metrics.*`/`judgeMetrics.*` name, so nothing loses query capability it
actually used.

**Migration note:** new/clean indexes self-heal on next boot (`ensureIndexes()`).
An **already-poisoned** index (i.e. one that already hit the ceiling before
this ships) needs one explicit `POST /api/storage/reindex {"index":"evals_runs"}`
call — an existing endpoint, not new. Full recipe, caveats, and a
zero-data-loss guarantee walkthrough: [docs/STORAGE_INDEX_FIELD_LIMITS.md](docs/STORAGE_INDEX_FIELD_LIMITS.md).
**This PR does not run that reindex against any live cluster** — it's an
operator action, deliberately not automatic (a migration triggered silently
on every boot is its own hazard).

**Cross-server hazard (flagged during review, not fixed here — pre-existing,
out of scope for this PR):** the manual `/api/storage/reindex` route calls
`reindexSingleIndex()` directly, and the migration lock it (and the boot-time
auto-fix path) uses is process-local — it does not coordinate across
multiple agent-health *server processes* pointed at the same shared cluster
(a supported, documented deployment pattern in this repo).
`reindexSingleIndex()`'s algorithm deletes the original index for the
duration of the copy (temp → original), so a **second server instance**
actively writing to `evals_runs` during that window sees `index_not_found`
errors, not just contention on the one server issuing the reindex. Read the
existing docs' "run it during a quiet window (no in-flight evaluation runs
writing reports)" as *no server sharing this cluster* should have in-flight
runs at that time, not just the one you're SSH'd into.

Tests: [tests/unit/server/constants/indexMappings.test.ts](tests/unit/server/constants/indexMappings.test.ts);
integration (real OpenSearch, skips gracefully without one):
[tests/integration/services/storage/testCaseRunMetricsMappingGrowth.integration.test.ts](tests/integration/services/storage/testCaseRunMetricsMappingGrowth.integration.test.ts)
(1000+ distinct custom metric names persist with zero mapping growth),
[tests/integration/services/storage/evalsRunsMappingMigrationRecipe.integration.test.ts](tests/integration/services/storage/evalsRunsMappingMigrationRecipe.integration.test.ts)
(the existing `reindexSingleIndex()` heals a deliberately-poisoned index with
zero data loss).

## 2. SPA fallback: fresh `index.html` after an in-place rebuild, no restart

`makeSpaFallbackMiddleware` used to be handed a pre-read `index.html`
**string**, captured once at process boot. Rebuilding (`npm run build`) in
the same long-lived server process rehashes `dist/assets/*`, but the cached
HTML kept referencing the old, now-deleted asset filenames — every
client-side route served a blank page (browser refuses the stale JS/CSS
with a strict-MIME 404) until the process restarted.

Fix (`server/middleware/spaFallback.ts`, `server/middleware/index.ts`): the
middleware now takes the on-disk **path**, not the contents, and re-reads it
whenever its mtime changes (one `fs.statSync` per request — negligible at
this traffic). Falls back to the last-good cached copy if a stat/read fails
mid-rebuild (never a 500 or a blank response); a 503 is returned only if the
file has never successfully been read at all.

Tests: [tests/unit/server/middleware/spaFallback.test.ts](tests/unit/server/middleware/spaFallback.test.ts)
(mtime-change serves new HTML, unchanged mtime skips re-read, read failure
serves last-good, never-read returns 503); integration:
[tests/integration/server/middleware/spaFallbackRebuild.integration.test.ts](tests/integration/server/middleware/spaFallbackRebuild.integration.test.ts)
(boots the real compiled server, rewrites `dist/index.html` + adds a new
`dist/assets/*` file while it's running, asserts the very next request
serves the new asset name with no restart).

## 3. Frontend: stop fetching the full test-case corpus into list views

Nine call sites called the bare `asyncTestCaseStorage.getAll()` — every
version, full `initialPrompt`/`context`/`expectedOutcomes`, and (for
code-imported test cases) the entire eval-file source text — just to render
list-view metadata. Measured against a real cluster (3,941 test cases):
**168,125,258 bytes → 1,791,935 bytes** for the equivalent request.

- **Count-only** (`Dashboard.tsx`): now `getAll({ summary: true, size: 1 })`,
  reads `.total` instead of `.length` off a full list.
- **List-metadata-only** (`BenchmarkEditor`/`BenchmarkResultsView`/`BenchmarksPage`/
  `EvalsPage`/`ReportsPage`/`TestCasesPage`/`UseCaseCompareView`): switched to
  `getAll({ summary: true })`.
- **Duplicated import flow** (`BenchmarksPage`/`TestCasesPage` + the evals3
  duplicates): the JSON-import handler called `bulkCreate()` and then
  **re-fetched the entire corpus** to resolve new ids by matching on `name` —
  both a full-payload bug and a correctness bug (two test cases sharing a
  name resolve to the wrong id). `bulkCreate()` now surfaces the created
  `testCases` directly in its return value; every import handler takes ids
  from it, no re-fetch. The evals3 duplicates were reading a non-existent
  `.ids` property (always `undefined` — import always failed with "created
  IDs are unavailable"); they now work. New e2e regression test drives this
  end-to-end via a real file upload on both evals3 pages.
- **Root cause of most of the corpus size** (`server/routes/storage/testCases.ts`):
  `toSummary()`'s `{ ...doc }` spread passed every field through untouched —
  ~165MB of the ~168MB corpus was `sourceCode` (full eval-file text on
  code-imported test cases). `toSummary()` now also nulls `sourceCode`.
- **Data-loss guard** (`TestCaseEditor.tsx`): once list views hand it a
  summary record (truncated prompt, empty context/expectedOutcomes), opening
  Edit → Save would persist that truncated data over the real content.
  `TestCaseEditor` now always refetches the full record by id before
  rendering the form, disabling Save until that resolves — **and** (a
  HIGH-severity codex_review finding, fixed before requesting approval)
  keeps Save disabled and shows a retry-able error banner if that refetch
  fails outright or resolves `null`, instead of silently falling back to
  the (possibly-truncated) prop data. The original version cleared the
  loading flag in a blanket `.finally()` regardless of success/failure,
  which would have re-enabled Save on exactly the failure path this fix
  exists to guard against — a transient network blip while opening the
  editor could have reintroduced the same data-loss bug. New unit tests
  cover both failure modes (rejected fetch, null response) and the retry
  recovery path.

Out of scope (flagged, not touched): `EvalsPage`'s N+1 per-test-case
run-count query; the CLI export path's unpaginated `getAll()`, which
legitimately needs full bodies.

Tests: 9 new/updated unit suites covering every touched component + the
storage/route layers; integration coverage for the summary+ids route
contract and the bulk-response id surfacing; e2e:
[tests/e2e/test-cases-summary-fetch.spec.ts](tests/e2e/test-cases-summary-fetch.spec.ts)
(network-level `fields=summary` assertions + the edit-without-data-loss
regression), [tests/e2e/evals3-import-json-flow.spec.ts](tests/e2e/evals3-import-json-flow.spec.ts)
(real file-upload import on both evals3 pages, asserting the created
benchmark's `testCaseIds` match the bulk-create response — the exact bug
class the "IDs unavailable" regression was).

## 4. Comparison page: category matrix renders subcategory rows on real data

The category × run matrix stayed empty on a real (non-trivial) storage
backend even when the underlying data qualified.

- **Root cause:** the matrix's category-name lookup called the unpaginated
  `asyncTestCaseStorage.getAll()` on every page load — every test case in
  the entire backend, full body included, regardless of how many the
  comparison actually needed. A failure or slow response was swallowed by a
  bare `console.error` with no retry, leaving the lookup permanently empty
  (every row → `(uncategorized)`). The lookup now fetches only the ids the
  selected runs actually reference (`getByIds`, chunked and
  request-id-guarded, same policy as the sibling `getReportsByIds` loader —
  the shared chunking primitive is extracted into `lib/chunkedFetch.ts`).
  **Scope of what this actually bounds** (per a codex_review finding): the
  chunking caps the *browser's* request pattern — never one unbounded
  `?ids=<all>` URL, never more than 8 concurrent HTTP requests of ≤100 ids
  each. It does not change the server route's own fan-out for the `ids`
  filter path (`server/routes/storage/testCases.ts`), which still issues
  one `Promise.all` per request (up to 100 concurrent `getById` calls per
  chunk) — that's pre-existing and deliberate (a comment in that file notes
  it replaced a slower sequential loop) and out of scope here; this PR
  changes how many ids the client ever asks for and how those requests are
  shaped, not the server route's internal fan-out.
- `extractRowCategory()` previously only read a `[bracket]` tag scraped from
  the test-case name, or the generic `topic:` label. It now checks the
  purpose-built `subcategory` field (via its `subcategory:<x>` label) FIRST,
  falling back to `[bracket]` (kept for pre-`subcategory` benchmarks) and
  then `topic:`.
- A regression test pins that `name`/`labels` survive the `fields=summary`
  projection the id-scoped lookup depends on — a lightweight projection
  silently stripping a field a consumer needs is a failure class that has
  bitten other summary endpoints in this codebase before.

Tests: unit coverage for `chunkedFetch`, `comparisonInsights` priority
ordering, and the storage layer's `getByIds`/summary passthrough;
integration coverage for the `ids+fields=summary` route contract; e2e:
[tests/e2e/comparison-category-matrix-scoped-fetch.spec.ts](tests/e2e/comparison-category-matrix-scoped-fetch.spec.ts),
[tests/e2e/comparison-insights-band.spec.ts](tests/e2e/comparison-insights-band.spec.ts).

**Scope note:** the source branch for this fix had accumulated an unrelated,
still-in-flight feature (an eval-source code viewer, tracked separately as
[#431](https://github.com/opensearch-project/agent-health/pull/431)) via a
merge. Only the 4 commits that are actually the category-matrix fix are
folded into this PR; the eval-source-code-viewer commits are excluded
entirely (confirmed zero file overlap with that feature's own files —
`EvalSourceCodeView.tsx` etc. do not exist in this diff). A small amount of
client-side plumbing that fix #4's own commits had implicitly relied on (an
options parameter on `asyncTestCaseStorage.getByIds`/`opensearchClient`'s
`testCaseStorage.getByIds` for requesting the summary projection) originated
from that excluded merge and is re-implemented fresh here, matching the
existing `getAll({ summary })` pattern and covered by
`tests/unit/services/storage/asyncTestCaseStorage.test.ts`.

## Test plan

- `npm run test:unit` — **4824/4824 passed**, 271/271 suites (5 skipped,
  pre-existing env-gated cases). Non-component (server/lib/services) files
  touched by this PR: 86–100% line coverage each (`indexMappings.ts` 100%,
  `spaFallback.ts` 96.6%, `testCases.ts` route 94.6%, `asyncRunStorage.ts`
  97.5%, `asyncTestCaseStorage.ts` 100%, `opensearchClient.ts` 86.5%,
  `chunkedFetch.ts`/`comparisonInsights.ts` 100%). React components are
  intentionally excluded from this repo's unit-coverage collection
  (`jest.config.cjs`'s `collectCoverageFrom` — component behavior is tracked
  via Playwright e2e coverage instead), but every touched component has a
  dedicated new/updated RTL unit suite that renders it and asserts behavior.
- `npm run test:integration` — **775/775 passed**, 78/78 suites, run against
  a real (non-ephemeral) OpenSearch cluster with `--maxWorkers=2` (a fully
  parallel run against a heavily-shared, multi-tenant cluster showed
  transient timeouts unrelated to any file this PR touches — the CI job
  runs each suite against a dedicated ephemeral OpenSearch instance and
  won't see that contention).
- e2e (Playwright, real backend, `fields=summary` network-level assertions
  + a live file-upload import flow): all new/modified specs pass in
  isolation — `evals3-import-json-flow.spec.ts` (2/2),
  `comparison-category-matrix-scoped-fetch.spec.ts` (1/1),
  `test-cases-summary-fetch.spec.ts` (3/4 consistently; see note below),
  `comparison-insights-band.spec.ts` (0/2; pre-existing, see note below).
  Every new/modified e2e spec imports `./fixtures/test-fixtures` (not raw
  `@playwright/test`) so e2e coverage is collected correctly.
- DCO: all 4 commits signed off.
- `CHANGELOG.md` updated under `## [Unreleased]` (one entry per feature,
  consolidated cleanly with no duplicate/overlapping content).

**Two pre-existing e2e flakes against a real, heavily-used shared cluster
(not caused by this PR, not reproducible against CI's ephemeral cluster):**
1. `comparison-insights-band.spec.ts` seeds `expectedOutcomes` as an object
   (`{ conclusions: ['x'] }`). Against a cluster old enough to have already
   locked in `expectedOutcomes` as OpenSearch's dynamic `text` type from
   ordinary plain-string usage, that object write is a structural mapping
   conflict (`mapper_parsing_exception`) — confirmed via a direct mapping
   query, independent of any code path this PR touches. Against a fresh
   (CI) cluster the field is unmapped and the spec's own write establishes
   the type, so this only surfaces against an aged shared cluster.
2. `test-cases-summary-fetch.spec.ts`'s "editing a test case…" case expects
   its freshly-created record to sort first within a 10s wait; against the
   real shared cluster's current size (~1,500+ test cases) a plain
   `fields=summary` list fetch measured **11.6s** end to end — above the
   test's hardcoded timeout. Neither is a functional regression in the
   folded code; both are pre-existing test-design assumptions that only
   break under a large/contended shared cluster, which CI's ephemeral,
   single-tenant, near-empty cluster does not reproduce.

## Codex review

Ran an adversarial second-opinion review (`codex_review`, GPT-5.4, a
different model family) against this diff + PR body before requesting
approval. Findings and disposition:

|---|---|---|---|
| 1 | High | `TestCaseEditor`'s full-record refetch cleared its loading flag in a blanket `.finally()` regardless of success/failure, re-enabling Save with stale/truncated data on a network error or a `null` (404) response — reintroducing the exact data-loss bug the refetch exists to close. | **Fixed** in this PR (see Feature 3 section above) — Save now stays disabled with a visible, retry-able error on either failure mode. New unit tests added (rejected fetch, null response, retry-recovers). |
| 2 | Medium | `bulkCreate()`'s client-side return type says `errors: boolean`; the actual server/storage-adapter contract is `errors: number` (`server/adapters/types.ts` et al) — a pre-existing type-contract lie, not introduced by this PR. | **Not fixed here** — pre-existing on `origin/main` before any of the 4 folded branches, spans benchmarks/runs/test-cases client types (out of scope for this PR), and confirmed unused/dead by every current caller (nothing reads `.errors` off a `bulkCreate()` result today) — no runtime impact, but worth a follow-up type-hygiene PR. |
| 3 | Medium | `lib/chunkedFetch.ts` bounds the *client's* request shape, not server-side OpenSearch call concurrency — the `ids` route still fires one `Promise.all` per request. | PR body's Feature 4 section **clarified** to state this precisely; the server-side behavior is pre-existing/deliberate and out of scope. |
| 4 | Medium | The SPA-fallback rebuild-detection keys freshness on `fs.Stats.mtimeMs` alone; a filesystem with coarse mtime resolution could theoretically miss a same-instant rebuild. | **Accepted as a known, low-probability edge case** — real `npm run build` runs take multiple seconds, and changing this would trade away the whole feature's point (avoiding a read on every request). Not touched. |
| 5 | Medium | The manual `/api/storage/reindex` operator recipe (inherited docs, not authored for this PR) doesn't explicitly call out the cross-*server*-process risk (only "in-flight runs"), given this repo's documented "servers share one cluster" pattern. | PR body's Feature 1 section **strengthened** with an explicit cross-server-hazard note (see above); the underlying inherited docs already flag the missing write-lock as a known, out-of-scope gap. |
| 6 | Low | `makeSpaFallbackMiddleware`'s constructor contract silently changed from "HTML string" to "file path" while keeping the parameter typed as plain `string` — a stale caller passing raw HTML wouldn't fail loudly, just 503 every route. | **Accepted** — confirmed exactly one call site in the repo, already updated correctly; internal (non-exported-as-public-API) refactor. |

## Folds

4 fork branches, cherry-picked/squashed into one clean commit per feature
(conventional commits, DCO signed off), built on `origin/main`:
- `goyamegh/evals-index-field-limit`
- `goyamegh/spa-fallback-fresh-html`
- `goyamegh/testcases-summary-fetch`
- `goyamegh/fix-category-matrix` (category-matrix commits only — see scope
  note in section 4 above)


goyamegh:
## codex_review (GPT-5.4) — adversarial review before requesting approval

Ran an adversarial second-opinion review against this diff + PR body (self-contained brief: all 4 features' diffs, the incident story, and the scope-exclusion judgment call on feature 4). Findings and disposition:

|---|---|---|---|
| 1 | **High** | `TestCaseEditor`'s full-record refetch cleared its loading flag in a blanket `.finally()` regardless of success/failure, re-enabling Save with stale/truncated data on a network error or a `null` (404) response — reintroducing the exact data-loss bug the refetch exists to close. | **Fixed** in this PR — Save now stays disabled with a visible, retry-able error banner on either failure mode. New unit tests added (rejected fetch, null response, retry-recovers). |
| 2 | Medium | `bulkCreate()`'s client-side return type says `errors: boolean`; the actual server/storage-adapter contract is `errors: number` (`server/adapters/types.ts` et al) — a pre-existing type-contract lie, not introduced by this PR. | **Not fixed here** — pre-existing on `origin/main` before any of the 4 folded branches, spans benchmarks/runs/test-cases client types (out of scope for this PR), and confirmed unused/dead by every current caller (nothing reads `.errors` off a `bulkCreate()` result today) — no runtime impact, but worth a follow-up type-hygiene PR. |
| 3 | Medium | `lib/chunkedFetch.ts` bounds the *client's* request shape, not server-side OpenSearch call concurrency — the `ids` route still fires one `Promise.all` per request. | PR body's Feature 4 section **clarified** to state this precisely; the server-side behavior is pre-existing/deliberate (a comment in `testCases.ts` explains it replaced a slower sequential loop) and out of scope here. |
| 4 | Medium | The SPA-fallback rebuild-detection keys freshness on `fs.Stats.mtimeMs` alone; a filesystem with coarse mtime resolution could theoretically miss a same-instant rebuild. | **Accepted as a known, low-probability edge case** — real `npm run build` runs take multiple seconds, and changing this would trade away the whole feature's point (avoiding a read on every request). Not touched. |
| 5 | Medium | The manual `/api/storage/reindex` operator recipe (inherited docs, not authored for this PR) doesn't explicitly call out the cross-*server*-process risk (only "in-flight runs"), given this repo's documented "servers share one cluster" pattern. | PR body's Feature 1 section **strengthened** with an explicit cross-server-hazard note; the underlying inherited docs already flag the missing write-lock as a known, out-of-scope gap. |
| 6 | Low | `makeSpaFallbackMiddleware`'s constructor contract silently changed from "HTML string" to "file path" while keeping the parameter typed as plain `string` — a stale caller passing raw HTML wouldn't fail loudly, just 503 every route. | **Accepted** — confirmed exactly one call site in the repo, already updated correctly; internal (non-exported-as-public-API) refactor. |

Applied: finding #1 (the only one that was an actual correctness bug in this diff, not a pre-existing/out-of-scope item). Findings #2–#6 are either pre-existing (not introduced by this PR), deliberate/inherited tradeoffs, or documentation-precision issues — addressed by clarifying this PR's own body language rather than changing code that isn't this PR's to redesign.

Full diff + updated test counts (unit 4826/4831 incl. the finding-#1 regression tests, integration 775/775) after applying the fix: see PR description.

goyamegh:
## Verification proof

**Commits (4, all DCO signed off):**
```
62252ed6 fix(comparison): category matrix renders subcategory rows on real data
b005bc14 fix(ui): stop pulling the full test-case corpus into list views
0e8faa36 fix: SPA fallback re-reads dist/index.html on mtime change (no restart needed)
29c7eb1f fix(storage): stop evals_runs metrics/judgeMetrics from blowing up total field-count mapping
```
`git log origin/main..HEAD | grep -c Signed-off-by` → `4` (one per commit).

**Build:**
```
$ npm run build:all
> tsc && vite build
✓ 4436 modules transformed.
dist/index.html, dist/assets/*.css, dist/assets/*.js  ... built in ~20s
> esbuild server/index.ts server/app.ts --bundle --platform=node --format=esm --outdir=server/dist
  server/dist/index.js  1.3mb
  server/dist/app.js    1.3mb
> esbuild cli/index.ts ... cli/dist/index.js  424.0kb
> tsc -p tsconfig.lib.json && tsc-alias ...
(exit 0, no errors)
```

**Unit tests:**
```
$ npm run test:unit
Test Suites: 271 passed, 271 total
Tests:       5 skipped, 4826 passed, 4831 total
Time:        ~40-51s
```
Per-file line coverage for every non-component (server/lib/services) file touched: `indexMappings.ts` 100%, `spaFallback.ts` 96.55%, `server/routes/storage/testCases.ts` 94.59%, `asyncRunStorage.ts` 97.52%, `asyncTestCaseStorage.ts` 100%, `opensearchClient.ts` 86.46%, `chunkedFetch.ts`/`comparisonInsights.ts` 100%.

**Integration tests** (against a real, non-ephemeral OpenSearch cluster, `--maxWorkers=2` to avoid shared-cluster contention from concurrent unrelated test activity):
```
$ AH_PORT=4651 npx jest --testPathPatterns='tests/integration' --maxWorkers=2
Test Suites: 78 passed, 78 total
Tests:       775 passed, 775 total
Time:        ~135-210s
```

**e2e (Playwright, real backend on an isolated port, never 4000/4001):**
```
$ PLAYWRIGHT_SKIP_WEBSERVER=1 npx playwright test tests/e2e/evals3-import-json-flow.spec.ts tests/e2e/test-cases-summary-fetch.spec.ts tests/e2e/comparison-category-matrix-scoped-fetch.spec.ts --reporter=list
  ✓ comparison-category-matrix-scoped-fetch.spec.ts › never fetches the unscoped test-case corpus, and the matrix renders with real categories (3.7s)
  ✓ evals3-import-json-flow.spec.ts › evals3 BenchmarksPage: importing a JSON file creates a benchmark with both test cases (26.5s)
  ✓ evals3-import-json-flow.spec.ts › evals3 TestCasesPage: importing a JSON file creates a benchmark with both test cases (26.1s)
  ✓ test-cases-summary-fetch.spec.ts › TestCasesPage list load requests fields=summary, not the full payload (2.2s)
  ✓ test-cases-summary-fetch.spec.ts › Dashboard test-case count requests summary + size=1, not the full payload (2.4s)
  ✓ test-cases-summary-fetch.spec.ts › BenchmarksPage list load + New Benchmark editor request fields=summary (11.0s)
  6 passed
```
(`editing a test case...` in the same file and both `comparison-insights-band.spec.ts` cases fail against this specific shared cluster for pre-existing, non-code reasons documented in the PR description's Test Plan section — reproduced and root-caused, not swept under the rug.)

**codex_review finding #1 fix, isolated verification:**
```
$ npx jest --testPathPatterns='TestCaseEditor'
✓ calls getById(testCase.id) on mount when editing an existing test case
✓ reseeds the form with the FULL context/expectedOutcomes/prompt, not the summary values
✓ does NOT call getById in create mode (testCase = null)
✓ disables Save while the full record is still loading, so stale summary data cannot be persisted
✓ keeps Save disabled and shows a retry-able error when the full-record refetch REJECTS (network failure)
✓ keeps Save disabled and shows an error when getById resolves null (e.g. 404 / deleted out from under the editor)
Tests:       6 passed, 6 total
```

**Field-limit incident, reproduced live** (this is *why* feature 1 matters, not a claim): while running the integration suite against the real shared cluster, a concurrent write to `evals_runs` from unrelated activity hit the exact pre-fix failure mode this PR's mapping change prevents going forward:
```
✖ Run failed: illegal_argument_exception: [illegal_argument_exception] Reason: Limit of total fields [5000] has been exceeded
```
Confirmed via a direct `GET .../evals_test_cases/_mapping` query that the cluster's `expectedOutcomes` field is separately, and permanently, locked to OpenSearch's dynamic `text` type from years of ordinary usage — the same mechanism (unbounded dynamic mapping growth) this PR's `dynamic: false` fix closes off for `evals_runs.metrics`/`judgeMetrics` going forward.

TackAdam (approved):

PR #445, opened 2026-08-28
Reviewer fixes: resource limits, judge authorization, import/link correctness, mapping compatibility, accessible navigation

ashwin-pc:
## Summary

This PR contains the owner-approved fix that applies to current `main`; the findings tied to features that have not merged to `main` were implemented directly on their existing open PR branches to avoid fabricating those features here.

- **#392 — restricted-command resource limits:** updated open PR #392 with regex-complexity guardrails and pre-read per-file/aggregate input limits (`801aa48`). Tests cover pathological regexes, oversized files/input sets, and normal regex use.
- **#392 — judge authorization:** updated open PR #392 to derive `agentKey` from stored evaluation-run/report metadata and reject caller mismatches (`a7fab41`). Tests cover mismatch rejection and run-derived keys.
- **#399 — import hashes:** updated open PR #399 to always recompute the canonical content hash and ignore/log stale supplied hashes (`189d661`). Regression test proves changed content with a stale hash creates a new document.
- **#399 — benchmark linking:** updated open PR #399 to link before terminal run finalization and make file/OpenSearch linking idempotent (`f4de381`). Tests cover duplicate linking.
- **#399 — legacy OpenSearch mappings:** updated open PR #399 with a keyword-or-legacy-text-compatible `docType` discriminator query and documented mapping contract (`7ba6ddb`). Test pins both mapping clauses.
- **#420 — canonical labels:** updated open PR #420 to reuse `parseLabels()` in `TestCaseDefinition` (`3eb0673`). Edge-case chip test pins canonical case-sensitive and invalid-difficulty behavior.
- **#400 follow-up — accessible mobile navigation (this PR):** replaces the custom drawer with the repository's Radix `Sheet`, giving focus trapping, Escape close, modal background inertness, scroll locking, and an unmounted/non-tabbable closed drawer. E2E assertions cover all requested keyboard behavior.

## Mobile verification (375×812)

Closed:

Closed drawer

Open:

Open drawer

## Verification

- `npx playwright test tests/e2e/mobile-responsive.spec.ts --workers=1` — 3 passed
- `npx tsc --noEmit` — passed
- `npm run build` — passed
- `npm run build:server` — passed
- `npm test` — 5,494 passed / 5 skipped; 50 pre-existing environment-dependent integration failures caused by the separately running port-4001 server/config fixture (not touched per task constraints). No failures involve `Layout` or `mobile-responsive`.


goyamegh:
**Advisory review (codex/gpt-5.4) — this PR is a draft, so this is not a merge gate, just tracking against the #392 blocker lineage.**

This PR's own description says the resource-limit and judge-authorization findings were "implemented directly on their existing open PR branches" rather than in this diff — confirmed: this PR's actual diff is unrelated (the Radix `Sheet`-based accessible mobile navigation drawer). The claimed fixes are three commits on **#392**'s branch: `801aa487d7` ("Bound restricted regex and file inputs"), `a7fab41015` ("Derive evidence agent from stored run metadata"), `9fd1fe8704` ("Mount judge workspaces without copying"). So checking this PR's claim means reviewing those commits on #392, which is what follows.

Checked against the two BLOCKER findings from the prior #392 review ([comment](https://github.com/opensearch-project/agent-health/pull/392#issuecomment-5432598284)):

| Severity | Finding | Notes |
|---|---|---|
| Major | **BLOCKER A (event-loop blocking via uncapped grep/sed) — only partially fixed.** `801aa487d7` adds real per-file (10MB)/aggregate (25MB) byte caps checked via `fs.stat` *before* any `fs.readFile`, closing the "slurp an arbitrarily large file into memory" half. But the regex guard is a narrow heuristic: it catches the classic `(a+)+` nested-quantifier shape, but I verified with actual regex tests that it **misses** alternation-based catastrophic-backtracking shapes like `(a|aa)+b` / `(a|ab)*c` (both pass the guard unflagged), while it also **false-positives** on a perfectly safe bounded pattern like `(ab+){2}` (rejected as "nested quantifiers" even though it can't blow up). A synchronous, uninterruptible `new RegExp(...).test()` against a missed pathological pattern on up to 10MB of text is still real event-loop-blocking risk. | Consider a regex-complexity library (e.g. detect star-height >1 more rigorously, or a timeout-capable engine/worker) rather than a single nested-quantifier regex heuristic. |
| Minor | **BLOCKER B (unbounded workspace copy → disk exhaustion) — fixed, and better than the originally suggested mitigation.** `9fd1fe8704` doesn't just cap the copy — it removes the copy entirely: the workspace is now exposed as a read-only virtual directory mount over the live source tree (new `DirectoryMount` type in `restrictedBash.ts`), with symlink rejection, path-escape rejection, and writes still confined to `scratch/`. No bytes are duplicated, so the disk-exhaustion vector is gone. This does trade an immutable snapshot for a live mount whose safety rests on a comment ("judge's post-quiescence contract: run workspace is immutable before judgment begins") rather than an enforced lock — a run-workspace cleanup or reuse race during judgment is a new (much narrower) integrity risk, worth a follow-up test/lock if that contract is ever violated in practice, but it is not a disk-exhaustion path. | Consider a lightweight lock/marker so a violated post-quiescence assumption fails loudly instead of silently reading stale/missing files. |
| Minor | **Major finding (caller-controlled `agentKey` → workspace leak) — fixed for the primary path.** `a7fab41015` now derives `trustedAgentKey` from stored run/report metadata and 403s on a mismatched caller-supplied `agentKey`; the workspace lookup keys off `trustedAgentKey`, not the caller's value. One residual gap: if `runId` is omitted or the storage lookup returns no record, `trustedAgentKey` stays `undefined` and the mismatch check is skipped (falls through to "no workspace resolved" rather than an explicit rejection) — a silent degrade rather than a leak, but worth confirming it's the intended behavior for that edge case. | Consider failing closed (reject) rather than silently resolving no workspace when `runId`/metadata can't establish a trusted agent key. |
| Info | This PR's own diff (mobile nav → Radix `Sheet`, focus trap/Escape/scroll-lock/inertness, matching e2e in `tests/e2e/mobile-responsive.spec.ts`) looks reasonable on its own merits and is unrelated to the above. | — |

**Verdict on the specific question this PR raises**: the accurate status is **"partially fixes both #392 blockers, not fully"** — the disk-exhaustion path (BLOCKER B) is closed by a structural improvement (mount instead of copy) rather than just a cap, but the event-loop/regex-DoS path (BLOCKER A) is only narrowed, not closed: the nested-quantifier heuristic both misses real ReDoS shapes and rejects at least one safe bounded pattern. None of this blocks #445 itself (draft, unrelated diff) — it's tracked here because #445's description is the pointer to where the #392 fixes actually live, and #392 is where the remaining regex-guard gap should be addressed before that PR merges.

PR #410, opened 2026-08-25
Run visibility: in-progress runs on benchmark page, running indicators, stale-run sweep

ashwin-pc:
Fixes #405, Fixes #406, Fixes #408 — three run-visibility gaps reported from real file-mode usage (a user started a run from a benchmark page and concluded it was lost).

## Changes
- **Evaluation Runs list (#406):** runs with `status: running` render a distinct animated **Running** badge with live progress ("2 of 3 cases", derived from per-case results vs. snapshot count), refreshing while visible — instead of a green check that reads as a finished run.
- **Benchmark detail (#405):** in-progress runs for the benchmark (from `GET /api/storage/evaluation-runs?benchmarkId=…&status=running`) appear alongside completed history, visually distinct, refreshing until completion lands them in `benchmark.runs[]` as today.
- **Startup stale-run sweep (#408):** on server start, eval-run docs stuck in `running` with no live execution registered are finalized as failed with an explanatory error ("interrupted: server restarted mid-run"), preserving any persisted per-case results. Idempotent.

## Screenshots
| View | Desktop | Mobile |
|---|---|---|
| Runs list: Running badge + progress; swept orphan shown finalized |  |  |
| Benchmark detail: in-progress run visible |  |  |
| Stale orphan finalized after restart |  |  |

Verified against a copy of real file-mode data with seeded running + orphaned runs; production build green; zero console errors on both viewports.

PR #447, opened 2026-08-28
Benchmark case review: Cases/Runs tabs, master–detail with suite health, per-run heat strips

### LINKED ISSUE #446: Benchmark page: reviewing the test cases in a benchmark doesn't scale
## Problem

The benchmark detail page splits the screen into `[test cases | runs]`. Reviewing the cases in a benchmark — the primary activity when validating or growing a suite — doesn't work well, and doesn't scale to benchmarks with hundreds of cases:

- The case list shows only names and labels. The prompt and expected outcomes are invisible until you navigate **away from the benchmark** to each case's page, so "go over the cases in this benchmark" is a click-out/click-back loop.
- Per-case cross-run signal — which cases discriminate, which are flaky, which fail everywhere — is not visible anywhere. You would have to open every run report and correlate by hand.
- Runs permanently occupy half the screen even when you are only reviewing cases; conversely, run rows show only aggregate counts (`2✓ 1✗ / 3`) with no way to see *which* cases failed without opening the report.
- Label chips render redundantly on case cards (category and difficulty each appear twice).

## Proposed design

The page answers two different questions, so give each a tab:

```mermaid
flowchart LR
    B["Benchmark page"] --> C["CASES tab (default)"]
    B --> R["RUNS tab"]
    C -->|"case's run strip → click"| RR["Run report"]
    R -->|"heat strip cell → click"| C
    C -.-> SH["suite health (nothing selected)"]
    SH -.->|"recent runs"| R
```

### Cases tab — master–detail

- **Left rail**: case list with name, difficulty chip, and a per-case sparkline of the last 5 run verdicts. Rollup counts (**needs attention / flaky / stable**) act as filters; text search over name and prompt; sort by pass rate (failing first). Keyboard `↑`/`↓` and `j`/`k` move the selection. Virtualization-ready for hundreds of cases.
- **Right pane**: the selected case in full — prompt, expected outcomes, context/fixture, provenance — plus a ✓/✗ strip of this case across recent runs, each chip linking to that run's report. Full pane width, so long expected outcomes render without truncation.
- **Nothing selected** → **suite health**: case/run totals, pass rate by difficulty (answers "is this suite still discriminating?" at a glance), needs-attention list, recent runs teaser linking to the Runs tab.
- **Mobile**: panes stack into list → case view; the case view is a **pager over the filtered list** (prev/next + position `3 / 14`, filter context strip, swipe where practical). Back returns to the list at the current position, so scan loops are resumable.

### Runs tab — existing run list, full width, plus per-case heat strips

Each run row gains a heat strip: one cell per case verdict, in a constant case order across all rows. Stacked rows read as the case×run matrix without rendering one — a vertical red seam is a case failing in every run; scattered red is flakiness. Hover shows the case name and verdict; click navigates to that case in the Cases tab. Existing actions (Add Run, Compare, Select all, delete) are unchanged. Strips are for pattern detection at any case count; precise navigation stays in the Cases tab.

### Bucket definitions (initial, disjoint by priority)

1. **needs attention** — failed ≥1 of the last 3 completed runs
2. **flaky** — mixed verdicts within the last 5 runs, but clean in the last 3
3. **stable** — passing

### Riding along

- De-duplicate label chips on case cards using canonical label parsing.
- Existing deep links (`…/benchmarks/:id/runs`) keep working and map to the Runs tab.

ashwin-pc:
Closes #446.

The benchmark page currently splits the screen into `[test cases | runs]`, which serves neither activity: reviewing cases means navigating away per case, and run rows show only aggregate counts. This PR restructures the page into two tabs, one per question:

### Cases tab (default) — master–detail case review

- Left rail: case list with difficulty chip, per-case **sparkline of the last 5 run verdicts**, rollup filters (**needs attention / flaky / stable**), text search, sort by pass rate (failing first), and `↑`/`↓` + `j`/`k` keyboard navigation.
- Right pane: the selected case **in full** — prompt, expected outcomes, context, provenance — plus a ✓/✗ strip of that case across recent runs linking to each run report.
- Nothing selected → **suite health**: totals, pass rate by difficulty, needs-attention list, recent runs.
- Mobile: panes stack; the case view is a **pager over the filtered list** (position indicator, prev/next, back preserves list position), using the existing component-scoped responsive utilities — no new global CSS.

### Runs tab — existing run list + per-case heat strips

Every run row gains a heat strip: one cell per case verdict in constant case order across rows, so stacked rows read as the case×run matrix (a vertical red seam = a case failing in every run). Hover = case name + verdict; click = that case in the Cases tab. All existing actions (Add Run, Compare, Select all, delete) unchanged; the legacy `…/benchmarks/:id/runs` deep link maps to the Runs tab.

### Implementation notes

- All aggregation is **client-side** from the latest completed runs' report documents (`lib/benchmarkCaseReview.ts`) — no new server endpoint, identical behavior in file mode and OpenSearch mode.
- Bucket definitions (disjoint by priority): needs attention = failed ≥1 of last 3 completed runs; flaky = mixed verdicts in last 5 but clean in last 3; stable = passing. ERRORED verdicts render as a distinct state, not pass/fail. Unit-tested in `tests/unit/lib/benchmarkCaseReview.test.ts`.
- The detail pane wraps the existing case renderer behind `BenchmarkCaseDefinition` so it can swap to the shared definition component from #420 in one line once that merges.
- Case-card label chips are de-duplicated via canonical label parsing.

### Sequencing note

This PR restructures `components/evals3/BenchmarkRunsPage.tsx` and will conflict with #399 (which restores a split view on the same page) and #410 (in-progress run surfacing). Whichever merges first, I'll rebase the others — the Cases tab here supersedes #399's split view, and #410's running-run indicators slot into the Runs tab.


goyamegh:
**Adversarial review (codex/gpt-5.4) at maintainer's request — correctness/regression + over-engineering pass.**

I read the full diff directly (9 files, ~1790 lines) and cross-checked the flagged behaviors against the actual source in `lib/benchmarkCaseReview.ts` and `components/evals3/BenchmarkCasesTab.tsx`/`BenchmarkRunsPage.tsx` before including them below.

|---|---|---|
| Major | `deriveCaseVerdict()`'s precedence checks `report?.metricsStatus === 'error'` **before** `report?.passFailStatus`, so a case that genuinely passed but whose metrics/cost enrichment failed renders as `errored` (→ `needs-attention` bucket, amber cell) instead of `passed`. Verified: `if (report?.metricsStatus === 'error') return 'errored'; if (report?.passFailStatus === 'passed') return 'passed';` — the order matters and metrics failure currently outranks a real pass/fail verdict. | Consider checking `passFailStatus` first when both are present, or documenting explicitly that a metrics-collection error is intentionally treated as more urgent than a known-good verdict. |
| Major | Removing the `testCaseVersion` selector (comment: "Cases always reflect the benchmark's current canonical case order") means the Cases tab can no longer browse a benchmark at a historical version — clicking through from an old run's heat-strip cell now shows the *current* case definition, not the one that actually produced that run's result. This is a capability regression versus main today, not just a UI simplification. | Worth confirming this is an intentional, accepted tradeoff (not just dropped in the restructure) — if version-pinned case review still matters to any user, it needs a path back in, even if not on this tab. |
| Major | No e2e/interaction test was added for any of the new UI-visible behavior — confirmed via the file list (only `tests/unit/lib/benchmarkCaseReview.test.ts`, which tests the pure geometry/verdict helper functions, not the components). The mobile swipe pager, `j`/`k`/arrow keyboard nav, heat-strip click-through, and filter/search UI have zero component or Playwright coverage. Per this repo's own testing bar, UI-visible behavior needs e2e coverage that would fail if it broke. | Add at least one Playwright spec exercising: search+filter+select on desktop, and the mobile pager (swipe or button next/prev) end to end. |
| Minor | `components/evals3/BenchmarkCasesTab.tsx` hand-rolls ~200 lines of touch/swipe/animation state (raw DOM refs, `requestAnimationFrame`, `setTimeout`-staged transform/opacity strings) for the mobile case pager, with no dedicated test for the gesture state machine itself (only the pure `computePagerDrag`/`resolveSwipeDirection` math is unit-tested). This is a large, easy-to-regress surface for a page-transition effect. | If no existing motion primitive is already a project dependency, consider whether CSS `scroll-snap` or a simpler "just navigate, no animation" fallback would deliver most of the value with far less bespoke code to maintain. |
| Minor | `BenchmarkCaseDefinition.tsx` is an explicitly-labeled placeholder ("Compatibility wrapper... callers can swap this for the shared TestCaseDefinition from #420 in one line") — reasonable scaffolding given the sequencing note, but worth a tracking item so it doesn't quietly become permanent. | No action needed now; just don't let it outlive #420 landing. |
| Info (positive) | The new keyboard nav (`j`/`k`/arrows) correctly guards against interfering with typing: `if (target?.matches('input, textarea, select, [contenteditable="true"]')) return;` before handling nav keys. | — |
| Info (merge hygiene, not a code defect) | Attempted a tunnel-integration merge of this branch against current `main-goyamegh` (which already carries `computeRunStats` — the issue #242 "don't trust denormalized run.stats" fix — and a `runVersionFilter` per-benchmark-key + self-heal fix). Both conflict with this PR's version of `BenchmarkRunsPage.tsx`, because this branch predates them. A straight merge/rebase must re-apply those two fixes on top of the new Cases/Runs structure, not silently drop them — flagging since neither fix is visible in this diff. (Not counting the already-disclosed #399/#410 sequencing conflicts, which are already tracked in the PR description.) | Rebase onto latest `main` before merge; explicitly re-verify `computeRunStats` and the per-benchmark `runVersionFilter` self-heal survive the rebase. |

**Overall**: the Cases/Runs redesign itself is sound in shape (route-backed tabs, a well-tested pure module for verdict/bucket logic, sensible bucket-priority semantics with disjoint buckets). The two things worth a second look before merge are (1) the `metricsStatus`-before-`passFailStatus` precedence, which can misrepresent a real pass as needing attention, and (2) the complete absence of interaction-level tests for the substantial new touch/keyboard UI. Neither is a five-alarm fire, but both are exactly the kind of thing that's cheap to fix now and expensive to debug from a bug report later. The version-pinned case browsing removal and the upstream-fix rebase risk are worth a maintainer sign-off, even if they turn out to be intentional/already-planned.

goyamegh:
Few points, not sure of the following:
1. how you are filtering Needs attention use-cases 
2. Also, scrolling doesn't work. Hope its infinite scrolling.

goyamegh:
**Fix-pack pushed** (maintainer edit, `benchmark-case-review` @ `c38c20364`) addressing the review feedback above:

1. **`deriveCaseVerdict` precedence** — a real `passFailStatus` now outranks `metricsStatus === 'error'`, so a genuinely passed/failed case whose later metrics/cost enrichment failed no longer renders as `errored`. `lib/benchmarkCaseReview.ts:132-152`; unit matrix in `tests/unit/lib/benchmarkCaseReview.test.ts` ("precedence: a real passFailStatus outranks metricsStatus === 'error'").

2. **Historical-version notice** — the Cases tab always shows a case's *current* definition, even when reached from a specific historical run. Rather than rebuilding the removed per-version selector, added a minimal, honest banner when a run in the list scored an earlier version of the case text: `lib/benchmarkCaseReview.ts:275-309` (`hasHistoricalVersionMismatch`), wired in `components/evals3/BenchmarkCasesTab.tsx` (`showsHistoricalVersionNotice` + banner render, `data-testid="historical-version-notice"`).

3. **"Needs attention" filter — how it's computed:** the exact predicate is `classifyCaseVerdicts()` in `lib/benchmarkCaseReview.ts:158-178` — a case needs attention if it **failed or errored in ≥1 of its last 3 completed runs**, or (sparse-history fallback) it has run data but no pass anywhere in the last 5. That's now surfaced as a `title=` tooltip on each filter chip in `components/evals3/BenchmarkCasesTab.tsx` (`FILTER_TOOLTIP`), so it's discoverable in the UI, not just in source.

4. **Cases-tab scrolling on large benchmarks:** root-caused to CSS, not just a perf issue — the `cases` `TabsContent` (`components/evals3/BenchmarkRunsPage.tsx`) was missing `display:flex`, so its `flex-1 min-h-0` child never got a bounded height and the list's own `overflow-y-auto` had nothing to clip against (verified live: `clientHeight === scrollHeight`, i.e. zero overflow). Fixed by adding `flex flex-col` to that `TabsContent`. Also added real infinite-scroll windowing in `components/evals3/BenchmarkCasesTab.tsx` (60-row initial render + `IntersectionObserver` sentinel, `CASE_LIST_PAGE_SIZE`) so a 400+ case benchmark doesn't render every row up front. New e2e: `tests/e2e/benchmark-cases-scroll.spec.ts` (seeds a 90-case benchmark, drives a real scroll gesture, asserts the DOM grows incrementally to completion).

Side fix discovered while getting CI green (unrelated to the above, but blocking): two `tests/e2e/evals3-benchmark-runs.spec.ts` tests asserted the split/tabs layout toggle from #399 — this PR's own description says it supersedes that toggle, and it no longer exists in the codebase, so those tests were failing independent of this fix-pack. Updated them to assert the new fixed two-tab layout's default active tab instead. Also fixed one pre-existing dead relative link in `CHANGELOG.md` (referenced a since-removed/consolidated spec file) that was failing `link-check`.

**CI status on `c38c20364`:** build-and-test (Node 18/20/22), lint-and-typecheck, security-scan, license-check, check-changelog, link-check, release-rehearsal, e2e-tests, codecov/patch, codecov/project, Code-Diff-Analyzer, Code-Diff-Reviewer, DCO — all green. `integration-tests` is red on a single, unrelated, pre-existing test (`POST /api/storage/evaluation-runs/:id/promote` re-promotion guard in `tests/integration/server/routes/storage/evaluationRuns.integration.test.ts`) that reproduced identically across 3 runs; none of this fix-pack's commits touch that endpoint or file, so it's independent of this PR and out of scope for this fix-pack.

Verified locally: `npm run build:all` clean; full unit suite (332 suites / 5614 tests) green; targeted e2e specs (this PR's own case-review coverage plus the scrolling/version-notice/tooltip additions) green against a real server; live-rendered against the actual preview server (screenshot on request) confirming the tooltip text, the historical-version banner, and real scroll+load-more behavior on an 80-case seeded benchmark.

goyamegh (approved):
Tested on the preview tunnel + fix-pack verified (verdict precedence, historical-version notice, needs-attention tooltip, cases-tab scrolling). Integration failure is the known pre-existing promote-flake, unrelated.

PR #449, opened 2026-08-28
fix(comparison): explain (not hide) an empty category breakdown

goyamegh:
## Problem

On the compare page, a benchmark whose test cases have no `[bracket]`/`topic:` facet loses the entire "By category" section (matrix + shared-weakness callout) with **zero indication why** — nothing renders between the Agreement chips and Table Compare. Reported live on `bench-1787700505202-xumpq6ibm` (WixQA-400) comparing two `eval-run-*` runs: *"I don't see similar categories for comparing other runs like WixQA benchmarks"* — while `EnterpriseRAG-84` (whose test-case names carry a `[bracket]` tag) shows the matrix fine on the same page.

**This PR went through two revisions.** The first only added an explanatory empty-state where the matrix would render. On review, the owner correctly pushed back: *"I don't like the fix. Why did the categories not work for WixQA data?"* — explaining the blindness isn't the same as fixing it. This revision changes the categorization logic itself so the matrix actually renders WixQA's real categories.

## Root cause

WixQA-400's 400 test cases carry `category:expertwritten` (200) / `category:simulated` (200) — a real, informative 2-way split — but no `[bracket]` tag or `topic:` label. `extractRowCategory()` (`lib/comparisonInsights.ts`) is **by design** indifferent to the plain `category:` label, on the assumption every imported benchmark stamps one coarse value (`category:RAG`) on every case, which would be a useless single-column matrix. That assumption is correct for `EnterpriseRAG-84` but wrong for WixQA-400, where `category:` genuinely varies.

Ruled out before landing on this: cancelled/partial runs (both `eval-run-*` docs are fully `400/400` completed), and an id-space/docType join bug (`EvaluationRun.results` is correctly a `Record<testCaseId, {...}>` map regardless of `eval-run-*` vs `run-*` id prefix).

## Fix

- `categoryLabelIsUsableFallback(rows)`: true only when, among rows with no real bracket/topic facet, `category:` (normalized consistently) varies with each candidate value appearing on ≥2 rows.
- `extractRowCategoryEffective(row, useCategoryFallback)`: the existing extraction, extended with that fallback.
- `buildCategoryBreakdown` computes the decision once and exposes it (`usesCategoryFallback`) so every other consumer stays consistent with what the matrix displays.
- `detectSharedWeakness`'s `allFailShare` math and `ComparisonPage`'s click-to-filter both re-derived a row's category independently — both updated to use the same effective extraction, or a fallback-derived matrix column would either report a wrong 0% share or filter every row out of the table when clicked.
- The empty-state from the first revision stays, now firing strictly less often (only when `category:` is *also* absent, uniform, or too sparse).

## codex_review (gpt-5.4) — two rounds

**Round 1** (on the empty-state-only revision): found the copy said "subcategory/topic" but omitted the `[bracket]`-tag path — applied. Other findings (defensive coding for an already-guaranteed invariant, message granularity, cross-component silence assumption) considered and left as-is with reasoning in the review thread.

**Round 2** (on this categorization-logic revision) — found 2 real bugs in the fallback gate, both fixed before push:

| # | Severity | Finding | Fix |
|---|---|---|---|
| 1 | **High** | Variance check compared *raw* label casing while the resolver lowercases — `category:RAG`/`category:rag` would wrongly count as 2 distinct values, reintroducing exactly the redundant single-column matrix the guard exists to prevent | Normalize (lowercase) before computing distinctness, matching the resolver |
| 2 | **Med** | One row with any `[bracket]`/`topic:` facet disabled the `category:` fallback for the *entire* dataset, even for other rows with nothing better (a mixed-convention benchmark would lose the fallback for everything) | Scope the variance check to only the rows a real facet doesn't already cover |
| 3 | **Med** | `>1` distinct values with no floor — a single stray/typo'd `category:` value among otherwise-uniform rows could flip the fallback on | Require each candidate value to appear on ≥2 rows before counting as "distinct" |
| 4 | Low | Boolean-threading (`useCategoryFallback`) across 4 call sites vs. a centralized resolver/precomputed map | Kept — a precomputed `Map<rowId, category>` would need its own staleness/identity management across renders; a plain per-dataset boolean, recomputed via `extractRowCategoryEffective(row, flag)` (O(1)/row) where needed, is the minimal state actually required |
| 5 | Low | No page-level test proving a fallback-derived column survives `ComparisonPage`'s click-to-filter | Already present (my brief just didn't quote it) — `tests/e2e/comparison-insights-band.spec.ts`'s new test clicks the `expertwritten` column and asserts the table filters correctly |

3 new unit tests pin findings #1–3, all verified red against the pre-review code, green after.

## Tests

- **Unit** — `tests/unit/lib/comparisonInsights.test.ts` (+13 tests: fallback/effective-extraction behavior including the two codex-found edge cases, the WixQA-400 shape end-to-end through `buildCategoryBreakdown`, the uniform-`category:` anti-noise guard, the `allFailShare` fix) and `tests/unit/components/comparison/ComparisonInsightsBand.test.ts` (rewritten: real matrix for the WixQA-400 shape, empty-state still fires for a uniform `category:`).
- **e2e** — `tests/e2e/comparison-insights-band.spec.ts`: real matrix + click-to-filter against a real seeded evaluation-run pair reproducing the WixQA-400 shape, plus a companion case for the uniform-`category:` control.
- All new/changed assertions verified red before the relevant diff, green after (via `git stash`), at every stage of both revisions.

## Verification

```
$ npx tsc --noEmit -p tsconfig.json          # exit 0
$ npm run build:all                          # exit 0
$ npm run test:unit                          # 262/262 suites, 4794/4799 tests (5 pre-existing skips), 0 fail
$ npx playwright test tests/e2e/comparison-insights-band.spec.ts   # 3/3 passed
$ npm audit --audit-level=high               # pre-existing findings only (package.json/package-lock.json untouched)
```

Before/after screenshots against the exact reported URL (real shared-cluster data — WixQA-400 + the two reported `eval-run-*` ids) show the matrix now rendering `expertwritten`/`simulated` columns with real per-run pass rates; reviewed in the originating chat session, not attached here per this repo's convention of keeping proof images out of PRs/commits.

## Note for maintainers (found during root-cause work, unrelated to this diff)

The shared verification cluster's `benchmarks` index is already at OpenSearch's 5000-field cap for new `evaluation-run` writes (matches the existing `[Unreleased]` CHANGELOG entry noting an already-poisoned index needs a reindex it left out of scope). Worked around locally by running e2e verification against isolated file storage.


goyamegh:
## Verification

**Local gates (mirrors CI):**
```
$ npx tsc --noEmit -p tsconfig.json
exit 0

$ npm run build:all
exit 0

$ npm run test:unit
Test Suites: 262 passed, 262 total
Tests:       5 skipped, 4782 passed, 4787 total   (skips pre-existing, unrelated to this diff)

$ npx playwright test tests/e2e/comparison-insights-band.spec.ts
  ✓ agreement chips + category matrix render and filter the table (3.1s)   [pre-existing test, still passing]
  ✓ shows an explicit empty-state (not a silent hide) when cases have only
    a category: label, no [bracket]/topic facet (3.2s)                    [new test]
  2 passed
```

**Red → green, confirmed both ways** (`git stash` the component diff, re-run):
- New unit tests (`ComparisonInsightsBand.test.ts`): 2 of 5 fail against pre-fix code (the two empty-state assertions); all 5 pass post-fix.
- New e2e test: fails (`insights-categories-empty` not found, 10s timeout) against pre-fix code; passes post-fix.

**Live root-cause verification** (read-only against the reporter's exact benchmark/run ids, before writing any code):
- `bench-1787700505202-xumpq6ibm` (WixQA-400): sampled all 400 test cases via `/api/storage/test-cases?ids=...` — `category: expertwritten` (200) / `category: simulated` (200), **zero** `subcategory:` labels, no `[bracket]` tag in any name (`wixqa_expertwritten_0`, …).
- Both reported `eval-run-*` docs: `status: completed`, `400/400` results, each `results[testCaseId]` correctly shaped as `{reportId, status, passFailStatus}` — ruled out a docType/id-space join bug before concluding this is a data-shape gap in the UI, not a join defect.

**codex_review (gpt-5.4) pass on the diff** — findings and disposition:

| 1 | Med | `categoriesEmptyReason`'s branch selection trusts `meaningfulCategories.length === 0` to mean "no facet at all"; flagged as relying on an invariant from the sibling `lib/comparisonInsights.ts` (every non-empty `rows` always yields ≥1 category, even if `(uncategorized)`) rather than defending against it locally | Verified the invariant holds today (`extractRowCategory` always returns a string); left as-is — adding defensive code for a contract already guaranteed by the sibling module would be exactly the kind of speculative complexity finding #3 also flags |
| 2 | Med | Message text said "subcategory/topic label" but omitted the `[bracket]`-tagged-name path, which `extractRowCategory` also checks — inaccurate/incomplete for a fix whose whole point is *accurate* explanation | **Applied** — both messages now name all three facet sources (`subcategory/topic label or a [bracket]-tagged name` / `every subcategory/topic/[bracket] facet`) |
| 3 | Low | Two distinct empty-state strings (all-uncategorized vs. all-rolled-into-other) may be more precision than the bug needs; a single generic message would cover the reported case | Kept both — the "(other)"-rollup case gives a materially different, useful signal ("you have facets, they're just too sparse") vs. "you have no facets at all" (add a `subcategory:` label), and the added string is one line, not new logic |
| 4 | Low | New sibling `<div>` could interact with the existing `categoriesOpen` collapsible state or break an assumed-empty-DOM-region test | Confirmed no interaction: the new div is mutually exclusive with `{showCategories && ...}` and doesn't touch `categoriesOpen`; existing e2e test asserting `insights-category-matrix` `toHaveCount(0)` on toggle still passes unchanged |
| 5 | Low | The `rows.length === 0` case is left silent on the assumption `ComparisonOverlapBanner` always covers it elsewhere on the page; flagged as a cross-component assumption that could leave both banners absent in some reuse | Noted, not changed — true today (verified `ComparisonOverlapBanner` renders whenever `overlap.totalTestCases > 0 && runCount >= 2`, the same page-level conditions), and expanding scope to a component this PR doesn't touch is out of scope for a comparison-page-only fix |

Before/after screenshots against this exact benchmark + the two reported `eval-run-*` ids (real shared-cluster data, not a fixture) were reviewed in the originating session; not attached here per this repo's convention of keeping proof images out of PR comments/commits.

**Note for maintainers** (found during root-cause work, unrelated to this diff): the verification cluster's `benchmarks`/`evals_experiments` index is already at (or very near) OpenSearch's `index.mapping.total_fields` cap, so new `evaluation-run` writes intermittently 500 with `Limit of total fields [5000] has been exceeded`. This matches the existing `[Unreleased]` CHANGELOG entry for the `EvaluationRun.results` mapping fix, which explicitly notes an *already-poisoned* index needs a reindex it left out of scope — flagging since it made local e2e verification against that specific cluster flaky (worked around here by running the changed spec against isolated file storage instead).

goyamegh:
## Update: revised per feedback — categories now actually work for WixQA, not just explained

The previous revision (commit `d290126`) only added an empty-state explaining why the matrix was missing. Correctly called out as insufficient — here's the real fix (commit `ec75a5b`), which makes the matrix render WixQA-400's real `expertwritten`/`simulated` categories with correct per-run pass rates, plus two `codex_review`-found bugs fixed in the fallback gate itself before this push. Full writeup in the updated PR description above.

**Local verification (mirrors CI):**
```
$ npx tsc --noEmit -p tsconfig.json && npm run build:all && npm run test:unit
tsc: exit 0
build:all: exit 0
Test Suites: 262 passed, 262 total
Tests:       5 skipped, 4794 passed, 4799 total

$ npx playwright test tests/e2e/comparison-insights-band.spec.ts
  ✓ agreement chips + category matrix render and filter the table (pre-existing)
  ✓ renders a real matrix from the category: label (not an empty-state) for the
    WixQA-400 shape, and click-to-filter works
  ✓ shows the empty-state (not a redundant 1-column matrix) when category: is
    uniform and no [bracket]/topic facet exists
  3 passed
```

Red→green confirmed via `git stash` at every stage (empty-state-only revision → categorization-fix revision → both codex-review fixes), including the 3 new unit tests pinning the codex-found edge cases (case-insensitive variance check, mixed-convention benchmarks, single stray/typo'd label).

ps48 (approved):

PR #411, opened 2026-08-25
Report page: verdict-first Overview tab; stop marking no-trace runs as errored

ashwin-pc:
Fixes #407, Fixes #409 — reported from real file-mode usage: a user's passing cases displayed as "ERRORED / Score 0%" with a red "Traces never arrived" banner, while the actual judge verdict (PASS, 1.0) sat buried in the third tab.

## Changes
- **Verdict-authoritative headline (#407):** report Status/Score now derive from the judge verdict (`matcherResults`/`passFailStatus`). Trace availability renders as diagnostic metadata — a muted info note when the agent has no traces, a warning when traces were expected — and can never contradict the verdict.
- **No trace polling for `useTraces: false` agents (#407):** the runner records a neutral "traces not configured" marker instead of scheduling a poll that inevitably lands a `trace_timeout` error on the report.
- **Overview tab, default-first (#409):** judge verdict + score card, per-expected-outcome breakdown, expandable reasoning excerpt, key stats (duration, tool calls, tokens when present). Tab order: Overview | Test Case Output | Judge Evaluation | Traces | Annotations.
- **Storage-mapper fixes:** file-backed reports expose `timestamp` while the frontend mapper read only `createdAt` → "Invalid Date" and dropped judge/metrics fields. Mapping now preserves both.

## Before / after (same report, real data)
| Before | After |
|---|---|
|  |  |

Mobile: 

Unit tests cover status derivation (pass+no-traces, fail+no-traces, trace-timeout-with-traces-expected), the mapper date/field preservation, and Overview default-selection.

### Relationship to #443

#443 redesigned the **run-level** header (`RunSummaryBand` / `RunInsightsPane`) and deleted `RunSummaryPanel`. This PR fixes the separate **case-level** view in `RunDetailsContent`'s Overview: the judge-authoritative verdict, one row per evaluation check with evidence and step citations, parsing of numbered headers in legacy prose, and the rule that no trace data does not mean the evaluation errored.

The rebase therefore dropped our `RunSummaryPanel` edits because upstream deleted that file. It also dropped commit `dbe53b8` because trace-failure handling is now owned upstream by #481. Everything case-level remains in this PR.

### Before / after (same real data, origin/main vs this branch)

| Report | origin/main | This branch |
|---|---|---|
[...]

PR #450, opened 2026-08-29
Test cases: first-class pinned fixture envelope (type/ref/integrity/payload)

### LINKED ISSUE #448: Test-case fixtures: representation gaps and evolution path (directive convention → typed envelope)
> Originally scoped to a presentation fix (derived fixture summary line). Discussion since widened it to the underlying representation question; this issue now records the problem, the options considered with tradeoffs, and a proposed evolution path. The presentation fix survives as step 1.

## Problem

Agent benchmark cases need to start from a **prepared world** — a fixture. The concept is universal even though the shape varies by agent type: coding agents need a seeded workspace, SQL agents a seeded database, support agents a conversation/ticket state. The schema has no first-class concept for this.

**Mainline today has no vocabulary for this at all.** `context` items are plain `(description, value)` pairs with no delivery semantics — there is no channel for case data that must *not* reach the agent:

```jsonc
// mainline: the only vocabulary available
"context": [
  { "description": "fixture: cache-refactor", "value": "…" }   // ← nothing marks this as "not for the agent"
]
```

A harness (ours) makes fixtures work anyway by convention: its connector recognizes `fixture: <name>`, materializes a content-pinned workspace, and **silently withholds the item from the agent**. This is a hack, and it is only safe inside that one connector — any consumer that honors the schema as written delivers the fixture name straight to the agent, contaminating cases that test *implicit* inference (the fixture name `cache-refactor` **is** the answer).

```mermaid
flowchart LR
    TC["TestCase<br/>prompt · description · outcomes · context[]"] --> CI["context item<br/>'fixture: cache-refactor'"]
    CI -->|"any schema-honoring consumer:<br/>delivered"| A["Agent ⚠ contaminated"]
    CI -.->|"one specific connector:<br/>intercepts by convention"| CN["Connector<br/>materializes pinned workspace"]
```

**In review**: #438 / #439 (rendering in #420) propose per-item **dispositions** that would make the withholding explicit and product-enforced:

```jsonc
// proposed in #439 (open PR — not shipped behavior)
"context": [
  { "description": "fixture: cache-refactor", "value": "…", "disposition": "connector" },        // not delivered
  { "description": "Fixture manifest", "value": "{ files, sha256, scan }", "disposition": "documentation" }  // not delivered
]
```

That fixes the delivery mechanics; the gaps below are what remains even then.

**What the convention (plus the proposed dispositions) gets right** — and any change must preserve:

- **Reproducibility**: the manifest content-pins the world (file tree + sha256 + bait scan); run N and run N+1 test the same thing.
- **Contamination protection**: neither item is delivered to the agent — essential when the case tests *implicit* inference and the fixture name alone (`cache-refactor`) would leak the answer.
- **Backend parity**: cases remain plain data in both file mode and OpenSearch.

**The gaps** (why this issue exists):

1. **No non-delivered channel exists upstream today** — the convention is safe only inside the one connector that knows it; every other consumer leaks the fixture reference to the agent. #439 fixes the mechanics, but it is an open proposal, not shipped behavior.
2. **Validation** — `fixture: <name>` is a stringly-typed convention. A typo or missing fixture fails at run time, not import time; the product cannot validate a fixture reference it cannot recognize.
3. **Identity** — the product cannot answer "which cases use fixture X" without parsing per-harness conventions.
4. **Presentation** — even with #439's grouped rendering, the fixture appears exiled in the context sections; a reader learns what world the agent starts in *last*, though it is part of the scenario (the original scope of this issue).
5. **Semantic overloading** — a schema is also documentation, and field names are its headings. `context` tells a reader "information for the agent"; a connector directive and an audit manifest are not context in any natural reading. Dispositions make the delivery mechanics explicit, but the data is still filed under a heading that misdescribes it. That the connector *can* interpret items stashed there is a capability argument, not a design justification — overloading a field because we can is the wrong default.

### Option 1 — keep the convention, fix presentation (step 1, still proposed)

Derive a one-line summary from the structured items and render it with the case narrative:

> **Workspace fixture:** `cache-refactor` — 6 files, integrity-pinned, not disclosed to the agent

Derived, never hand-authored (cannot drift); links to the grouped context detail. Zero schema risk. Does not address validation or identity, and presupposes the definition rendering and dispositions currently in review (#420 / #428 / #439).

### Option 2 — a `cwd` field (rejected)

`"cwd": "/path/to/dir"` replaces pinned *content* with a *pointer to a location*:

| | `cwd` pointer | pinned fixture |
|---|---|---|
| Reproducible | ❌ directory drifts between runs | ✅ sha256-pinned |
| Portable | ❌ machine-local path; meaningless in shared storage/CI | ✅ case travels as data |
| Run isolation | ❌ agent mutates the shared dir; parallel runs collide | ✅ fresh materialization per run |
| Judge baseline | ❌ needs a pre-run snapshot (re-derives the manifest) | ✅ manifest is the baseline |

*Where `cwd` belongs*: the run record — which directory a run actually materialized is an execution fact, and connectors already record it in run metadata. `cwd`'s one real virtue (authoring ease) belongs in tooling: an authoring command that snapshots a directory *into* a pinned fixture.

### Option 3 — a filesystem-specific `fixture` field (rejected)

Bakes one connector family's world-shape into the product schema. The next connector wants `database_seed`, then `docker_image`, then `browser_profile`; the schema becomes the union of every harness's setup needs, and the eventual escape hatch is a generic bag — which is what context items already are.

### Option 4 — `before()` / `after()` hooks (rejected)

Cases stop being data: they cannot be fully rendered, hashed, versioned, or audited without execution; arbitrary setup code is an execution vector inside the evaluation product; generated worlds carry no content pin; executable hooks cannot be stored as OpenSearch documents. Note the declarative twin already exists: the manifest **is** `before`, and expected outcomes evaluated against the final workspace state **are** `after`.

### Option 5 — generic pinned-fixture envelope (proposed evolution, step 3)

Promote the *envelope*, not any payload shape. The product owns exactly the three properties it must be able to guarantee; everything else stays opaque to it:

```jsonc
"fixture": {
  "type": "filesystem-workspace",   // opaque discriminator — product never interprets it
  "ref": "cache-refactor",          // identity: queryable, validatable at import
  "integrity": "sha256:e3b1c442…",  // content pin: required, product-verified
  "payload": { "manifest": "…" }    // connector-owned; product stores and renders only
}
```

**Contract invariants** (enforced by the product, not per-case flags):

- A fixture is **never delivered to the agent** — nothing in the delivery path reads this field. If a case wants the agent told about its world, that is prompt content, authored as a deliverable context item. (An earlier draft had `"delivered": false`; it was removed as a constant masquerading as a choice.)
- `integrity` is required — an unpinned fixture is a pointer, i.e. option 2.

```mermaid
flowchart LR
    TC["TestCase<br/>prompt · description · outcomes<br/>context[] · fixture{}"] --> FX["fixture envelope<br/>type · ref · integrity · payload"]
    TC --> CI1["context items<br/>(prompt / documentation)"]
    FX -->|"payload interpreted by"| CN["Connector<br/>materializes world"]
    FX -->|"integrity + payload"| J["Judge / auditor baseline"]
    FX -->|"rendered first-class"| UI["Definition page"]
    CI1 -->|delivered| A["Agent"]
    FX -. "never delivered (contract)" .-> A
```

**Why this rigidity and no more**: identity, integrity, and non-delivery are the properties that make benchmark results comparable and trustworthy across connectors — the product must be able to guarantee them without trusting each harness to remember. Payload shape is where legitimate variety lives, so it stays flexible.

**Why promote it at all, with only one connector family today**: two independent arguments, one of which already holds.

1. *Semantic honesty (holds now)* — fixture data misfiled under `context` misleads every reader of a case definition today, at n=1 connectors. A named `fixture` field makes the definition self-describing.
2. *Cross-connector validation and identity (grows with adoption)* — import-time validation and "which cases use fixture X" queries require the product to recognize the concept.

The first argument means promotion need not wait for a second payload type; the open question for maintainers is migration cost and timing, not whether the concept has earned a field.

**Relationship to #438/#439**: the envelope removes fixture data from `context` entirely, so fixtures would no longer need the `connector` disposition. Dispositions remain valuable for genuinely contextual items with different audiences (delivered prompt context vs judge/reader documentation). If the envelope lands first, #439 shrinks in scope; if #439 lands first, it is the bridge that makes today's convention safe until the envelope exists. Either order works; they are complementary, not competing.

## Decision: implement the envelope directly

Earlier revisions of this issue staged the work (presentation fix → structured directive JSON → envelope later). That staging was hedging: every intermediate step is either rebuilt or discarded once the field exists, and the envelope stands on its own merits today (gaps 1 and 5). So the plan is simply:

1. **Implement option 5 as a PR**: `fixture` field, product-enforced invariants (integrity required; nothing in the delivery path reads the field), import-time validation, first-class rendering on the case definition (which subsumes the original presentation ask of this issue).
2. Existing cases without a `fixture` field are untouched (the field is optional); harnesses using context-item conventions migrate on their own schedule.
3. #439 (dispositions) remains complementary for genuinely contextual items with different audiences; fixtures simply stop needing it.

*(Revision notes: an earlier version gated the envelope on "demonstrated cross-connector need" — that weighed only the mechanical axis; the naming argument holds already. Another version presented context dispositions as existing behavior; they are an open proposal (#438/#439), and mainline has no delivery semantics at all — gap 1. The staged sequencing was then dropped in favor of direct implementation.)*

ashwin-pc:
Closes #448 (see the issue for the full design record: problem, options considered — `cwd` field, filesystem-specific field, before/after hooks — and why each was rejected).

Agent benchmark cases need to start from a prepared world — a fixture. The concept is universal even though the payload shape varies by agent type (workspace, database seed, conversation state), but the schema has no vocabulary for it: `context` items carry no delivery semantics, so harnesses resort to conventions the product can't see, validate, or protect. For cases that test *implicit* inference, an unrecognized fixture reference delivered to the agent is the answer key.

This PR adds an **optional, generic fixture envelope** — the product owns exactly the properties it must guarantee; connectors own the payload:

```ts
fixture?: {
  type: string;        // opaque discriminator (e.g. "filesystem-workspace") — product never interprets it
  ref: string;         // identity — enables "which cases use fixture X"
  integrity: string;   // required content pin ("sha256:…") — runs N and N+1 test the same world
  payload?: Record<string, unknown>; // connector-owned (e.g. file manifest) — opaque to the product
}
```

### Product-enforced contract

- **Never delivered to the agent** — nothing in the delivery path reads the field; a regression test asserts a case with a fixture produces delivered content containing no fixture data. There is deliberately no `delivered` flag: non-delivery is what "fixture" means, not a per-case choice.
- **Validated at import/create** — non-empty `type`/`ref`, `integrity` in `algorithm:digest` form; actionable errors at import time instead of silent run-time failures.
- **Part of case identity** — editing the fixture versions the case like any content edit.
- **Backend parity** — file mode and OpenSearch; `type`/`ref`/`integrity` are keyword-searchable, `payload` is stored opaque (non-indexed) following the existing pattern for unbounded objects, so manifests can't blow up index mappings.
- **Optional end to end** — cases without the field are untouched; no migration required.

### UI

The case detail renders the fixture as part of the scenario narrative — `ref`, pinned integrity, and explicit "not disclosed to the agent" language — with the payload collapsible. Cases without a fixture render exactly as before.

### Relationship to #439

Complementary: with this field, fixtures no longer need to ride in `context` items at all; dispositions (#439) remain valuable for genuinely contextual items with different audiences (delivered prompt context vs judge/reader documentation).


ashwin-pc:
**Merge-order map for the open PRs from this fork** (for whoever is sequencing — every PR is independently CI-green, so out-of-order merging only costs extra rebases; this order minimizes them):

| Tier | PRs | Why this position |
|---|---|---|
| 1 — no dependencies | #410 run visibility · #445 a11y nav · #459 storage projections | Pure UI/storage; merge anytime |
| 2 — data model first | **#450 fixture envelope** → #439 dispositions | Three later PRs read #450's schema; #439 shrinks once fixtures leave `context` |
| 3 — judge stack, strict order | #392 evidence judge → #453 JSON→SDK compile | #453's compiler calls #392's judge as its claim evaluator |
| 4 — consumers last | #411 case-level report (approved) · #455 connector registry | #411 renders what #453 emits; #455 is a large file-move and rebases trivially over everything else, while everything else conflicts painfully over it |

Dependency edges: #450 → #439 (context semantics), #450 → #453 (`workspace-diff` reads the manifest), #392 → #453 (judge API), #453 → #411 (per-outcome rows), #450 → #455 (pi-web connector verifies envelope integrity). #445 tracks #392's B1 lineage but has no code dependency.

Happy to re-sequence if you'd prefer a different order — just say which and I'll rebase accordingly.

goyamegh:

PR #451, opened 2026-08-29
fix: stop pulling the full test-case corpus (168MB) into list views

goyamegh:
## Description

Every list-view page (Dashboard, Benchmarks, Evals, Reports, Benchmark editor/results, Use-case compare) fetched the **entire test-case corpus with full bodies** via bare `GET /api/storage/test-cases`. On a real deployment (3,941 test cases) that response is **168 MB** and takes 16+ seconds to build — multiplied by every page mount. The dashboard fetched all 168 MB just to display a count.

### Root cause
- The summary projection (`?fields=summary`) never stripped `sourceCode` (full eval-file text) — ~165 MB of the ~168 MB.
- 9 call sites ignored the summary/pagination API added earlier and called bare `getAll()`.
- The JSON-import flow called `bulkCreate()` then re-fetched the whole corpus to discover new IDs **by name matching** (also a correctness bug on duplicate names); the bulk API already returns the created records.

### Changes
- `toSummary()` strips `sourceCode`; summary payload drops **168 MB → 1.8 MB (−99%)**.
- Dashboard: `getAll({summary:true, size:1})`, reads `total` — never materializes the list.
- 6 list views: `getAll({summary:true})` (every rendered field survives the projection).
- Import flows (BenchmarksPage, TestCasesPage, evals3 variants): take IDs from the `bulkCreate` response; the evals3 variants were reading a non-existent `.ids` field, so import there was silently broken.
- `TestCaseEditor` refetches the full record by id before editing (prevents saving a summary-truncated record).

### Testing
- **Unit**: 9 fetch-spy regression tests (one per page) that fail if any list view reverts to a bare full-body fetch; summary-projection tests; bulkCreate response-surface tests.
- **Integration**: summary+pagination `total` correctness; import flow uses bulk-response ids with no full re-fetch.
- **e2e (Playwright)**: network-asserts `fields=summary` on TestCasesPage/BenchmarksPage/Dashboard + edit-without-data-loss regression.

### Measured (real deployment, 3,941 test cases)
| Request | Before | After |
|---|---|---|
| Test-case list payload | 168,125,258 B / 16.6 s | 1,791,935 B / 14.2 s (list views now fetch summary) |

## Check List
- [x] New functionality includes testing — unit + integration + e2e


goyamegh:
## Verification update — codex-review follow-up (3 new commits: `e11549c7..22310b1c`)

Addressed the codex-review findings on this PR's `TestCaseEditor` refetch-guard. Summary of what changed, how it was verified, and current CI status.

### 1. Fail-closed on refetch failure (HIGH)

`TestCaseEditor`'s full-record refetch previously only `console.error`'d on failure and cleared the loading flag in `finally` — Save re-enabled against whatever (possibly summary-truncated) form state the caller seeded, reintroducing the exact data-loss bug the refetch exists to prevent.

Now: a `loadError` state keeps Save disabled and the entire form surface locked (read-only) whenever `getById` rejects **or** resolves falsy, with an inline "Couldn't load the full test case — editing disabled to prevent data loss" banner and a Retry button that re-runs the fetch. Only a successful full-record load re-enables Save.

### 2. Hydration race (MED)

Form inputs stayed editable while the refetch was in flight, so anything the user typed was silently overwritten the moment the full record arrived and reseeded state. Every `Input`/`Textarea`/`Button` in the form now takes an explicit `disabled={isFormLocked}` prop (`isFormLocked = isLoadingFullTestCase || loadError`).

Note: a wrapping `<fieldset disabled>` was tried first (less markup churn), but **jsdom does not implement the HTML fieldset-disabled-propagates-to-descendants algorithm** — confirmed empirically with a minimal repro (`<fieldset disabled><input/></fieldset>` → `input.disabled === false` under jsdom, even though real browsers correctly propagate it per spec). Switched to explicit `disabled` props on every control so the behavior is both correct in real browsers and actually testable/verifiable in jsdom-based unit tests.

### 3. VERIFY-3 — does a form-only save drop sourceCode/sourceFile/sourceHash/versions?

Traced the save path end to end: `TestCaseEditor.handleSave()` → `asyncTestCaseStorage.update()` (which itself fetches the current full record and merges the form-managed fields on top, client-side, before calling the HTTP client) → `PUT /api/storage/test-cases/:id` → server route (passes `req.body` straight through) → `FileTestCaseOperations.update()` / `OpenSearchTestCaseOperations.update()` (both do `{ ...current, ...updates, id, version, currentVersion, createdAt, updatedAt }` — a **merge onto the current document**, not a replace).

Since the client's `toStorageFormat()` never emits `sourceCode`/`sourceFile`/`sourceHash`/`versions` keys at all, and both server adapters merge rather than replace, these fields survive a form-only update at every layer.

**Verdict: fields ARE preserved.** Added regression tests proving it (rather than a code fix), at three levels:

- Unit (file backend): `tests/unit/server/adapters/file/StorageModule.test.ts` — form-only update payload preserves `sourceCode`/`sourceFile`/`sourceHash`/`versions`.
- Unit (OpenSearch backend): `tests/unit/server/adapters/opensearch/StorageModule.test.ts` — same, asserting the indexed document body directly.
- Integration (real HTTP + real shared OpenSearch cluster): `tests/integration/server/routes/storage/testCases.integration.test.ts` — create a test case with `sourceCode`/`sourceFile`/`sourceHash`/`versions`, PUT an edit with **only** form-managed fields, GET and assert the provenance fields survived.

### 4. Bonus finding while wiring the integration proof

`checkBackend()` in `testCases.integration.test.ts` compared `GET /api/storage/health`'s `status` field against `'connected'` — a value the health route has **never** emitted (`server/adapters/index.ts` `HealthStatus` and both `StorageModule.health()` implementations only ever return `'ok' | 'error'`). `backendAvailable` was therefore always `false`, and every test in that file silently no-op'd regardless of whether a real backend was running — confirmed by running the suite before the fix (17/17 "passed" in <1ms each) vs. after (17/17 passed, real HTTP round trips, 90s total). Fixed to compare against `'ok'`. Scoped to just this file — the same dead comparison exists in ~13 other integration test files repo-wide; flagged separately, not fixed here.

### Test evidence

```
Unit (repo-wide):        270 suites, 4804 passed, 5 skipped, 0 failed  (39.7s)
Integration (repo-wide): 75 suites, 769 passed, 0 failed  (run --runInBand to avoid
                          shared-cluster contention with concurrent workers; the default
                          parallel-worker run hits sporadic timeouts on this large shared
                          cluster — pre-existing/environmental, unrelated to this diff,
                          reproduced and confirmed independent of these changes)
TestCaseEditor unit:      12 tests (4 pre-existing + 8 new failure-path/hydration tests), all passed
Adapter unit (new):       2 tests (file + OpenSearch backend), all passed
Integration (new):        1 test (sourceCode/sourceFile/sourceHash/versions survive
                          form-only PUT), passed against the real shared cluster (5.8s)
e2e:                      tests/e2e/test-cases-summary-fetch.spec.ts — all 4 specs pass
                          (network fields=summary assertions + the edit-without-data-loss
                          regression), including the exact save-without-data-loss flow this
                          PR is built around
tsc --noEmit:             clean
npm run build:all:        clean
npm audit --audit-level=high: pre-existing findings only (vite, undici via unrelated
                          dev-tooling dependency) — package.json/package-lock.json
                          untouched by this diff
```

Performance baseline this PR fixes (unchanged, re-confirmed): **168,125,258 bytes / 16.6s → 1,791,935 bytes / 14.2s** for the equivalent list-view request, measured against the real shared cluster (3,941 test cases).

### CI status note

`DCO`, `Mend Security Check`, and the internal PR-review bot (`Code-Diff-Reviewer`/`Code-Diff-Analyzer`, a `pull_request_target` workflow) all ran and are green/neutral on the latest commit. However, **`ci.yml` (Build and Tests), `changelog.yml`, `dependency-review.yml`, and `links-checker.yml` — all `pull_request`-triggered workflows — have not run at all**, on any commit in this PR's history (verified via the Actions API by exact SHA for all 9 commits, before and after this push). The `pull_request_target` bot re-ran cleanly on the new push (confirming GitHub *is* receiving and processing the webhook), so this isn't a delivery/timing issue — something appears to be gating `pull_request`-event workflows specifically for this PR/branch. This likely needs a maintainer to check the Actions tab for an "approve and run workflows" prompt, or a repo-admin check of Actions permissions for this fork.

Text-only per policy — no screenshots/images attached.

goyamegh:
Closing as folded: this change ships inside #444 (Server & storage hardening — the `testcases-summary-fetch` commit is part of that fold, per the maintainer's call to keep #444 as the single vehicle). If #444 stalls, this can reopen as the standalone carve-out.

PR #452, opened 2026-08-31
test: shared cleanup harness for integration/e2e suites; fix leaked shared-storage entities and the always-false backend guard hiding 54 vacuous tests

goyamegh:
## Problem

Integration and e2e suites created real test cases, benchmarks, evaluation runs, and per-test-case report docs in whatever backend `AH_PORT` pointed at — and never deleted most of them. On a shared OpenSearch cluster (the default setup for this repo's dev workflow), that meant permanent, growing clutter mixed in with real team data. An audit of the suite found **8 files with no cleanup at all** and **9 files with partial cleanup** (happy-path-only, or missing one of several created entities).

The most common reason "partial" was the norm rather than the exception: `DELETE /api/storage/benchmarks/:id` and `DELETE /api/storage/evaluation-runs/:id` are simple document deletes with **no cascade** to the per-test-case report docs the run/benchmark points at (see `OpenSearchBenchmarkOperations.delete` / `OpenSearchEvaluationRunOperations.delete` in `server/adapters/opensearch/StorageModule.ts`). A test that dutifully deletes the parent it created still orphans every report doc underneath it.

## What this adds

- **`TestDataTracker`** (`tests/helpers/testDataTracker.ts`): a shared harness suites register every entity they create with (test cases, benchmarks, evaluation runs, per-test-case report docs, evaluators, images, custom agents, remote servers, assistant sessions). `cleanup()` deletes children before parents, is 404-tolerant, and never throws unless `AH_TEST_CLEANUP_STRICT=1` (CI sets this).
- **Crash-recovery ledger**: because this repo's Jest config sets `forceExit: true`, `afterAll` doesn't run when a worker is killed or times out. Every tracked entity is also appended to an on-disk ledger; `jest.globalTeardown.cjs` drains any ledger left behind by a dead run (pid-liveness check), so a crashed suite still gets cleaned up at the end of the overall test run.
- **Playwright `testData` fixture** (`tests/e2e/fixtures/test-fixtures.ts`): wraps the tracker as a per-test fixture so cleanup runs on Playwright's own teardown — unconditionally, regardless of pass/fail or `--grep` selection (the pattern that produced several of the leaks below: cleanup implemented as its own trailing `test()`, which silently never runs if an earlier test aborts the worker or `--grep` excludes it).
- **`scripts/sweep-test-data.mjs`**: a dry-run-by-default sweeper for pre-existing leaks. It only ever touches the unambiguous `ahtest-` prefix by default; a `--legacy` flag additionally matches specific historical literal names, one at a time, and prints every candidate for review before any `--apply`.

## Bugs found and fixed along the way (please look hardest here)

- **`__mocks__/@/lib/config.ts` hardcoded the backend port**, so every service-layer Jest suite silently ignored `AH_PORT` and always talked to `localhost:4001` — on a dev box, a live server. Test isolation for that whole layer was an illusion. Fixed to resolve `resolveBackendPort()` **per property access** (via getters), not once at module-import time — a suite that sets `AH_PORT` mid-run now takes effect immediately, without needing `jest.resetModules()`.
- **`checkBackend()` in five storage integration files compared `health.status === 'connected'`**, but the health endpoint actually returns `'ok'` on both the file and OpenSearch backends. That comparison was never true, so **54 tests silently early-returned without asserting anything** in every environment, forever. Un-skipping them surfaced three real, previously-invisible defects:
  - `asyncBenchmarkStorage`'s `deleteRun` / `refreshAllStats` / `refreshRunStats` built bare relative `fetch('/api/storage/...')` URLs. Node's `fetch` rejects relative URLs, so every non-browser caller (CLI, SDK, Jest, the localStorage→OpenSearch migration path) silently got `false`/`null` back.
  - The client-side write mapper (`toStorageFormat` in `services/storage/asyncRunStorage.ts`) dropped `evaluatorId` and `judgeModelId` before persisting, even though the read mapper and the server-side save path both carry them.
  - The OpenSearch adapter's `deleteRun` reported success for **any** run id on an existing benchmark (the removeIf script never checked whether anything actually matched), while the file adapter correctly returned `false` for a run id that doesn't exist — the two backends disagreed on the same request, and the route (which maps `false` → 404 "Run not found") behaved differently depending on which backend was active.
- **`evals_experiments` index mapping growth**: custom evaluator metric names minted new mapped fields index-wide until the index hit its 5000-field cap and started rejecting writes. Fixed with `dynamic: false` on the two growth vectors while keeping the four legacy metric names typed (mirrors the fix in #418, which covered a sibling index).
- **Session metadata had no delete path at any layer** — `ISessionMetadataOperations` only had `get`/`put`/`list`. Added `delete(agentKind, sessionId)` (+ the route + the file-adapter implementation) so its own e2e spec can clean up after itself instead of leaking annotated sessions permanently.
- **Removed a `--orphans` sweeper mode and several "list + delete by name" cleanup sweeps.** On real, shared data, "parent reference is missing" and "name looks test-ish" are both *not* proof that a run created the entity — real clusters hold benchmarks literally named `mstest` and `Jason Test`, and importing the bundled OTEL demo file produces docs named exactly `OTEL Demo: …`. Deleting on those signals deletes other people's data. This is now enforced at **runtime**, not by scanning test source for the anti-pattern: `TestDataTracker.cleanup()` only ever issues a `DELETE` for an id that was explicitly passed to `track()` — there is no code path in it that lists or enumerates storage — and a unit test drives that guarantee directly against the real implementation.

## A note on a type-only change

`bulkCreate`'s return type across the storage layer was corrected from `{ created: number; errors: boolean }` to `{ created: number; errors: number }` (every adapter always returned a count; the old type was simply wrong, which is exactly what let several of the vacuous tests above get away with asserting `toBe(false)` against a value that could never be `false`). This only affects internal service-layer code — none of `testCaseStorage`/`benchmarkStorage`/`runStorage`/`asyncRunStorage`/`AsyncRunStorage.bulkCreate` are re-exported from `lib/index.ts`, the package's only public entry point (`package.json`'s `exports` map declares no subpaths), so no external consumer of `@opensearch-project/agent-health` can reach these symbols. Not a breaking change to the published surface; see the `### Changed` entry in `CHANGELOG.md`.

## Testing

- `npm run test:unit` — 5003 passed, 5 skipped, 0 failed (286 suites).
- `npm run test:integration` — 786 passed, 7 skipped, 0 failed, run against a dedicated local server (port 4316, `BENCHMARK_RUN_RECOVERY_DISABLED=1 EVALUATION_RUN_RECOVERY_DISABLED=1`) wired to the same shared OpenSearch cluster the team's other tooling uses, per this repo's storage convention — never against the ports the CI/dev servers own.
- `npm run test:e2e` (Playwright) — all files this PR actually changes behavior in pass cleanly: `test-cases.spec.ts` (19/19), `code-sdk-observio.spec.ts`, and `session-metadata.spec.ts` (8/8, including the new DELETE endpoint's create → delete → confirm-404 round trip). A handful of unrelated e2e specs (pages doing full-collection scans, e.g. the Benchmarks/Test Cases list pages) hit timeouts against this shared cluster's current real-world scale (84 benchmarks, 1700+ test cases accumulated over months) — confirmed independently as an infra/query-latency artifact of this specific dev cluster (a raw OpenSearch client `count()` against the same index hung for over a minute before self-healing, unrelated to any code in this diff) rather than a regression; CI's own e2e job runs against a fresh, disposable environment.
- Added a dedicated cross-backend contract test (`tests/unit/server/adapters/deleteRunParity.test.ts`) exercising `deleteRun` against a real `FileStorageModule` and a realistically-mocked `OpenSearchStorageModule` for all three cases (existing run, missing run, missing benchmark), asserting identical results — plus a regression test proving the OpenSearch adapter now fails **closed** or an unrecognized update-result shape (verified the real `@opensearch-project/opensearch` v3.5.1 response shape against a live cluster rather than guessing).
- Added an integration round-trip test (`tests/integration/services/storage/runStorage.integration.test.ts`) proving `judgeModelId`/`evaluatorId` survive `saveReport()` → real server → `getReportById()`/`getReportsByTestCase()`.
- Every entity this PR's own test runs created was tracked and deleted; `scripts/sweep-test-data.mjs` reports `0 test-created` residue after the full local suite run.
- `npm audit --audit-level=high` — pre-existing findings unrelated to this PR (no `package.json`/`package-lock.json` changes in this diff); CI's audit step is informational-only.

CHANGELOG.md updated under `## [Unreleased]`.


goyamegh:
## Verification proof

**Build:** `npm run build:all` — green (server, cli, lib all build clean; one pre-existing `direct-eval` warning from `lib/testCases/loader.ts`, unrelated to this PR).

**Unit tests:** `npm run test:unit` — **5004 passed, 5 skipped, 0 failed** across 286 suites.

**Integration tests:** `npm run test:integration` — **787 passed, 7 skipped, 0 failed** across 76 of 79 suites (3 intentionally skipped: no OTel-instrumented backend / test-admin-endpoints not enabled in this run). Run against a dedicated local server on a non-default port (`AH_PORT=4316`, `BENCHMARK_RUN_RECOVERY_DISABLED=1 EVALUATION_RUN_RECOVERY_DISABLED=1`), wired to this project's shared OpenSearch storage cluster — never against the ports the team's live dev/CI servers own.

**e2e tests:** all Playwright specs pass in CI (`e2e-tests`: success, `e2e-gate`: success). Locally, the three specs whose behavior this PR actually changes pass cleanly against a real server: `test-cases.spec.ts` (19/19), `code-sdk-observio.spec.ts`, and `session-metadata.spec.ts` (8/8, including the new DELETE endpoint's create → delete → confirm-404 round trip).

**`npm audit --audit-level=high`:** pre-existing findings unrelated to this PR — `git diff origin/main..HEAD -- package.json package-lock.json` is empty (no dependency changes in this diff). CI's own audit step is informational-only (`|| true`).

### Before/after: the 54 previously-vacuous storage integration tests

Five integration files compared `health.status === 'connected'`, but the health endpoint returns `'ok'` on every backend — so the comparison was never true and every guarded test silently early-returned without asserting anything, in every environment, forever:

```ts
// BEFORE (server/adapters/file/StorageModule.ts and the OpenSearch adapter
// both report status: 'ok' — this check never fired):
return health.status === 'connected';   // always false → every "if (!backendAvailable) return;" fires

// AFTER:
return health.status === 'ok';          // matches what the backend actually returns
```

Waking those 54 tests up surfaced 3 real, previously-invisible defects (all fixed and now covered):
1. `asyncBenchmarkStorage`'s `deleteRun`/`refreshAllStats`/`refreshRunStats` built bare relative `fetch('/api/storage/...')` URLs — Node's `fetch` rejects relative URLs, so every non-browser caller (CLI, SDK, Jest, the migration path) silently got `false`/`null` back.
2. `toStorageFormat` (client write mapper) dropped `evaluatorId`/`judgeModelId` before persisting.
3. The OpenSearch adapter's `deleteRun` reported success for **any** run id on an existing benchmark (no actual match check), while the file adapter correctly returned `false` — the two backends disagreed on the same request.

Live proof for #3 — ran against the real shared OpenSearch cluster (`tests/integration/services/storage/asyncStorage.integration.test.ts`, new test):

```
✓ deleteRun removes a run that actually exists, against the REAL configured
  backend (true), and leaves the benchmark 404-safe once deleted (false) (3245 ms)

  1. addRun(bm.id, {id: runId}) → true
  2. getById(bm.id) → run is present
  3. deleteRun(bm.id, runId) → true          (an existing run WAS removed)
  4. getById(bm.id) → run is now absent      (real removal, not just a boolean)
  5. deleteRun(bm.id, runId) AGAIN → false   (regression guard: the original
                                               bug would wrongly return true here)
  6. deleteRun(missing-benchmark-id, runId) → false
```

### Storage-entity totals unchanged after a full integration run

`scripts/sweep-test-data.mjs` (dry-run) against the shared cluster, run immediately after the full 787-test integration suite:

```
evaluation-run: 155 total, 0 test-created
test-case: 1745 total, 0 test-created
benchmark: 86 total, 0 test-created
evaluator: 11 total, 0 test-created

[dry run] matched 0, deleted 0
```

`0 test-created` for every entity kind — every test case, benchmark, and run this suite created during the run was tracked and deleted by `TestDataTracker.cleanup()` (children before parents, 404-tolerant), leaving zero `ahtest-*` residue on the shared cluster.

### Cross-backend `deleteRun` parity (unit-level)

`tests/unit/server/adapters/deleteRunParity.test.ts` drives a real `FileStorageModule` (tmp-dir backed) and the `OpenSearchStorageModule` (client mocked to the exact response shapes verified against a live cluster) through the same three scenarios and asserts identical results:

```
✓ existing benchmark + existing run: both backends return true
✓ existing benchmark + missing run: both backends return false
✓ missing benchmark: both backends return false (never throw)
✓ OpenSearch adapter refuses to assume success on an unrecognized result shape (fail-closed)
```

TackAdam (approved):

PR #453, opened 2026-08-31
Compile declarative JSON cases into SDK v2 matcher sessions (one verdict/render path)

ashwin-pc:
Unifies declarative (JSON) and code-based (JS/SDK) test cases onto one execution path: JSON cases compile into the SDK v2 matcher session (RFC 004), so both share one verdict rule (all gates pass), one result shape (one `matcherResults` entry per check), and one rendering path.

### Defaults preserve today's authoring

```jsonc
"expectedOutcomes": ["The agent makes no changes"]        // plain string → judged claim, role: gate
```

All LLM claims batch into ONE judge call returning per-outcome results. Opt-in specificity per outcome:

```jsonc
{ "outcome": "engages with design", "role": "observe" }          // score-only, never gates
{ "outcome": "makes no changes", "check": "workspace-diff" }     // deterministic SHA-256 vs fixture manifest — no LLM
```

### Behavior delta (measured, with a correction)

Verdicts become mechanical (all gates pass) instead of the judge's holistic ≥70%-threshold call. Re-deriving stored reports initially showed 8 historical PASS→FAIL flips — but investigation revealed most were **harness artifacts**: the run harvester froze transcripts before agents' final post-worker synthesis turns (since fixed via settlement-based harvesting). A fresh complete-transcript run of the same cases passes 3/3 at 100% under strict gates. Historical stored verdicts are never rewritten; only new runs compile through this path.

### Relationship to open PRs

- #392: the compiler invokes the evidence judge as its batched claim evaluator (per-outcome output landed there).
- #450: `workspace-diff` reads the fixture envelope's manifest.
- #411: already renders per-entry matcher rows; no changes needed.


goyamegh:

PR #454, opened 2026-08-31
fix(judge): no fabricated score-0, no duplicated reasoning; judge rows answer WHY it failed / HOW to fix it

goyamegh:
## Description

The Judge Evaluation tab on a test-case run report was unreadable in practice: the verdict rendered as one giant red prose blob, duplicated twice, headlined by a bogus `score 0%` — while the actually-useful structure (per-dimension scores, per-fact verdicts, the judge's own improvement strategies) was either buried at the bottom of the expanded row or trapped inside the prose. Found on real runs of a custom evaluator (`a custom internal evaluator`).

### Bug 1 — `score 0%` on every verdict, even passing ones
The SDK `judge()` matcher headlined `raw.metrics?.accuracy ?? 0`. Custom evaluators declare their own scoring dimensions (e.g. `answer_correctness`/`trust_honesty`/`readability` with weights 0.55/0.30/0.15) and never emit the legacy `accuracy` key, so **every** verdict persisted `score: 0`.

**Fix:** `parseJudgeResponse` now computes a weight-averaged `overallScore` from the evaluator's own `scoringConfig.metrics` — **all-or-nothing** (a declared metric the judge failed to emit voids the overall rather than renormalizing over the emitted subset, which would flatter partial/malformed responses). Both matcher write paths (SDK `judge()` and the legacy `buildJudgeMatcherEntry`) headline `accuracy → overallScore → nothing` and never fabricate a 0. The UI additionally suppresses the bug signature on already-persisted rows (score exactly 0 alongside non-zero dimension metrics — an inconsistent combination only the old bug produced; a genuine all-zero failure keeps its honest 0%).

### Bug 2 — reasoning persisted and rendered twice
The write path mirrored the full `reasoning` string into `errorMessage` on failure; the panel rendered both — the same multi-KB prose once red (`error:`) and once under `reasoning`. Mirror removed on both write paths (`errorMessage` is reserved for real endpoint errors); the UI dedupes old persisted rows where the two fields are identical. Verified consumers: `expect().toPass()` already falls back to `reasoning`; no CLI/export reader keys off judge-matcher `errorMessage`.

### Redesign — the tab answers the only two questions that matter
For a failed judge matcher:
- **Header**: per-dimension score chips (`answer correctness 10 · trust honesty 35 · readability 78`) so N judge calls are scannable without expanding; failing dimensions tinted red.
- **Why it failed**: expected-vs-cited source-id mismatch + per-required-fact `STATED/PARTIAL/MISSING/CONTRADICTED` verdicts, conservatively parsed from the judge's reasoning prose (`lib/matchers/judgeReasoningParse.ts` — bounded 20KB parse window, no lookbehind, adversarial-input regression tests; output is explicitly labeled *derived from judge reasoning* and never replaces the verbatim text).
- **How to fix it**: the judge's own `improvementStrategies` — already persisted, previously buried at the bottom of the row — promoted to a first-class panel.
- **Full reasoning**: collapsed, single copy.

**Durable structured path:** judge prompts may emit `facts` / `failure_causes` / `evidence` JSON keys alongside `reasoning`. The server already captures unknown keys into `extraFields`; the SDK matcher now forwards them as `MatcherResult.judgeExtraFields`, and the UI prefers those (shape-validated) structured fields over the prose parse — upgrading an evaluator prompt upgrades the tab with zero further code changes.

**No additional LLM calls anywhere** — everything renders from the single existing judge response.

## Related Issues
N/A (found during operator review of real benchmark runs)

## Check List
- [x] New functionality includes testing (unit: judge matcher, accessor, response parser, prose parser w/ verbatim real-run fixtures, component render tests; e2e: `tests/e2e/judge-tab-whyfix.spec.ts`)
- [x] New functionality has been documented (CHANGELOG under Unreleased)


goyamegh:
## Verification proof

**Unit** (all local, node 24):
```
Test Suites: 286 passed, 286 total
Tests:       5 skipped, 5025 passed, 5030 total
```
Includes new suites: `judgeReasoningParse.test.ts` (14 — fixtures are verbatim reasoning strings from real persisted runs), `MatcherResultsPanel.test.ts` (15 — score-0 suppression, why/fix panels, fact chips, single-copy reasoning, keyboard toggles), `judge.test.ts` +6 (headline precedence accuracy→overallScore→none, no errorMessage mirror, judgeExtraFields forwarding), `judgeResponseParser.test.ts` +6 (all-or-nothing weighted overall).

**Integration**:
```
Tests: 768 passed, 768 total
```

**E2E** (new spec, real server + file storage):
```
✓ judge-tab-whyfix.spec.ts › failed judge matcher renders Why/Fix panels, fact chips, and a single reasoning copy (1.9s)
1 passed
```
The spec seeds a report carrying the exact pre-fix artifacts old data has (`score: 0`, `errorMessage === reasoning`) and asserts: no `score 0%` headline, dimension chips render, Why/Fix panels present, per-fact PARTIAL chips, and the reasoning text appears **exactly once** on the page.

**Manual verification on a live server** against a real failed run (wixqa smoke, `a custom internal evaluator` evaluator, claude-sonnet-4-6 judge): header shows `answer correctness 5 · trust honesty 55 · readability 78` chips (no bogus score 0%), Why-it-failed lists the missing required fact, How-to-fix shows the judge's 4 strategies with priorities, reasoning collapsed in a single copy.

**Design review**: an adversarial `codex` (gpt-5.4) pass on the diff produced 3 applied hardenings — all-or-nothing `overallScore` (no flattering renormalization over partial judge output), zero-suppression narrowed to the inconsistent bug signature only (genuine all-zero failures keep their honest 0%), and the prose parser rebuilt without lookbehind (older-WebKit import-time crash) plus a bounded 20KB parse window with pathological-input regression tests.

ps48 (approved):

PR #456, opened 2026-08-31
test: raise unit-flag coverage — lines 73.6%→85.2%, functions 68.4%→78.2%, branches 62.2%→71.9%

goyamegh:
### Description
Test-only branch (zero production-source changes) that raises the jest unit-flag coverage across `services/`, `server/`, `lib/`, and `cli/`:

| Metric | Before | After | Δ |
|---|---|---|---|
| Lines | 73.56% | **85.23%** | +11.67pp |
| Statements | 72.64% | **84.24%** | +11.60pp |
| Functions | 68.39% | **78.21%** | +9.82pp |
| Branches | 62.23% | **71.94%** | +9.71pp |

~230 new tests across 17 new/extended suites. Biggest wins: `cli/commands/benchmark.ts` 0%→46% (largest gap in the repo; also solved the `import.meta.url` blocker that prevented importing the real module in tests), `server/adapters/file/StorageModule.ts` 30%→96%, `server/routes/storage/evaluationRuns.ts` 10%→96%, coding-agent readers (kiro/claudeCode/codex/workspaces) to 92–100%, `lib/telemetry/opensearchExporter.ts` 5%→98%, plus full-route suites for evaluators/codingAgents/config/skills/comparison.

An adversarial review pass was run over the diff; its three HIGH findings (a tautological id assertion, two race-prone cancellation tests, a vague bulkUpsert expectation) were fixed in the follow-up commit.

### Issues Resolved
Coverage-ratchet maintenance — gives codecov/patch headroom for in-flight feature PRs.

### Check List
- [x] New functionality includes testing (this PR *is* testing)
- [x] All tests pass (5230 unit / 299 suites; 768/768 integration)
- [x] Commit messages DCO signed-off
- [x] CHANGELOG.md updated under Unreleased

PR #457, opened 2026-08-31
fix: PiConnector duplicate --model, CLI exit fall-through, reader guard hardening

goyamegh:
## Summary

Batch of 4 source bugs found during the coverage-uplift session (`goyamegh/coverage-uplift-70`, PR #456), fixed together with regression tests for each. Each fix was verified red (fails against the pre-fix code) before green (passes after the fix).

### 1. `PiConnector` emitted two conflicting `--model` flags (`services/connectors/pi/PiConnector.ts:289-302`)

When both `connectorConfig.model` and `request.modelId` were set, the connector appended a `--model` flag while applying `connectorConfig`, then unconditionally appended a *second* `--model` flag for the request override — two conflicting `--model` flags in the final `pi` CLI argv. Same bug class as PR #435's `ClaudeCodeConnector` duplicate-arg fix.

Fix: `stripModelFlag()` removes any prior `--model` occurrence (both the two-token `--model <value>` form and a single-token `--model=value` form, in case `additionalArgs` carries one) before appending the request-derived one, so `request.modelId` always wins and exactly one `--model` flag survives. It also avoids swallowing an unrelated flag if a bare `--model` has no following value (malformed argv safety).

Tests: `tests/unit/services/connectors/pi/PiConnector.test.ts` (8 cases).

### 2. CLI `benchmark`/`report` commands fell through past `process.exit(1)` (`cli/commands/benchmark.ts:567-574`, `cli/commands/report.ts`)

`process.exit(1)` doesn't stop control flow by itself in a position TypeScript can't type as `never`, so several guards ran `console.error(...); process.exit(1);` with no following `return`/`throw`, then unconditionally used the value the guard was protecting against — a null-deref landmine whenever `process.exit` is stubbed (tests, or any future caller/wrapper).

- `benchmark.ts`: `agentKey = enabledAgent.key` ran immediately after the `!enabledAgent` exit guard. Extracted into an exported, directly-testable `resolveDefaultAgentKey()` that throws a descriptive `Error` after `process.exit(1)` instead of falling through to the null-deref (or returning a poisonous empty-string sentinel — see codex_review findings below).
- `report.ts`: the same pattern occurred three times — the `ensureServer` failure catch (fell through to `serverResult.baseUrl` on an unassigned variable), the benchmark-not-found guard (fell through to `benchmark.name`/`.id`), and the failed-report-fetch guard (fell through to **writing the error response body to disk/stdout as if it were a valid report**). Added an explicit `return` after each.

Tests: `tests/unit/cli/commands/benchmark.test.ts` (`resolveDefaultAgentKey`, 4 cases), `tests/unit/cli/commands/report.test.ts` (3 new fall-through regression cases hitting the real `command.parseAsync()` action).

### 3. `updateMemoryFile()` path-containment guard was bypassable (`server/services/codingAgents/readers/claudeCodeWorkspace.ts:145`)

The guard gating a file **write** was `resolved.startsWith(claudePath('projects')) && resolved.includes('/memory/') && resolved.endsWith('.md')`. A bare string-prefix check with no separator boundary let a sibling directory sharing the `projects` prefix (e.g. `~/.claude/projectsEVIL/memory/x.md` or `~/.claude/projects-leaked/memory/x.md`) pass; the loose `.includes('/memory/')` substring check was satisfied by any path containing that substring anywhere (e.g. extra nesting past the real `<slug>/memory/` segment).

Fix: `path.relative()`-based containment (any escape resolves to a relative path starting with `..`) plus an exact `<slug>/memory/<file>.md` 3-segment shape check. Additionally hardened against symlink escape (raised in codex_review): resolves the real path of the containing `memory/` directory via `fs.realpath()` and re-checks containment against the *real* projects root — free of legitimate-write breakage since a real write already requires that directory to exist.

Tests: `tests/unit/server/services/codingAgents/claudeCodeWorkspace.test.ts` (15 cases: happy path, prefix-trick/traversal/substring-decoy/non-.md bypass attempts, and 4 symlink-escape cases).

### 4/5. `kiro.ts:95`, `codex.ts:105`, `report.ts` line flagged during coverage drafting — assessed, no additional fix needed beyond report.ts's fall-through above

- `kiro.ts:95` (`userCount++` under `kind === 'Prompt'`) — plain counter increment guarded by the enclosing try/catch; no crash or wrong-guard risk found.
- `codex.ts:105` (`typeof item.output === 'string' ? item.output : ''`) — defensive type guard for a heuristic error-pattern match on tool output; a non-string `output` just skips the heuristic, no crash. Benign.
- `report.ts` line 105 (`if (options.stdout)`) sits immediately after the same `process.exit(1)` fall-through pattern as #2 above — fixed there.

## Note for #456 (coverage-uplift-70)

That branch's `PiConnector` unit tests currently **pin the duplicate `--model` flag as expected behavior** (documented as-is during the coverage session). If this PR lands first, #456 will need a small rebase/adjust to those specific assertions once its branch is rebased onto this fix.

## codex_review

Ran an adversarial `codex_review` pass (OpenAI gpt-5.4) over the diff. Findings and disposition:

| Finding | Severity | Applied? | Notes |
|---|---|---|---|
| `stripModelFlag` doesn't handle single-token `--model=value` form from `additionalArgs` | HIGH | ✅ Applied | Now strips both two-token and single-token forms. |
| `stripModelFlag` could swallow an unrelated flag after a bare `--model` with no value | HIGH | ✅ Applied | Only skips the next token if it doesn't look like another flag (`-` prefix check). |
| `resolveDefaultAgentKey`'s `return ''` sentinel after `process.exit(1)` is a poisonous value if `process.exit` is ever not a true halt | HIGH | ✅ Applied | Now throws a descriptive `Error` instead of returning `''`. |
| `isValidMemoryFilePath` is lexical-only and doesn't defend against a symlink planted inside the projects tree | HIGH | ✅ Applied | Added `fs.realpath()`-based containment re-check on the containing directory; fails closed, no legitimate-write breakage. |
| Exact 3-segment shape may be more rigid than the original substring check, blocking future nested memory-file organization | MED | ❌ Rejected | Mirrors the actual read-side contract in `getMemoryFiles()`, which only ever constructs `<slug>/memory/<file>.md` — no deeper nesting exists in the codebase today to break. |
| Inconsistent fatal-exit idiom between `benchmark.ts` (extracted helper that now throws) and `report.ts` (inline bare `return;`) | MED | ❌ Rejected | `report.ts`'s guards are inside a `void`-returning action — a bare `return` has no consumer to poison, so there's no equivalent risk to fix; introducing a shared abstraction for 3 one-off inline guards would be the over-engineering this same review process is supposed to catch. |
| Bundling 4 unrelated fixes into one PR makes review harder / rollback uglier | LOW | ❌ Rejected | Explicitly requested by the repo owner for this batch. |

## Test evidence (red → green, abbreviated — full transcripts in commit history / session)

**PiConnector** (pre-fix): `expect(modelFlagCount).toBe(1)` → `Expected: 1, Received: 2`
**PiConnector** (post-fix): all 8 cases pass.

**benchmark.ts `resolveDefaultAgentKey`** (pre-fix, guard removed): `TypeError: Cannot read properties of undefined (reading 'name')`
**benchmark.ts** (post-fix): all 4 cases pass, function throws a controlled `Error` instead.

**report.ts** (pre-fix): `TypeError: Cannot read properties of undefined (reading 'baseUrl')` / error body actually written via `writeFileSync('out.html', 'error body')`
**report.ts** (post-fix): all 3 cases pass, no writes/derefs occur on the failure paths.

**claudeCodeWorkspace.ts** (pre-fix): 4 of 11 path-bypass cases returned `true` (should be `false`) — real bypasses reproduced.
**claudeCodeWorkspace.ts** (post-fix): all 15 cases pass (11 original + 4 symlink-hardening cases added post-codex_review).

## Verification

- `npm run build:all` — green
- `npm run test:unit` — 306 suites, 5215 passed / 5 skipped / 0 failed (full local run, no regressions vs. baseline)
- `npm audit --audit-level=high` — pre-existing high-severity findings in `react-router`/`undici`/`vite` (unrelated dev/transitive deps, not introduced by this change; baseline also shows them on `origin/main`)
- DCO signoff on all 4 commits (`git commit -s`)
- `CHANGELOG.md` updated under `## [Unreleased]` → `### Fixed`


goyamegh:
## Verification proof (local, pre-push)

### Build

```
$ npm run build:all
...
✓ built in 19.85s   (vite build)
⚡ Done in 66ms      (esbuild server)
⚡ Done in 16ms      (esbuild cli)
tsc -p tsconfig.lib.json ... (no errors)
```
All four build steps (`build`, `build:server`, `build:cli`, `build:lib`) completed with no errors.

### Full unit suite

```
$ npm run test:unit
...
Test Suites: 306 passed, 306 total
Tests:       5 skipped, 5215 passed, 5220 total
Snapshots:   0 total
Time:        ~40s
```
No regressions vs. baseline (`origin/main` also reports 306/306 suites green).

### Red → green transcripts (abbreviated), one per bug

**1. PiConnector duplicate `--model` flag** — stashing only the source fix (tests kept):
```
● PiConnector › regression: does NOT emit two --model flags ...
  expect(received).toBe(expected)
  Expected: 1
  Received: 2
```
After the fix: `PiConnector` suite — 8/8 passed.

**2a. `benchmark.ts` fall-through** — with the `return`/throw removed:
```
● resolveDefaultAgentKey › exits with code 1 and throws ...
  TypeError: Cannot read properties of undefined (reading 'name')
    at resolveDefaultAgentKey (cli/commands/benchmark.ts:96:51)
```
After the fix: `resolveDefaultAgentKey` block — 4/4 passed (throws a descriptive `Error`, never null-derefs).

**2b. `report.ts` fall-through** — with the 3 `return;` statements removed:
```
● stops after a failed report fetch instead of writing the error body as the report
  expect(jest.fn()).not.toHaveBeenCalled()
  Expected number of calls: 0
  Received number of calls: 1
  1: "out.html", "error body"
```
(The error response body was actually written to disk as if it were a valid report.) After the fix: 3/3 passed, no writes/derefs on any of the 3 failure paths.

**3. `claudeCodeWorkspace.ts` path-containment guard** — with the fix reverted, guard back to
`resolved.startsWith(...) && resolved.includes('/memory/') && resolved.endsWith('.md')`:
```
● rejects a "-leaked" sibling directory sharing the projects prefix
  Expected: false
  Received: true
● rejects a path with a "/memory/" substring that is not the real memory segment
  Expected: false
  Received: true
● rejects a path with a "memory" segment that is a decoy inside a deeper subtree
  Expected: false
  Received: true
```
(4 of 11 original bypass attempts succeeded against the old guard.) After the fix: 15/15 passed (11 original + 4 symlink-hardening cases added after the codex_review pass).

### npm audit

```
31 vulnerabilities (2 low, 15 moderate, 14 high)
```
All pre-existing, in `react-router`/`undici` (transitive, via `@earendil-works/pi-coding-agent` dev dep)/`vite` — unrelated to this diff, present identically on `origin/main` before this PR.

### codex_review (OpenAI gpt-5.4, adversarial second opinion)

Ran against the full diff. 4 HIGH findings applied (PiConnector `--model=value`/malformed-argv handling, `resolveDefaultAgentKey` throw-instead-of-sentinel, `claudeCodeWorkspace.ts` symlink-escape hardening), 2 MED + 1 LOW rejected with reasons — full table in the PR description above.

goyamegh:
**Update:** CI's `integration-tests` job caught a real pre-existing gap — `tests/integration/services/connectors/PiConnector.integration.test.ts`'s `should apply model as --model arg` test set both `connectorConfig.model` and `request.modelId` and asserted `connectorConfig.model` ("claude-sonnet-4-5") won, which was exactly the pre-fix (buggy) behavior this PR fixes. Updated that test to assert the correct/intended precedence (`request.modelId` wins, exactly one `--model` flag survives) and force-pushed. Re-ran locally: 32/32 pass in that file, full `npm run test:unit` still 306/306 suites, 5215 passed / 0 failed.

ps48 (approved):

PR #458, opened 2026-09-01
fix(images): digest on legacy import path + migrate-images dry-run default

goyamegh:
## Summary

Two follow-ups surfaced by PR #437's evidence pack (content-addressed benchmark images / benchmark doctor):

1. **Legacy digest gap.** `POST /api/storage/benchmarks/:id/execute` — the route behind `benchmark -f test-cases.json` (single JSON file, no code-import) and `benchmark -n "Existing Benchmark"` — never stamped `imageDigest` on the resulting `BenchmarkRun`. Only the unified `POST /api/storage/evaluation-runs` path did. `BenchmarkRun` had no `imageDigest` field at all, so these runs were invisible to the content-addressed image/dedup story (`benchmark doctor --migrate-images`, image-scoped comparisons) with no warning. Fixed by reusing the *same* `computeImageDigest`/`buildImageDoc` helpers (`lib/benchmarkImage.ts`) the unified path already uses — one full test-case fetch shared between digest stamping and the existing SDK code-import re-materialization (previously fetched twice), failure-safe exactly mirroring `evaluationRuns.ts`'s pattern.

2. **`--migrate-images` had no dry-run.** It executed immediately on every invocation — the only mutating action in `benchmark doctor` not gated behind `--apply`, inconsistent with the rest of the command's dry-run-by-default contract. `POST /api/storage/images` now accepts `dryRun: true` and returns the real would-be digest/tags/`alreadyExists` without creating or tagging anything; `migrateBenchmarksToImages()` defaults to `dryRun: true`; the CLI only executes (creates/tags) with `--migrate-images --apply`, printing a "would create" / "already an image" plan otherwise.

## Tests

- Unit: `tests/unit/server/routes/storage/benchmarks.test.ts` (imageDigest stamped on the first persisted doc; failure-safe when the image write throws), `tests/unit/server/routes/storage/images.test.ts` (`dryRun: true` preview: no writes, real digest, `alreadyExists`, `missingTestCaseIds`), `tests/unit/services/benchmarkDoctor.test.ts`, `tests/unit/cli/commands/benchmarkDoctor.test.ts` (readOnly stays true for `--migrate-images` alone; `--apply` executes).
- Integration (real OpenSearch-backed server): `tests/integration/server/routes/storage/benchmarkExecuteImageDigest.integration.test.ts` (real sha256 digest + backing image doc via the legacy execute path; re-running identical content converges on the same digest), `tests/integration/cli/benchmarkDoctor.integration.test.ts` (preview computes the real digest and persists nothing; `--apply` then actually creates it; re-preview reports `alreadyExists: true`).
- Also ran full `npm run test:unit` (307 suites, 5232 tests, 0 failures) and `npm run build:all` clean.
- `npm audit --audit-level=high`: pre-existing findings only (react-router/undici/vite transitive deps) — no diff to package.json/package-lock.json in this PR.

## Verification

Both fixes verified red-then-green locally against a real OpenSearch-backed server (shared cluster) before/after the change.


goyamegh:

Ran an adversarial review of the initial diff before pushing the final version. Findings and disposition:

|---|---|---|---|
| 1 | Legacy digest stamped even when the full test-case fetch only resolved a **partial** set (deleted case / >10k corpus / transient gap) — silently produces a digest for less content than the run covers, a wrong identity not a harmless skip | HIGH | ✅ Applied — now requires `fullTestCases.length === benchmark.testCaseIds.length` before stamping, warns explicitly when partial |
| 2 | Images comparable-runs read path (`GET /api/storage/images/:digest`) only queries `EvaluationRun`, never legacy `BenchmarkRun`-embedded digests — "parity" claim overstated | HIGH | ⚠️ Documented, not fixed — needs a storage-layer query capability (embedded `runs[]` search) neither adapter has today; noted in-code + CHANGELOG as a tracked follow-up rather than expanding this PR's storage-layer surface |
| 3 | Dry-run preview fabricated a fresh `buildImageDoc()` (new `createdAt`, synthetic id) even when `alreadyExists: true`, instead of the real stored doc | MED | ✅ Applied — preview now returns the real stored image when it exists |
| 4 | `alreadyExists: true` could be misread as "`--apply` is a no-op" when the image exists under a *different* tag | MED | ✅ Applied — added `wouldAddTags`; CLI messaging updated ("already an image, would add tag(s)" vs "no change") |
| 5 | `getByDigest(...).catch(() => null)` swallowed storage read failures as "image doesn't exist", so a preview could report a false "would create" during an outage | MED | ✅ Applied — read failures now propagate to a 500, same as the non-preview path |
| 6 | Full-corpus test-case fetch moved earlier in `/execute`, lengthening time-to-first-byte before SSE headers flush | LOW | Not applied — fetch count is unchanged (one fetch, moved/shared, not added); accepted as a minor, bounded latency tradeoff for correctness (digest present on the very first persisted run doc) |

## Verification

- `npm run test:unit` — **307 suites / 5241 tests (5236 passed, 5 pre-existing skips), 0 failures**.
- Targeted integration tests against a real OpenSearch-backed server (shared cluster): `benchmarkExecuteImageDigest.integration.test.ts` (new — real sha256 digest + backing image doc via the legacy execute path; re-running identical content converges on the same digest), `benchmarkDoctor.integration.test.ts` (preview computes the real digest and persists nothing; `--apply` then actually creates it; re-preview reports `alreadyExists: true`), plus sibling `benchmarkExecuteEvaluator`/`benchmarkExecuteJudgeModel`/`imageDedup` integration suites — all green, no regressions.
- `npm audit --audit-level=high`: pre-existing findings only (react-router/undici/vite transitive deps) — no diff to package.json/package-lock.json in this PR.
- DCO signoff present on both commits.

ps48 (approved):

PR #460, opened 2026-09-01
feat(comparison): comparison page overhaul — all metrics on run rows, comparison-wide trace-grounded deep-dive (async), list-page running-run fixes

goyamegh:
## Summary

Overhauls the comparison page across 5 iterations (~12 commits) based on live owner UX feedback and a repeated bug-hunt against a real running tunnel/server.

### 1. Scoreboard declutter — every metric on the run rows
- Removed the recharts "Quality metrics" bar chart entirely (no charts in this flow).
- Every `RunAggregateMetrics` field now lives directly on the scoreboard row (Pass Rate, Cost, Avg Duration, Coverage, Avg Accuracy, Tokens, LLM Calls, Tool Calls) using the existing formatters — the "All metrics" expander and `MetricComparisonPanel` are deleted.
- The per-run expandable drawer (`RunDetailDrawer`) is removed; each row now has an inline "Open run" link + inline Remove (X) button.
- The judge model is shown once (not per row) via a single muted line, unless the compared runs use different judges.

### 2. Editable deep-dive system prompt (browser-cache only)
- `GET /api/comparison/deep-dive/system-prompt` returns the built-in default prompt.
- `POST /api/comparison/deep-dive` accepts an optional `systemPrompt` override (validated, capped at 20,000 chars).
- The "What's actually different" panel gained a collapsed-by-default editable textarea, prefilled from localStorage, with a "Reset to default" button. Nothing is persisted server-side.

### 3. Comparison-wide deep-dive
- The default prompt now receives a compact results table across every compared test case (`rows: ComparisonRowSummary[]`), instructed to selectively surface disagreements/score gaps/category patterns rather than following a fixed rubric.
- The deep-dive agent can now trace ANY case on demand via `query_spans`/`query_logs` (not just one fixed representative case), with case-aware span citations (`span:<caseId>:<runId>:<spanId>`).
- Fixed a bug where the scoreboard's Cost/Tokens/LLM Calls/Tool Calls columns showed `--` for trace-judged runs when a real value was available; kept the fallback honest (no fabricated proxies) where no data exists.
- Added a batch-metrics `session.id` correlation fix so span/trace lookups scoped to a comparison correlate correctly.

### 4. Live-bug-hunt fixes (round 4, found via a live tunnel repro)
- Deep-dive: elapsed-seconds loading indicator, a 200s client-side timeout with clear error+retry, and a server-side 180s generation deadline so a stuck agent loop fails clearly instead of hanging the HTTP response forever.
- Traces tab: fixed a permanent spinner when both compared runs have zero OTel correlators — resolves immediately to an honest empty state instead of issuing a doomed fetch; added a client-side fetch timeout with a clear "timed out" error.
- Evaluation Runs page / Benchmark Runs page: in-progress runs now show a "Running" badge and the true **planned** total (not just the count of test cases that have started), and both pages poll while a run is in flight. The Benchmark Runs page also now surfaces runs associated via `benchmarkId` that aren't embedded in `benchmark.runs[]` (previously invisible).
- Coverage banner reworded from "N shared / M total" to "N in both · M only in A" (names which side carries the extra cases).
- Scoreboard run-name cell: the hover tooltip now always matches the displayed text (previously blank when `runName` was falsy even though the row showed the agent-name fallback).

### 5. Async deep-dive job pattern + breadcrumb fix (round 5)
- `POST /api/comparison/deep-dive` now starts (or dedupes onto) a background job and returns `{ jobId }` in well under a second instead of holding the connection open for the full generation — this fixes a public-tunnel 524 gateway timeout that a synchronous long-running request was hitting even though the generation itself completes fine on localhost.
- `GET /api/comparison/deep-dive/jobs/:jobId` polls status (`running`/`done`/`error`); de-duped by an order-independent hash of the report pair + prompt inputs; capped at 3 concurrent generations (429 beyond that); jobs are swept 30 minutes after their terminal time.
- The client now polls every ~2.5s instead of holding one long-lived fetch open; existing loading/caching UX (elapsed counter, localStorage cache) is unchanged.
- Comparison page breadcrumb now names the benchmark for benchmark-scoped comparisons ("Evaluations > `<benchmark name>` > Compare Runs").
- Per owner feedback, the "Comparing A vs B · benchmark X" summary line added in round 4 was removed outright (not reworded) to avoid extra vertical space — benchmark identity now lives only in the breadcrumb, run identity only on the scoreboard rows.

## Testing

All test levels updated per level touched. Full details in `CHANGELOG.md` under `[Unreleased]`.

- **Unit**: `npm run test:unit` — **310 suites / 5308 tests passing, 5 skipped**, 0 failing.
- **Integration**: new suites added for the async deep-dive job endpoint (`tests/integration/server/routes/comparisonDeepDiveJobs.integration.test.ts`) and the benchmark-associated-runs filter (`tests/integration/server/routes/storage/evaluationRunsBenchmarkIdFilter.integration.test.ts`); existing `comparisonDeepDive.integration.test.ts` re-verified green.
- **E2E (Playwright, production build, matching CI's `workers: 2` config)**: **460 / 477 passing** (11 skipped, 1 flaky-then-passed-on-retry, 5 failing). All 5 failures + the 1 flaky test are in files this PR does not touch (`dashboard.spec.ts`, `skills.spec.ts` ×2, `evaluation-runner.spec.ts`, `evaluation-runs.spec.ts`, `test-case-runs.spec.ts`) — pre-existing Playwright strict-mode locator ambiguity / first-load timing issues unrelated to this change. Every comparison-page e2e spec (`comparison*.spec.ts`, 40+ specs) passes 100%.
- Manually verified against a real (non-mocked) local server: the async deep-dive job POST returned `{jobId}` in single-digit milliseconds while the generation continued in the background.

CHANGELOG.md is updated under `## [Unreleased]` with full per-round detail and file-by-file test references.


goyamegh:
## CI verification — all checks green

**Head SHA:** `82aba5709383fde720213e465e37ea63fa81f435`

| Check | Result |
|---|---|
| build-and-test (18/20/22) | ✅ success |
| integration-tests | ✅ success |
| e2e-tests | ✅ success — **473 passed, 3 flaky (recovered on retry), 11 skipped, 0 hard failures** (477 total) |
| release-rehearsal | ✅ success |
| lint-and-typecheck | ✅ success |
| security-scan | ✅ success |
| license-check | ✅ success |
| check-changelog | ✅ success |
| link-check | ✅ success (fixed in 82aba57 — see below) |
| DCO | ✅ success (all 13 commits signed off) |

### Local verification (pre-push), matching CI's config (production build, 2 workers)

- **Unit**: `npm run test:unit` — 310 suites / 5308 tests passing, 5 skipped, 0 failing.
- **Integration**: new suites for the async deep-dive job endpoint and the benchmark-associated-runs filter pass; existing comparison-deep-dive integration suite re-verified green against a real (non-mocked) local server — a `POST /api/comparison/deep-dive` returned `{jobId}` in single-digit milliseconds while generation continued in the background, confirming the tunnel-524 fix.
- **E2E** (Playwright, `CI=true`, production build, 2 workers — same config the CI job runs): 460/477 passed locally on the first pass before the CI-confirmed run above; every `comparison*.spec.ts` spec (40+ specs covering the scoreboard declutter, editable/comparison-wide deep-dive, async job polling, span citations, traces-tab states, coverage wording, breadcrumb) passed 100%.

### Fix applied after first CI run

The first CI run caught a real issue: `link-check` failed because `CHANGELOG.md` still had markdown links (`[path](path)`) to 4 files this branch deletes (`components/comparison/DeepDiveHeaderMetrics.tsx`, `components/comparison/AggregateMetricsChart.tsx`, `components/comparison/MetricComparisonPanel.tsx`, `tests/unit/components/comparison/DeepDiveHeaderMetrics.test.ts`). Commit `82aba57` converts those 8 occurrences to plain inline-code spans (the historical narrative stays, it just no longer links to a path that no longer exists). All checks are green as of that commit.

ps48 (approved):

PR #461, opened 2026-09-01
fix(traces): judge runId resolution for REST connectors + rest-agent service-name default

goyamegh:
## Summary

Fixes the code-layer half of a `trace_timeout` incident on `useTraces:true` + `agent-trace-judge` runs against REST-connector agents. The other half of the fix (an agent config's missing `traceServiceName`) is a local `agent-health.config.ts` change and is **not** part of this PR — see "Config vs. code split" below.

Once the service name is corrected (config-side), the poller finds the real spans **on attempt 1** — but that then exposed a second, previously-masked bug: `POST /api/judge`'s `agent` provider hard-requires a truthy `report.runId` ("runId is required for the agent (trace) judge provider"), and `RESTConnector.execute()` returns `data.runId || data.id || null` — this agent's response body has neither, so the judge call 400'd immediately instead of ever producing a verdict.

## Config vs. code split

- **Code (this PR):** two defects in the shared trace-correlation/judge code that (a) caused the *symptom* to be a silent 10-minute timeout instead of a fast, loud failure, and (b) blocked a real verdict even after the config was corrected:
  1. `resolveJudgeRunId()` — the `agent-trace-judge` provider's hard runId gate now has a safe fallback for connectors (REST, and any future connector) that never return a native `runId`.
  2. `resolveAgentServiceName()` — stops fabricating `${protocol}-agent` for connector protocols where the real service name is fundamentally unknowable (`rest`/`openai-compatible`/`langgraph`/`strands`/`mock`), and instead warns loudly + correctly skips Strategy C. This is the actual generalizable fix: it prevents *any* future REST/OpenAI-compatible/LangGraph/Strands agent with a missing `traceServiceName` from silently burning its entire poll budget on a name that can never match — the exact failure mode that made this incident take 10 minutes (times every run) to surface instead of being visible in server logs immediately.

## Changes

- `services/traces/judgeAgentsHints.ts`
  - `resolveJudgeRunId(report)`: `report.runId || report.traceId || undefined`. Used at the 3 trace-poll → judge call sites (`benchmarkRunner.ts` x2, `evaluationRunner.ts` x1) that previously forwarded `report.runId` directly.
  - `resolveAgentServiceName()`: added `UNKNOWN_SERVICE_NAME_PROTOCOLS` (`rest`, `openai-compatible`, `langgraph`, `strands`, `mock`). For these, returns `undefined` + a loud `console.warn` (naming the agent key + protocol) instead of guessing `${protocol}-agent`. Protocols agent-health itself owns (`claude-code`/`kiro`/`pi`/`agui-streaming`/`subprocess`) are unaffected. An explicit `traceServiceName` override always wins, silently. The `${protocol}-agent` last-resort guess is preserved only for connector protocol strings outside both tables (genuinely unknown custom connectors).
- `services/benchmarkRunner.ts`, `services/evaluationRunner.ts` — use `resolveJudgeRunId(report)` instead of `report.runId` at the 3 call sites.
- Tests: `tests/unit/services/traces/judgeAgentsHints.test.ts` (both new functions' fallback/warning behavior), `tests/unit/services/benchmarkRunner.test.ts` (regression: fallback forwarded to the mocked `callBedrockJudge` call, and `undefined` — not a fabricated value — when neither `runId` nor `traceId` is available).
- `CHANGELOG.md`.

## codex_review finding applied

An earlier draft of `resolveJudgeRunId` fell back to `report.runId || report.traceId || report.id`. An adversarial `codex_review` pass (gpt-5.4) flagged **`report.id` as unsafe** and it was removed before this PR:

> The safety argument is not airtight because it only reasons about `query_spans`. The same fallback value also scopes `query_logs`, and `/api/logs` treats any truthy `runId` as a cluster-wide `match` on `message` with **no time/agent window constraint**. … `report.id` should not be the final fallback; it converts a fail-closed 400 into silent mis-scoping.

Concretely: `server/services/logsService.ts`'s `fetchLogs()` treats a present `runId` as the *only* filter — `match: { message: runId }`, no time-range bound at all ("searching by runId, we want to find logs regardless of age"). `match` is analyzed, so a hyphenated report id like `run-<ts>-<rand>` tokenizes into common words (e.g. `"run"`) that would OR-match unrelated log lines across the entire shared cluster. A 32-hex-char OTel `traceId` has no word boundaries to split on, so it behaves as one high-entropy token with realistically zero false-positive risk — that's why `resolveJudgeRunId` stops at `report.traceId` and returns `undefined` (fail-closed, preserving the original 400) rather than reaching for `report.id`.

Two other findings from that review are **documented, not fixed, in this PR** (scope):
- `services/traces/browserRecovery.ts` (browser-side pending-report recovery) calls `callBedrockJudge` without `evaluatorId`/`runId`/`agents` at all — same 400 risk, different call site, tracked separately.
- `report.traceId` is set in-memory before the judge call but is currently dropped by `saveReportWithClient`'s (`server/services/storage/index.ts`) hardcoded field allowlist before persistence — it only survives for the remainder of the same process. Confirmed live during the smoke test below (the persisted doc has no `traceId`, but the in-process judge call still saw it).

Findings verdict was "not a clean small bugfix yet" on that earlier draft; both actionable findings from it (unsafe fallback, missing generalized guard against future wrong-service-name guesses) are applied in this PR.

## Extension: hints-only gate matrix (a THIRD, previously-masked 400)

`resolveJudgeRunId()` above only helps the two TRACE-MODE (`useTraces:true`)
judge call sites in `benchmarkRunner.ts`/`evaluationRunner.ts`, which stamp
`report.traceId` before judging. The two SYNCHRONOUS (`useTraces:false`)
judge call sites in `services/evaluation/index.ts`
(`runEvaluationWithConnector`'s "STANDARD MODE" — the path a REST agent takes
when it picks the `agent-trace-judge` model WITHOUT enabling trace-mode
polling) have no `report.traceId` at that point in the flow, but they DO
already forward Strategy C `agents` correlation hints
(`buildJudgeAgentsHints`, [#264](https://github.com/opensearch-project/agent-health/issues/264))
on every request. `POST /api/judge`'s gate was `if (!runId) return 400`
unconditionally — it never looked at `agents` at all, so it rejected these
requests even though the hints the tools needed were already sitting in the
same request body. The hints mechanism (#264) exists precisely for no-runId
agents; the gate just fired before anything downstream could use it.

- `services/traces/judgeAgentsHints.ts`: added `hasTraceCorrelation(runId,
  agents)` — true when `runId` is truthy OR at least one `agents` entry
  carries a `serviceName` or `sessionId`.
- `server/routes/judge.ts`: the `agent` provider gate now calls
  `hasTraceCorrelation()` instead of `if (!runId)`, 400-ing only when
  NEITHER is present (clear error naming both options). The cross-run/
  cross-tenant 403 defense-in-depth check (trajectory `runId` corroboration)
  now only runs when a `runId` was actually supplied — a hints-only request
  has nothing there to corroborate against.
- `server/services/traceJudgeTools.ts`: `query_spans`/`query_logs` used to
  hard-disable on `!runId`, which would have silently defeated the route fix
  (gate accepts the request, tools then report "no run id" and the judge
  quietly degrades to trajectory-only reasoning). Both tools now use the same
  `hasTraceCorrelation` check. `query_spans` sends `agents` alone to
  `/api/traces` when there's no `runId` (that route already treats `agents`
  as a first-class id filter). `query_logs` has no `serviceName` filter of
  its own on `/api/logs`, so it derives a `startTime`/`endTime` window from
  the widest span across the `agents` hints instead of falling back to
  `/api/logs`'s unscoped last-60-minutes default.
- Tests: [tests/unit/services/traces/judgeAgentsHints.test.ts](tests/unit/services/traces/judgeAgentsHints.test.ts)
  (`hasTraceCorrelation` gate matrix: runId-only, hints-only via
  serviceName+window, hints-only via sessionId, both, neither),
  [tests/unit/server/routes/judge.test.ts](tests/unit/server/routes/judge.test.ts)
  (route-level: hints-only requests reach `evaluateWithPiAgenticTrace`,
  empty/no hints still 400, 403 guard skipped on hints-only requests),
  [tests/unit/server/services/traceJudgeTools.test.ts](tests/unit/server/services/traceJudgeTools.test.ts)
  (tool-level scoping), and a new integration test
  [tests/integration/server/routes/judgeAgentProviderHints.integration.test.ts](tests/integration/server/routes/judgeAgentProviderHints.integration.test.ts)
  (real HTTP round trip through the express app + real `judge.ts` router,
  `evaluateWithPiAgenticTrace` mocked at its module boundary only — verified
  red against pre-extension code, green after).

## Verification

- **Unit:** `npm run test:unit` — 5237/5242 passing (5 pre-existing skips), including 23 new/changed tests across `judgeAgentsHints.test.ts` and `benchmarkRunner.test.ts`.
- **Integration (targeted):** `evaluationRunner.integration.test.ts`, `judge.integration.test.ts`, `benchmarkExecuteJudgeModel.integration.test.ts`, `agentTraces.integration.test.ts` — 58/58 passing.
- **Build:** `npm run build:all`, `tsc --noEmit` — clean.
- **Live end-to-end smoke test** (temp server, same shared storage/observability cluster as the reported incident, config-side `traceServiceName` fix applied locally + this PR's code): the poller found the agent's real spans on **attempt 1** (was a guaranteed 60/60 timeout), and `agent-trace-judge` returned an actual `PASSED` verdict (`answer_correctness=100`) backed by real `query_spans` tool calls against the run's OTel data — re-verified after tightening `resolveJudgeRunId` per the codex finding above. Test data created for this (`smoke-judge —` prefixed) was deleted after verification.
- `npm audit --audit-level=high` and the full `test:all` (Playwright e2e) have **pre-existing, unrelated** failures on this branch (known CVEs in `postcss`/`react-router`/`undici`/`vite`; a handful of integration tests that need a live `AH_PORT` backend and flake with `ECONNRESET`/timeouts in this sandbox) — neither is introduced by this diff; not addressed here to keep the change scoped.

## CHANGELOG

Updated under `## [Unreleased]` → `### Fixed`.


goyamegh:
## codex_review — adversarial findings (gpt-5.4, applied vs. rejected)

Ran before opening this PR, against an earlier draft of `resolveJudgeRunId` that fell back to `report.runId || report.traceId || report.id`. Findings:

| # | Severity | Finding | Applied / Rejected |
|---|---|---|---|
| 1 | **HIGH** | `report.id` fallback is safe for `/api/traces` (`query_spans`, no-op OR-clause) but **unsafe for `/api/logs`** (`query_logs`): `fetchLogs()` treats a truthy `runId` as an unbounded, analyzed `match: { message: runId }` query with no time window — a hyphenated report id (`run-<ts>-<rand>`) tokenizes into common words ("run") that pull in unrelated cluster logs. `report.id` should not be the final fallback; it converts a fail-closed 400 into silent mis-scoping. | **Applied** — dropped `report.id`; `resolveJudgeRunId` now stops at `report.traceId`, returning `undefined` (original 400 stands) when that's also absent. Re-verified live afterward: the smoke test's in-memory `report.traceId` was still available at judge-call time, so the fix still produces a real `PASSED` verdict. |
| 2 | **MED** | The `traceId` fallback is described as "Strategy A" but is never sent as a real `traceId` param — `traceJudgeTools.ts` still sends it as `runIds:[runId]`, matched only against `attributes.agent_health.run.id` / `attributes.gen_ai.conversation.id`, not literal trace-id equality. So it's really just a pseudo-run-id that's either a no-op or a stray match; the real correlation work is done entirely by the `agents` (Strategy C) hints. | **Rejected (documented, not fixed)** — correct observation, but threading a genuine `traceId` param end-to-end through `/api/judge` → `evaluateWithPiAgenticTrace` → `traceJudgeTools.ts` → `/api/traces` is a larger refactor than this bugfix's blast radius should cover. Doc comment on `resolveJudgeRunId` now describes the *actual* mechanism (safe no-op union) rather than mislabeling it "Strategy A". |
| 3 | **MED** | Missed call site: `services/traces/browserRecovery.ts` (browser-side pending-report recovery) calls `callBedrockJudge` without `evaluatorId`, `runId`, or `agents` hints at all — REST-connector reports recovered from the UI can still hit the original 400 or lose Strategy C entirely. | **Rejected (documented as follow-up)** — real, pre-existing gap, but a different call site/pattern (browser context, no server-side `agentConfig` access) than the 3 poller call sites this PR targets. Called out explicitly in the PR description and CHANGELOG as a tracked follow-up rather than silently left unaddressed. |
| 4 | **LOW** | Added tests mostly ratify the new fallback chain rather than the safety boundary itself; `mock.calls[0][6]` positional-arg assertion is slightly brittle; no route/tool-level test proves a fallback value can't widen trace/log scope. | **Partially applied** — kept positional-arg assertions (matches this file's existing test style, no named-args refactor introduced), but added the `resolveJudgeRunId` fail-closed-to-`undefined` test (finding #1's actual safety boundary) and the `resolveAgentServiceName` "no fabricated name + loud warning" tests for the generic-transport-protocol fix. Did not add a full `/api/logs`-route-level test proving the unbounded-match risk — the risk is structural (in `fetchLogs()`'s existing, unchanged behavior) rather than something this PR's code could regress. |

Overall verdict on the earlier draft: *"not a clean small bugfix yet... patches the symptom by fabricating a runId instead of fixing the judge/tool contract."* Finding #1 (the actionable safety bug) is fixed in this PR's final `resolveJudgeRunId`; #2/#3 are legitimate but out-of-scope observations, documented rather than silently dropped.

goyamegh:
## Text proof

### Unit tests (this diff's changed/new tests)

```
PASS tests/unit/services/traces/judgeAgentsHints.test.ts
  buildJudgeAgentsHints
    ✓ resolves the protocol-default service name
    ✓ returns a window hint with serviceName + bounded window
    ✓ anchors the window SYMMETRICALLY around the timestamp (subprocess timestamp = run start)
    ✓ threads report.sessionId onto the hint for Strategy D (#313)
    ✓ returns [] when no service name can be derived
    connectorProtocol=rest with no traceServiceName (2026-09-01 trace_timeout root cause)
      ✓ does NOT fabricate "rest-agent" — returns undefined instead
      ✓ logs a loud, always-on warning naming the agent and protocol
      ✓ buildJudgeAgentsHints returns [] for a rest agent with no traceServiceName (Strategy C correctly skipped)
      ✓ an explicit traceServiceName override still wins — no warning, no skip
      ✓ applies the same treatment to the other generic-transport protocols (openai-compatible, langgraph, strands, mock)
      ✓ known protocols we own (claude-code/kiro/pi/agui-streaming/subprocess) are unaffected — no warning
      ✓ preserves the last-resort <protocol>-agent guess for a genuinely unknown (not-in-union) connector protocol string
  resolveJudgeRunId
    ✓ prefers the real connector runId (Strategy B) when present
    ✓ falls back to the eval traceId (Strategy A) when runId is absent
    ✓ does NOT fall back to a fabricated value when neither runId nor traceId is set — fails closed
    ✓ treats empty-string runId/traceId as absent, not as a valid value

Tests:       16 passed, 16 total
```

Full `npm run test:unit`: **5237/5242 passing** (5 pre-existing skips, 0 failures).

### Targeted integration tests

```
PASS tests/integration/services/evaluationRunner.integration.test.ts
PASS tests/integration/server/routes/judge.integration.test.ts
PASS tests/integration/server/routes/benchmarkExecuteJudgeModel.integration.test.ts
PASS tests/integration/components/traces/agentTraces.integration.test.ts

Test Suites: 4 passed, 4 total
Tests:       58 passed, 58 total
```

### Build / typecheck

```
$ npx tsc --noEmit -p .          # exit 0, clean
$ npm run build:server           # esbuild — clean, no errors
```

### Live end-to-end smoke test (real trace-mode judge verdict)

Reproduced the production failure mode against a temp server pointed at the **same shared storage/observability cluster** as the incident, with the config-side `traceServiceName` fix applied locally plus this PR's code:

**Before either fix** — the exact reported symptom:
```
"llmJudgeReasoning": "**Evaluator could not run.**\n\n...\n\n**Reason (trace_timeout):** traces not available after 60 attempts (10 minutes)"
```

**After the config fix alone** (traceServiceName corrected, this PR's code NOT yet applied) — trace correlation works (attempt 1), but the judge call now 400s immediately instead of timing out:
```
[BedrockJudge] Attempt 1 failed: Bedrock Judge validation error (not retryable):
  runId is required for the agent (trace) judge provider — its trace tools scope to it
```

**After both fixes (config + this PR)** — a real, trace-backed verdict:
```
[BenchmarkRunner] Waiting for 1 trace-mode evaluations to complete...
[BedrockJudge] Sending request to backend proxy...
[BedrockJudge] Model: agent-trace-judge
[BedrockJudge] Attempt 1/10
[BedrockJudge] ✓ Evaluation completed in 50870 ms
[BedrockJudge] Pass/Fail Status: PASSED
[BedrockJudge] Metrics: { answer_correctness: 100, trust_honesty: 100, readability: 95 }
[BenchmarkRunner] All 1 trace-mode evaluations completed
```

The judge's `rawResponse` explicitly cites data only obtainable via the real `query_spans` tool call (a longer, more complete list of retrieved document ids than what's visible in the trajectory text alone) — confirming it reasoned over actual OTel spans, not just the trajectory narrative:

```
## Verification from Trace Data
The span data confirms:
- Question: "Looking for a sturdy manual card shuffler with a steel body..."
- Retrieved documents: ["474", "792013", "660029", "733947", "51411", "708010", "355527", "890532", "113561", "9210"]
- Gold product ID: 474
- Agent recommended: Product ID 474 (first/top result)
```

`traceFetchAttempts: 1` on the persisted report (vs. `60` / 10-minute timeout pre-fix).

All test data created for this verification (benchmark + reports, `smoke-judge —` prefixed) was deleted after verification; no residue left on the shared cluster.

### Known pre-existing, unrelated gate failures (not introduced by this diff)

- `npm audit --audit-level=high`: known CVEs in `postcss`/`react-router`/`undici`/`vite` (present on `origin/main` before this branch).
- Full `test:all` (Playwright e2e) and a handful of integration tests that require a live `AH_PORT` backend flake (`ECONNRESET`/timeout) in this sandbox — the specific integration tests relevant to this change (listed above) are green.

goyamegh:
This PR was extended (commit `c24fffa8`) with a fix for a THIRD, previously-masked 400 on this same code path — see the new "Extension: hints-only gate matrix" section above for the full writeup.

**tl;dr:** `resolveJudgeRunId()` (this PR's original fix) only helps the two TRACE-MODE (`useTraces:true`) judge call sites. The two SYNCHRONOUS (`useTraces:false`) call sites in `services/evaluation/index.ts` never had a `runId` OR a `traceId` to fall back to, but they already forward `agents` Strategy-C correlation hints ([#264](https://github.com/opensearch-project/agent-health/issues/264)) on every request — the `/api/judge` route's `if (!runId) return 400` gate never looked at them. Added `hasTraceCorrelation(runId, agents)` and used it at the route gate + in `traceJudgeTools.ts`'s tool-enablement check (so the tools don't silently degrade to trajectory-only reasoning once the gate accepts a hints-only request).

Proof (verified red against pre-extension code, green after):
```
tests/integration/server/routes/judgeAgentProviderHints.integration.test.ts
  POST /api/judge — agent provider, hints-only scoping (integration)
    ✓ reaches the agent trace judge over real HTTP with NO runId but a serviceName+window hint (the reported REST-connector bug)
    ✓ reaches the agent trace judge with NO runId but a sessionId-only hint
    ✓ still 400s over real HTTP when NEITHER runId nor usable agents hints are present
    ✓ still 400s when agents hints are present but empty of any usable field
    ✓ regression: a runId-carrying request still works unchanged (Strategy B alone)

Tests:       5 passed, 5 total
```
Full unit suite: 307 suites / 5254 passed, 5 pre-existing skips. `npm run build:all` clean. DCO signoff present on the new commit.

ps48 (approved):

PR #462, opened 2026-09-01
feat(runs): retry judgement for judge-failed cases — no agent re-run

goyamegh:
## What

Adds `POST /api/storage/evaluation-runs/:id/retry-judgement` to salvage a **TERMINAL** run whose agent executions completed fine but whose **judge** phase failed per test case (trace timeouts, judge 400s, "evaluator could not run"). Re-runs **ONLY** the judge pipeline (a bounded trace re-fetch for trace-mode agents, then the judge call) against each case's already-stored trajectory — the agent is **never re-invoked**.

Owner-reported motivating case: a single run (`eval-run-1788238635378-evbqzpx9g`) with 46 completed-agent / errored-judgement cases. "add a retry judgement button as I assume this may happen frequently ... retry judgement on failed kinda functionality at a run level."

## Selection predicate

Mirrors what the run-report UI already labels **ERRORED** (`getResultStatus()` in `components/evals3/ResultStatus.tsx`): `report.metricsStatus === 'error'` on a `completed` agent execution — written exclusively by `buildEvaluatorErrorPatch()` for both `judge_failed`/`trace_*` kinds and `agent_failed` (a genuine agent crash). The two are told apart by whether the report actually has a `trajectory` to re-judge — an `agent_failed` report has none, so it's excluded (nothing stored to salvage). `?scope=all` forces a full re-judge pass over every rejudgeable case regardless of current verdict.

## Server

- `services/evaluation/retryJudgement.ts` — `isJudgeFailedCase`, `hasRejudgeableOutput`, `selectRetryableCases`, `retryJudgementForCase`, `retryJudgementForRun` (bounded concurrency 3).
- For trace-mode agents, attempts a bounded (3-attempt) fresh trace re-fetch — a `trace_timeout` retry can succeed if spans have since landed — before falling back to the report's stored trajectory.
- Updates each report doc (verdict + matcherResults, or the canonical evaluator-error patch on a repeat failure), the run's `results` map, and recomputes `run.stats` via `lib/runStats` `computeRunStats`.
- `POST /api/storage/evaluation-runs/:id/retry-judgement` returns `{retried, succeeded, failed, results}`; 404 if the run doesn't exist; 409 if the run is still `running`/`pending`.

## UI

`components/evals3/RunInspectorPage.tsx` header gets a "Retry judgement (N)" button (eval-run mode only, next to Re-run) — disabled with a tooltip when N=0 or the run isn't terminal — opening `RetryJudgementConfirmDialog` (case count + judge model, then an in-place retried/succeeded/failed summary before the caller refreshes).

## Tests

- **Unit** (`tests/unit/services/evaluation/retryJudgement.test.ts`): selection predicate (all branches incl. agent-crash exclusion), trace re-fetch + fallback, stats recompute, repeat-judge-failure re-patch, `scope=all`. 96% line / 83% branch coverage on the new module.
- **Integration** (`tests/integration/server/routes/storage/evaluationRuns.retryJudgement.integration.test.ts`): real server + file storage + a MOCK judge (`judgeModelId: 'demo-model'`, routes to the built-in demo provider — no Bedrock/AWS creds needed). Covers 404, 409-while-running, judge-failed-only selection + stats recompute, agent-crash exclusion, `scope=all`.
- **E2E** (`tests/e2e/retry-judgement.spec.ts`): button renders with count, dialog opens with count + judge model, confirming shows a summary, disabled when there are no judge-failed cases.

All new tests pass; full `npm run test:unit` (5272 tests) and `npm run test:integration` (801 tests) pass with no regressions. `npm run build:all` is clean (tsc + vite + server + cli + lib). `npm audit --audit-level=high` shows only pre-existing transitive-dependency findings (react-router/undici/vite dev deps) — `package-lock.json` is unchanged by this PR.

CHANGELOG.md updated under `## [Unreleased]` → `### Added`.


goyamegh:
## codex_review (gpt-5.4) — findings & disposition

Ran an adversarial review against the full diff before merge, per this repo's own standing convention. Findings, severity, and what was applied vs. rejected:

| # | Finding | Severity | Disposition |
|---|---|---|---|
| 1 | 409-if-running gate is not a lock — two concurrent retry-judgement requests on the same run can both pass and race the final write | Blocker | **Applied** — added an in-process `activeRetryJudgementRuns` guard (mirrors the existing `activeCancellationTokens` pattern in this same file); a second request while the first is in flight now 409s. Covers same-process double-submit (the realistic case given this repo's single-writer-discipline convention); does not cover true multi-process races, same limitation every other mutation route in this file has today. |
| 2 | `scope=all` retrying a previously-passed case and then failing again leaves stale `matcherResults`/`improvementStrategies` inconsistent with the new `metricsStatus:'error'` | High | **Applied** — the failure path now also clears `matcherResults: []` / `improvementStrategies: []` alongside the canonical `buildEvaluatorErrorPatch`. |
| 3 | Selection predicate (`metricsStatus==='error'` + non-empty trajectory) can retry a `trace_incomplete`/`trace_timeout` case against a genuinely partial trajectory and produce a definitive verdict from incomplete evidence | High | **Rejected (documented, not a regression)** — this is the explicit scenario the owner asked to salvage by name ("trace_timeout ... retry judgement"). The retry attempts a fresh bounded trace re-fetch first (which can genuinely resolve `trace_timeout` if spans have since landed), and falls back to exactly the same trajectory fidelity the ORIGINAL run would have judged against on error — no new trust is extended beyond what a fresh run already does with the same data. |
| 4 | Final `evaluationRuns.update({results, stats})` is a stale-snapshot overwrite; concurrent writers to `run.results` are lost | High | **Mitigated by #1** — closes the realistic same-process window. Cross-process concurrent mutation of a terminal run has no guard anywhere in this codebase; out of scope for this PR. |
| 5 | Judge model resolution falls back to `report.modelId` (the AGENT's model) without an env default in between — risks silently judging with the wrong model | High | **Applied** — added `BEDROCK_MODEL_ID`/`AGENT_HEALTH_BEDROCK_MODEL_ID` env fallback between `report.judgeModelId` and `report.modelId`, matching the exact priority chain `waitForTracesAndJudge` already uses in `services/evaluationRunner.ts`. |
| 6 | Re-labelling every repeat failure as `judge_failed` loses the original failure kind (trace_timeout vs. judge 400, etc.) | Medium | **Partially applied** — the persisted `traceError`/reasoning now says `Retry judgement: <message>` so it's distinguishable from a first-attempt failure; the `kind` stays `judge_failed` because that genuinely describes what happened on THIS attempt (the trace re-fetch, if any, already ran and either succeeded or fell back — the judge call is what failed). Full provenance chaining (kind1 -> kind2) rejected as scope creep for this pass. |
| 7 | Synchronous 200 response risks browser/proxy timeouts for large N (bounded concurrency 3, could be minutes for dozens of cases) | Medium | **Rejected for this PR, tracked as follow-up** — matches the task's explicit "concurrency small (2-3)" instruction; converting to the fire-and-forget + poll pattern `/rerun` uses would require incremental per-case run.results writes (currently one batch write at the end) and a client polling loop in the dialog — a real design change, not a quick fix. Flagging for a follow-up PR if the owner hits this in practice with the 46-case run mentioned in the ask. |
| 8 | Confirm dialog's displayed count (`erroredCount`) can exceed what the server actually retries (agent-crash cases are silently skipped) | Medium | **Applied** — relabelled the field "Judge-failed cases" and added a one-line note that agent-crash cases are skipped automatically and the retried count may be lower. |
| 9 | Bounded-concurrency helper duplicates a similar helper already in `services/evaluationRunner.ts` | Low | **Rejected** — that helper is a private, unexported function in a large, actively-churned file; exporting it to share 8 lines of generic fan-out logic isn't worth the coupling for this PR. |
| 10 | `hasRejudgeableOutput` vs `isJudgeFailedCase` predicate split is speculative | Low | **Rejected** — they answer genuinely different questions ("is there anything to re-judge at all" vs. "did the judge specifically fail") and both are exercised by `scope=all` vs. the default scope; keeping them separate and named is clearer than one parameterized predicate. |
| 11 | Result array ordering is nondeterministic under concurrent fan-out | Low | **Applied** — results are now sorted by `testCaseId` before being returned, for a stable, testable summary. |
| 12 | `traceError: undefined` clearing idiom flagged as "selectively careful" vs. the broader retry-state story | Nit | **No change** — reviewer confirmed this specific idiom is correct given the storage backends' full-read-modify-write semantics; no action needed. |

All 5 codex-actionable fixes are in commit `fix: address codex_review findings on retry-judgement` (adds 5 unit + 1 integration test covering them). Full findings/reasoning kept here rather than assumed.

goyamegh:
## Verification proof

**TypeScript** (`npx tsc --noEmit -p .`): clean, no errors.

**Unit** (`npm run test:unit`): all green, including new module tests.
```
tests/unit/services/evaluation/retryJudgement.test.ts
  isJudgeFailedCase (5 tests) / hasRejudgeableOutput (1) / selectRetryableCases (3)
  retryJudgementForRun (9 tests: trace re-fetch success+fallback, happy path +
    stats recompute, repeat-judge-failure re-patch, matcherResults clearing on
    scope=all failure, BEDROCK_MODEL_ID fallback, no-op summary, scope=all,
    test-case-not-found)
Tests: 18 passed, 18 total

Full suite: Test Suites: 310 passed, 310 total
            Tests: 5 skipped, 5277 passed, 5282 total
```
Coverage on the new module: **94% lines / 83% branch** (well above the 75% changed-lines target).

**Integration** (`npm run test:integration`, real server + file storage + the
built-in demo/mock judge — no AWS creds needed):
```
tests/integration/server/routes/storage/evaluationRuns.retryJudgement.integration.test.ts
  ✓ returns 404 when the run does not exist
  ✓ returns 409 when the run is still running
  ✓ returns 409 when a second retry-judgement request arrives while the first is still in flight
  ✓ retries only the judge-failed case, leaves the passed case untouched, and recomputes stats
  ✓ does not retry a judge-failed case with no stored trajectory (agent crash — nothing to salvage)
  ✓ scope=all re-judges every rejudgeable case, including an already-passed one
Tests: 6 passed, 6 total

Full suite: Test Suites: 85 passed, 85 total
            Tests: 802 passed, 802 total
```

**E2E** (`npx playwright test tests/e2e/retry-judgement.spec.ts`, real Chromium against the built app):
```
✓ button renders with the judge-failed count and is enabled
✓ clicking opens a confirm dialog showing the count + judge model
✓ confirming POSTs to retry-judgement and shows a succeeded/failed summary
✓ button is visible but disabled when there are no judge-failed cases
4 passed (4.3s)
```
Existing `tests/e2e/rerun-evaluation-run.spec.ts` (sibling feature, same page) re-run clean: 10 passed, 1 pre-existing skip unrelated to this change.

**Build** (`npm run build:all`): tsc + vite + server + cli + lib all clean.

**Security** (`npm audit --audit-level=high`): only pre-existing transitive-dependency findings (react-router/undici/vite dev deps, same on `origin/main`) — `package-lock.json` is unchanged by this PR.

**DCO**: both commits signed off (`git log origin/main..HEAD | grep Signed-off-by`).

**CHANGELOG.md**: updated under `## [Unreleased]` → `### Added`.

goyamegh:
Follow-up pushed (`363da8a5f`): Retry judgement's button + confirm-dialog mount were keyed on route `mode === 'evalRun'` instead of `run.docType === 'evaluation-run'`, so they never rendered when a benchmark-linked run was viewed via the benchmark-scoped inspector route (`/evaluations/benchmarks/<id>/runs/<runId>/inspect`) even though the same run is reachable there — found live during independent verification against the shared cluster. Fixed the same way as the Re-run button's `docType`-keyed fix (same bug class), plus had the benchmark-route load path prefer the first-class `EvaluationRun` doc when one exists (so results/stats don't go stale after a retry there either). Unit + e2e regression coverage added, verified red pre-fix / green post-fix.

goyamegh:
## Fix for "failed to retry judgement" on large runs

**Root cause (confirmed, not assumed):** the owner's real incident run (`eval-run-1788238635378-evbqzpx9g`, 62 errored cases) showed the "failed to retry judgement" toast, but the server log had a burst of `[BedrockJudge] Pass/Fail` lines with **no completion log**, and the run's report docs showed **fresh judge verdicts/reasoning timestamps landing after the click and the toast** — the retry pipeline was genuinely still progressing server-side. `POST .../retry-judgement` was `await`ing the entire judge pipeline inline (62 cases × real Bedrock latency ÷ concurrency 3 ≈ 20-30+ minutes), which blows past any client fetch/proxy timeout.

**Fix:** async job pattern, mirroring the existing `/rerun` endpoint already in this file:
- `POST /api/storage/evaluation-runs/:id/retry-judgement` now responds **`202 { jobId, total, status: 'running' }`** immediately and runs the judge pipeline in the background.
- New `GET /api/storage/evaluation-runs/:id/retry-judgement/status` polls an in-memory job record (single-process, 1h TTL after completion) for `{ status, total, completed, summary?, error? }`.
- `retryJudgementForRun()` gained an `onProgress(completed, total)` hook (fired once up front at `0/N`, then after every case) and a new `countRetryableCases()` helper computes the pre-flight `total` with zero judge work.
- The client's `retryJudgement()` polls to completion internally (2s interval, 30min ceiling) — existing callers are source-compatible. `RetryJudgementConfirmDialog` shows live `N/total` progress instead of a bare spinner while it polls.

**Tests (red→green):**
- Unit — `onProgress` fires `[0,N]` up front and ends at `[N,N]` with a non-decreasing sequence; `countRetryableCases()` matches `selectRetryableCases().length` with zero judge calls.
- Integration — POST returns 202 + total, GET status polls to `completed` with the full summary (happy path, agent-crash-skip case, `scope=all`); double-submit guard still returns [202, 409]; unknown-job GET returns 404.
- E2e — mocks the 202 + status-poll network flow on the run inspector page and asserts the live progress row renders before the summary.

**Note on the double-submit guard test:** with a single trivial case + the demo/mock judge, the background job can complete in ~4ms — fast enough to occasionally finish before the second concurrent request is even dispatched by the test client (confirmed via timestamped debug logging), which is a test-timing artifact of the mock judge, not a guard bug — a real Bedrock call takes ~40-90s/case, so this window never gets remotely that tight in production. Widened the seeded run to 15 cases so the pipeline reliably outlasts both requests' dispatch; verified 8/8 clean runs plus the full `test:integration` suite green (apart from 2 pre-existing, unrelated environment-flaky tests — `portIsolation` and `sigv4CredentialRetry` — that spawn their own server processes and fail in isolation on this sandbox, touching neither file in this diff).

Build/test verification performed locally: `npm run build:all` clean (tsc, vite, esbuild ×2, tsc-alias), full `test:unit` (5284 tests) green, `test:integration` green for every retry-judgement case, `npm audit --audit-level=high` unchanged from base (pre-existing transitive-dep findings only, no new deps added).

goyamegh:
Heads up: #466 lands a proper `isEvaluationRun(run): r is EvaluationRun` predicate in `types/index.ts` for the exact same route-vs-doctype bug class this PR's Retry-judgement button/dialog hit (`(run as any).docType === 'evaluation-run'`), plus the `loadData()` fix that makes `docType` non-vacuous on the benchmark-scoped route. Once #466 merges, please rebase and swap this PR's `(run as any).docType` checks for `isEvaluationRun(run)` so we don't ship two competing patches for the same problem.

TackAdam (approved):

PR #459, opened 2026-09-01
Storage: lightweight run-list projections + on-demand report detail (fixes V8 string overflow, 36MB first-paint fetches)

ashwin-pc:
Complete transcripts made report documents honest — and heavy (36.7MB observed). Two endpoints buckled; this PR makes report size a non-issue everywhere:

**List endpoints** (`4a23c38`): previously loaded ALL report docs into memory pre-pagination (~1.1GB observed) then failed `res.json()` with V8 "Invalid string length" (HTTP 500 after 9s). Now: filename-first pagination, centralized lightweight projection (`server/adapters/runSummary.ts`) stripping rawEvents/trajectory/raw judge payloads across all seven list routes, `.summaries` sidecars with lazy backfill, OpenSearch `_source` projections. Measured: 9.1s/500 → 5ms/200.

**Detail endpoint** (`7fa14e4`): run-inspect pages eagerly fetched full docs per case (×3 ≈ 110MB before first paint). Now `GET /runs/:id?include=core|full` — core (268KB, everything except rawEvents/trajectory/raw responses) paints the page; heavy sections hydrate per-tab on demand. `full` preserves exact previous behavior for CLI/export consumers (callers audited). Measured: first paint from 8.2KB summaries + 92.7KB core.

Regression tests include a synthetic 50MB-rawEvents report: lists stay small, detail stays complete, nothing serializes unbounded documents by accident.

PR #463, opened 2026-09-01
fix(api): contract hardening across storage/metrics/skills routes

goyamegh:
## Summary

Fixes 9 remaining findings from an API KPI probe sweep (F4, F5, F6, F7, F8, F9, F10, F11, F12), plus carries the already-tested storage create/bulk validation fix (F1-F3, cherry-picked from `goyamegh/storage-create-validation`, commit 3b08dccb — this was cherry-picked clean, no conflicts against post-#444 `main`). F1 (pagination) is deliberately **not** touched here — covered by merged #444 + open #459.

Each finding below maps to one or two commits (see commit list). All source commits carry unit tests; a consolidated integration suite (real server + real shared OpenSearch cluster) covers the HTTP-visible findings in the final commit.

---

### F11 (P1) — unknown `agentKey` started a real 30s+ run and persisted junk

**Before:** `POST /api/storage/benchmarks/:id/execute` with `{ agentKey: "bogus" }` returned `200` SSE, started `executeRun()`, and wrote a run entry into `benchmark.runs[]` — all before the connector eventually failed.

**After:** `agentKey` is resolved against the exact same source `GET /api/agents` uses (`loadConfigSync().agents` + `getCustomAgents()`) **before** the benchmark is even fetched. Unknown key → fast `400 { error: "Unknown agentKey: ..." }`, nothing persisted.

**Test (integration, real cluster):**
```
POST /api/storage/benchmarks/<id>/execute { agentKey: "nonexistent-agent-..." }
→ 400 { error: "Unknown agentKey: nonexistent-agent-..." }
GET  /api/storage/benchmarks/<id> → runs[] unchanged (length before === length after)
```

### F4 (P2) — malformed JSON leaked an HTML stack trace

**Before:** `POST` with body `{not json` → Express's default HTML error page, including the raw `SyntaxError` stack (internal file paths).

**After:** body-parser `entity.parse.failed` → clean `400 { "error": "invalid JSON body" }`; a final catch-all error handler (registered after routes + SPA fallback) ensures any other uncaught error also answers JSON, never HTML/stack.

**Test (integration, 3 real routes):**
```
POST /api/storage/test-cases '{not json' → 400 { error: "invalid JSON body" }, content-type: json, no "<html", no "SyntaxError"
POST /api/storage/benchmarks  '{not json' → same
POST /api/logs                '{not json' → same
```

### F12 (P2) — `PATCH .../metadata` persisted type-confused values

**Before:** `PATCH .../metadata { name: 12345 }` → `200`, benchmark's `name` field stored and returned as the **number** `12345`.

**After:** `name` (non-empty string when present) and `description` (string when present) are validated; violations → `400` before the storage adapter is called.

**Test (integration, real cluster):**
```
PATCH .../metadata { name: 12345 } → 400
GET   .../<id> → name is still the original string (unchanged)
PATCH .../metadata { description: { nested: true } } → 400
PATCH .../metadata { name: "Renamed" } → 200, name: "Renamed" (valid updates still work)
```

### F5 (P2) — dead, unmounted leaderboard routes

**Judgement call:** `server/routes/leaderboard.ts` (11 `/api/coding-agents/leaderboard/*` endpoints) is never imported/mounted by `server/routes/index.ts` — every one of these routes has always 404'd. Grepped the whole repo (frontend, CLI, tests) for callers: **none exist**. Deleted the file rather than wiring it up, since (a) nothing calls it, (b) its backing `server/services/codingAgents/leaderboard/*` service modules are themselves unreferenced elsewhere, and (c) wiring up 11 endpoints nobody has exercised is riskier than removing dead code that behaves identically either way (404 before, 404 after). **Flagging for reviewers** in case there was unshipped intent here — happy to revert if maintainers want it wired up instead.

**Test (integration):** the 3 sampled leaderboard endpoints still return `404` post-removal (regression-lock on the unchanged observable behavior).

### F6 (P2) — skills folder-picker leaked the shell command on failure

**Before:** `POST /api/skills/browse` on a headless Linux box with neither `zenity` nor `kdialog` installed → `500` with the full failed shell command (`Command failed: zenity --file-selection ... /bin/sh: zenity: command not found`) in the response body.

**Checked the frontend caller** (`services/client/skillsApi.ts` → `components/skills/SkillsPage.tsx`): the Skills page already has a manual-path text-entry fallback, so a clean error message is sufficient.

**After:** `501 { error: "Native folder picker unavailable on this server. Enter the skill folder path manually." }`; the real error is still logged server-side (`debug` + `console.warn`) for operators. Cancel/success paths unchanged.

**Test (unit, mocked `execSync`):** asserts the response body never contains `zenity`/`kdialog`/`Command failed` substrings.

### F7 (P3) — dishonest/inconsistent delete-of-nonexistent semantics

**Before:**
- `DELETE /api/storage/test-cases/:nonexistent` → `200 { deleted: 0 }`
- `DELETE /api/storage/benchmarks/:nonexistent` → `200 { deleted: true }` (a lie — nothing was deleted)
- `GET` on the same ids correctly `404`s in both cases.

**After:** both routes now check the adapter's already-accurate `{ deleted }` return value (`boolean` for benchmarks, a version-count for test cases — both adapters already return this correctly, the routes just ignored it) and answer `404` on a no-op delete. **Implemented entirely at the route layer** — no adapter changes, per the conflict-avoidance note (open PR #452 owns adapter delete internals).

**Test (integration, real cluster):**
```
DELETE /api/storage/test-cases/<nonexistent> → 404
DELETE /api/storage/benchmarks/<nonexistent> → 404
DELETE /api/storage/benchmarks/<real-id>     → 200 { deleted: true } (unchanged for real deletes)
```

### F8 (P3) — nonexistent run indistinguishable from a real zero-cost run

**Before:** `GET /api/metrics/:runId` for a nonexistent runId → `200` with all-zero metrics and `status: "pending"` — the exact same shape a real zero-cost run (e.g. tool-only, no LLM calls) would produce.

**Checked the frontend consumer** (`services/metrics.ts` → run-report polling, which relies on `status: "pending"` + `200` for legitimate in-flight polling): changing this to a `404` would break real in-flight-run polling, so I picked the **non-breaking option** explicitly suggested by the finding.

**After:** added an additive `found: boolean` field — `false` when no spans exist for the runId, `true` whenever spans exist (even at zero cost). Existing status codes/shape are unchanged.

**Test (unit + integration):** `found:false` for empty spans, `found:true` for a zero-cost span set; integration hits a real nonexistent runId against the live observability cluster.

### F9 (P3) — unscoped query on empty body

**Before:** `POST /api/logs` with `{}` → `200` with an unscoped, cluster-wide `match_all`-equivalent query (`must: []`).

**Checked the real callers** (`server/services/traceJudgeTools.ts`, `server/services/comparisonTraceTools.ts` — the judge/comparison `query_logs` tools): both already guard "no runId available" before calling this route and always send a `runId`. An empty body has no legitimate meaning here.

**After:** requires the body to be a JSON object and `runId` to be a non-empty string; `400` otherwise, before the query is built.

**Test (unit + integration):** empty body, non-string `runId`, non-object body all `400`; a real `runId` still queries scoped.

### F10 (P3) — boot log doesn't distinguish "disabled" from "scanned, found nothing"

**Before:** `recoverOrphanBenchmarkRunsSafely()` / `resumePendingTracePollsSafely()` print the identical `no orphan running/pending ... [0ms]` line whether recovery actually scanned and found zero orphans, or was short-circuited by `BENCHMARK_RUN_RECOVERY_DISABLED=1` / `TRACE_RECOVERY_DISABLED=1` — this repo's own test servers boot with these env vars set (per this repo's `AGENTS.md`), so the ambiguity is hit on every local/CI test-server boot.

**Judgement call:** the task description also names `EVALUATION_RUN_RECOVERY_DISABLED` as a sibling env var, but it doesn't actually gate anything in this codebase (no `evaluationRunRecoveryOnBoot.ts` module exists) — test servers set it defensively/future-proofingly. The real sibling with the identical bug is `server/services/traceRecoveryOnBoot.ts` (gated by `TRACE_RECOVERY_DISABLED`), fixed alongside `benchmarkRunRecoveryOnBoot.ts`.

**After:** both stat shapes carry a `disabled` flag, set on the disabled short-circuit path; the boot log now prints a distinct `recovery disabled via env (...)` line.

**Test (unit):** asserts the distinct log line and correct `disabled` flag in both the "disabled" and "scanned, found nothing" cases.

---

## Verification

**Build:** `npm run build:all` — clean (frontend `tsc && vite build`, `build:server`, `build:cli`, `build:lib` all succeed).

**Unit tests:** `npx jest tests/unit` → **309/309 suites, 5259/5264 tests passing** (5 skipped, pre-existing/unrelated — same skip set as `origin/main`).

**Integration tests** (real server, real shared OpenSearch cluster, `AH_PORT=4342`, `BENCHMARK_RUN_RECOVERY_DISABLED=1 EVALUATION_RUN_RECOVERY_DISABLED=1 TRACE_RECOVERY_DISABLED=1`):
- `createValidation.integration.test.ts` (the cherry-picked F1-F3 fix's own suite): **23/23 passing**
- `apiContractHardening.integration.test.ts` (this PR's new suite, F4/F5/F7/F8/F9/F11/F12): **15/15 passing**

```
Tests:       15 passed, 15 total
  F4: malformed JSON body handling (3/3)
  F11: unknown agentKey execute (1/1)
  F12: metadata type validation (3/3)
  F7: delete-of-nonexistent semantics (3/3)
  F8: metrics found:false (1/1)
  F9: logs runId validation (1/1)
  F5: removed leaderboard routes stay 404 (3/3)
```

All test-created data was tracked by id and deleted in `afterAll` (never a name-scan delete); verified zero leaked documents on the shared cluster via a post-run `includeSample=false` scan for the suite's unique name marker.

**`npm audit --audit-level=high`:** 31 pre-existing vulnerabilities (react-router, undici, vite transitive deps) — these are **pre-existing on `origin/main`**, unrelated to this change, and are being tracked/repaired by a parallel effort per the standing note in this repo's workflow docs.

**e2e:** not run for this PR — every change here is server/API-only; no UI-visible behavior changed (confirmed via full repo grep for frontend callers of every touched route).

**DCO:** all 9 commits signed off (`git log origin/main..HEAD | grep -c "Signed-off-by"` → 9/9).

**CHANGELOG.md:** updated under `## [Unreleased]` → `### Fixed` with one entry per finding, each linking the touched files and tests.

---

## Codex adversarial review

Findings and applied/rejected decisions will be posted as a follow-up comment on this PR before requesting approval, per this repo owner's standing review process.


goyamegh:
## Adversarial review (codex_review / gpt-5.4)

Per this repo owner's standing practice, I ran an adversarial review of this diff with a different model family (OpenAI gpt-5.4) before requesting approval. Findings and what I did with each:

|---|---|---|---|
| 1 | `agentName`/`modelName` relaxation on `POST /api/storage/runs` only checked `typeof === 'string'`, so `""` passed | HIGH | **Applied** — now requires a non-empty string when the field is present (still optional). Added unit tests for empty-string `agentName` and whitespace-only `modelName`. |
| 2 | `POST /api/logs`'s `runId` check (`!runId \|\| typeof !== 'string'`) let whitespace-only strings through | MED (raised while reviewing #1's pattern) | **Applied** — added a `.trim()` check. Added a unit test for `runId: '   '`. |
| 3 | The JSON body-parser error handler only mapped `entity.parse.failed` (malformed JSON); an oversized body (`entity.too.large`) wasn't given the same clean-JSON treatment and would fall through to the generic 500 handler | MED | **Applied** — added an `entity.too.large` → `413` branch alongside the existing `400`. Added a unit test. |
| 4 | `found:false` in the metrics response only reflects "no trace spans were found for this runId", not literal run-existence (ingestion lag / wrong time window would also show `found:false`) | HIGH | **Acknowledged, not changed** — this is an inherent limitation of a trace-based metrics endpoint that has no run-document lookup to consult (by design, to avoid the false-positive risk of adding one — see the F8 write-up in the PR description for why I didn't wire in a storage lookup). Tightened the code comment to state this limitation explicitly rather than implying certainty. |
| 5 | `isKnownAgentKey()` on benchmark execute reads current config, so a benchmark referencing a since-renamed/removed agent key becomes unexecutable | HIGH | **Rejected** — this isn't a regression: before this fix, executing with an unknown/renamed-away agentKey *also* failed, just later (after starting a real run and persisting a junk entry) rather than fast. Nothing that used to succeed now fails. `loadConfigSync()` per-request is the same pattern `GET /api/agents` already uses synchronously today. |
| 6 | Deleting `server/routes/leaderboard.ts` (11 dead endpoints) is scope creep for a "bug fix" PR and should have maintainer sign-off | MED | **Rejected (already flagged)** — this was explicitly one of the 12 findings this PR was asked to fix (F5); the PR body already calls it out as a judgement call and invites reviewers to object/revert. |
| 7 | `testCases.ts` delete-of-nonexistent check (`if (!result.deleted)`) assumes the adapter's version-count is 0 iff nothing existed, which the finding's own writeup describes as a "version count", not a boolean | MED | **Reviewed, rejected** — every test case is created with an initial `version: 1` document; a real test case with zero stored versions is not a reachable state in either adapter, so `deleted === 0` reliably means "nothing existed." |
| 8 | Skills folder-picker fix still logs the sensitive `err.message` server-side via `debug()`/`console.warn()` | LOW | **Rejected (by design)** — the finding was specifically about the *client-facing response body* leaking shell command text; server-side operator logs are a different, expected trust boundary, and the assigned task explicitly asked for exactly this ("logging the real error server-side for debugging"). |
| 9 | The single consolidated integration suite against a real shared cluster is less rigorous than targeted tests for the new edge cases | MED | **Partially applied** — added the missing unit-test edge cases (see #1/#2/#3 above) rather than adding more integration surface against the shared cluster, which would add flakiness without adding signal for these specific checks. |

Net: 3 real hardening gaps closed (empty-string `agentName`/`modelName`, whitespace `runId`, `entity.too.large`), 1 doc clarification, 5 findings reviewed and rejected with reasoning above.

goyamegh:
## Update: CI fixes + additional codex_review follow-up applied

New head: `7fb647095729c0de56fe84f6084d780767c18ca9`

### CI fixes
- **`link-check`**: `CHANGELOG.md`'s F452 e2e-cleanup bullet referenced `tests/e2e/resume-run-benchmark-link.spec.ts`, which has never existed on `origin/main` — this exact regression was already fixed once before (commit `1f5d5fd58`) but PR #452's squash-merge carried the pre-fix wording back. Applied the identical fix (describe only the three real specs). Verified no other `.md` link in the repo points at a nonexistent local file.
- **`integration-tests`**: one flaky failure on `evaluationRuns.integration.test.ts`'s re-promote check (expected 400, got 200). Traced the full path (`promoteRunToBenchmark()` → `refresh:'wait_for'` write → real-time doc `GET` read) — there's no plausible storage-layer race for this specific check, and neither `evaluationRuns.ts` nor `benchmarkPromotion.ts` is touched by any commit in this PR. Hardened the test with a short bounded poll before the re-promote call regardless, since the root cause (heavy concurrent CI load) couldn't be pinned down further and the mitigation is free in the normal case.

### Additional codex_review pass (per follow-up request)
|---|---|---|---|
| 1 | Bulk routes (`runs/bulk`, `benchmarks/bulk`) silently folded validation-dropped items into `errors` | HIGH | **Applied** — added `invalid`/`invalidIndexes` to both responses (additive; `errors` keeps its meaning). PR body's bulk section now documents the two failure models explicitly: `test-cases/bulk` rejects the whole batch (bulkUpsert 500-risk), `benchmarks/bulk`/`runs/bulk` are partial-success with reported indexes. |
| 2 | `found` on metrics conflated "run missing" with "no spans yet" | HIGH | **Applied** — renamed to `hasSpans` (types + service + tests + CHANGELOG), doc comment now states the limitation explicitly. |
| 3 | Verify `isKnownAgentKey()`'s resolution source matches `executeRun()`'s | HIGH | **Investigated and confirmed**: `services/benchmarkRunner.ts`'s `buildAgentConfigForRun()` uses the exact same `[...config.agents, ...getCustomAgents()].find(...)` pattern, no remote-server/alias resolution elsewhere. Documented this directly in the function's doc comment, and closed a latent gap found while verifying: `isKnownAgentKey()` didn't defend against `loadConfigSync()` throwing (an uncaught rejection outside an async Express 4 handler's `try` doesn't reach the error middleware) — now fails closed with try/catch, mirroring `buildAgentConfigForRun()`'s own defensive pattern. |
| 4 | `validateBenchmarkCreate`'s `testCaseIds` allowed `''` | MED | **Applied** — now requires non-empty strings. |
| 5 | `POST /api/logs` had no bound on `size`, no type-check on `startTime`/`endTime` | LOW | **Applied** — `size` clamped to 1-1000 (default 100), `startTime`/`endTime` rejected if present and non-numeric. |

Rejected items (per your note, logged here for the record): 404-on-nonexistent-delete stays; SSE mid-stream error semantics on `/execute` stays out of scope (separate finding); `501` on `skills/browse` stays; final-error-handler message left as `'Internal server error'` (not changed to `'Bad request'` for <500 statuses — happy to change if preferred, deferred to owner call).

All unit (330 suites / 5634 tests) and integration tests (58/58 across the 3 touched suites) verified green locally against the real shared OpenSearch cluster before this push.

TackAdam (approved):

PR #464, opened 2026-09-01
fix(evals3): run inspector falls back to standalone eval-run doc when missing from benchmark.runs[]

goyamegh:
## Summary

A Claude Code (or any run-first/CLI-started) evaluation run tied to a benchmark
was **not clickable** on the benchmark runs page while `Running` — clicking the
row silently did nothing, no error shown.

## Root cause

`benchmark.runs[]` is only appended to when a run-first evaluation-run document
reaches a **terminal** state (`linkCompletedRunToBenchmark` / the create route's
success-path linking in `server/routes/storage/evaluationRuns.ts` only fires on
completion). A still-`running` evaluation-run tied to a benchmark therefore has
**no entry** in `benchmark.runs[]` yet, even though it exists as a standalone
document (`GET /api/storage/evaluation-runs/:id` already serves it fine).

`components/evals3/RunInspectorPage.tsx`'s benchmark-mode loader looked **only**
inside `benchmark.runs[]` and, on a miss, silently `navigate()`d straight back
to the runs list — from the user's perspective, landing on (or clicking into)
that run's inspect URL did nothing, with no visible error (same "hide, don't
explain" anti-pattern as #449).

## Fix

- Falls back to `GET /api/storage/evaluation-runs/:id` (the same fetch
  eval-run mode already uses) when the run is absent from `benchmark.runs[]`.
- A run missing from **both** stores now renders an explicit "Run not found"
  state (`data-testid="run-inspector-not-found"`) with a reason and a
  "Back to runs" action — never a silent redirect.

## Note on reachability today vs. after this lands

The exact live symptom (a visible-but-dead row) requires a benchmark runs-list
page that already unions `benchmark.runs[]` with standalone evaluation-run
docs — that union currently only exists on `main-goyamegh` (fork), not yet on
`origin/main`. This fix targets the shared root cause
(`RunInspectorPage.tsx`'s lookup) directly and is fully testable/valid on
`origin/main` today (e.g. via a direct URL, a "Compare" deep link, or once the
list-union feature lands) — see the e2e test's docstring for detail.

## Tests

- **Unit** — `tests/unit/components/evals3/RunInspectorPage.test.ts`: fallback
  fetch renders the run; genuine not-found-anywhere renders the explicit state
  with no navigate call (regression guard against the old silent bounce).
- **Integration** — `tests/integration/server/routes/storage/benchmarkRunRowLink.integration.test.ts`:
  real Express route handlers exercise the `benchmark.runs[]` precondition,
  the fallback `GET .../evaluation-runs/:id`, the benchmarkId-scoped list
  union, and 404 for a genuinely missing run.
- **E2e** — `tests/e2e/benchmark-run-row-link.spec.ts`: real browser navigation
  (via `./fixtures/test-fixtures`) to a not-yet-linked run's inspect URL
  renders the run instead of bouncing; a nonexistent run renders the explicit
  not-found state with a working "Back to runs" action.

## Verification

- `npm run build:all` — clean (tsc + vite + server + cli + lib).
- `npm run test:unit` — 5,319 passed (0 unrelated failures after rebuilding
  the CLI artifact once).
- Targeted integration/e2e runs for this change: all green (see PR comment
  for terminal output).
- `npm audit --audit-level=high` — pre-existing findings only (undici via
  `@earendil-works/pi-coding-agent`, vite/launch-editor Windows-only); no new
  dependencies touched by this PR.
- DCO signoff present on the commit.
- `CHANGELOG.md` updated under `## [Unreleased]` → `### Fixed`.


goyamegh:
## codex_review (gpt-5.4) — findings and disposition

Ran an adversarial review against this diff before/after the fix. Applied all HIGH findings; the MED/LOW findings were evaluated and mostly already covered, with reasoning below.

| 1 | HIGH | Fallback fetch didn't verify the returned `EvaluationRun` is actually associated with `:benchmarkId` — could render a run from a different benchmark (or none) under this benchmark's URL. | **Applied.** Added an explicit association check (`benchmarkId` field match, or a `sources` entry of `type: 'benchmark'` with matching `benchmarkId`); rejects with the same "not found" state otherwise. Unit test added: cross-benchmark run is rejected, never rendered. |
| 2 | HIGH | The `try/catch` around the fallback fetch collapsed **all** failures (network error, 500, etc.) into "Run not found", masking real failures. | **Applied.** Fallback now does a direct `fetch()` (not the shared `getEvaluationRun()` helper, which doesn't expose status codes) so a genuine `404` is the only thing that renders "not found"; anything else re-throws into the existing outer `loadError`/Retry state. Unit test added: a 500 renders the pre-existing error+Retry UI, not "not found". |
| 3 | HIGH | Rendering a standalone `EvaluationRun` under benchmark mode could leak eval-run-only UI affordances (e.g. Re-run) if any UI gating keyed off data shape instead of `mode`. | **Verified, no code change needed.** All benchmark-mode gating in this file keys off the `mode` const (derived from the route/`benchmarkId`), not the data shape — added a unit assertion proving the Re-run button stays disabled with the benchmark-mode tooltip even when the underlying data came from the fallback fetch. |
| 4 | MED | `notFoundReason` state could go stale across route changes (e.g. navigating from an invalid run to a valid one within the same mount). | **Already handled, not a gap.** `setNotFoundReason(null)` runs at the top of every `loadData()` call, and `loadData` is already keyed by `[benchmarkId, runId, mode, ...]` in its `useCallback`/`useEffect`, so a route change always clears it before the next fetch. No test gap found on inspection; left as-is. |
| 5 | MED | The e2e test forces a permanently-failed (never-embedded) run via an unresolvable `agentKey`, rather than a genuinely `running` run — a different terminal state than the live report. | **Accepted tradeoff, documented.** A real `running` window is racy to hit deterministically in CI; the code path under test (the `bm.runs[]` lookup miss + fallback) doesn't branch on run status, and the unit-level fixture (`makeStandaloneEvalRun()`) already defaults to `status: 'running'` to cover that shape directly. Noted in the e2e spec's docstring. |
| 6 | MED | The exact user-visible symptom (dead list row) requires a runs-list union feature that isn't upstream yet on `origin/main` — is this PR "ahead of" its own justification? | **Acknowledged, framing kept.** This fixes the shared root cause in `RunInspectorPage.tsx`'s lookup directly, independent of which caller reaches the URL (direct link, "Compare" deep link, or the list union once it lands) — documented explicitly in the PR description and the e2e docstring rather than glossed over. |
| 7 | LOW | The explicit not-found screen may be scope creep beyond "make the fallback work". | **Kept, by design.** Explicitly required by the task brief ("never a silently dead row") — replacing one silent failure mode (bounce to list) with another (generic error) would repeat the same anti-pattern this fix exists to remove. |
| 8 | LOW | Integration tests (mocked storage) don't protect the highest-risk client logic (association check, 404-vs-other). | **Addressed via unit tests instead** (see #1/#2) — that logic is pure client-side branching, which unit tests exercise directly and deterministically; the integration tests remain scoped to proving the real route contracts (`benchmark.runs[]` shape, the fallback GET, the list union) the client relies on. |

Full codex transcript available on request; happy to paste the raw output if useful for review.

goyamegh:
## Verification proof

**Build (`npm run build:all`):** clean — `tsc` typecheck, `vite build`, `esbuild` server/cli/lib all succeed.

**Unit — full suite (`npm run test:unit`):**
```
Test Suites: 313 passed, 313 total
Tests:       5 skipped, 5312 passed, 5317 total
Time:        40.453 s
```

**Unit — this change's suite (after the codex_review revision):**
```
RunInspectorPage — benchmark-mode fallback for not-yet-linked runs
  ✓ falls back to the standalone evaluation-run document when the run is missing from benchmark.runs[], instead of bouncing to the runs list
  ✓ rejects a standalone run that exists but is NOT associated with this benchmark (cross-benchmark data must never render) — codex_review finding
  ✓ treats a non-404 fallback failure (transient 500/network error) as a load error with Retry, NOT as "not found" — codex_review finding
  ✓ renders an explicit not-found state (not a silent bounce) when the run is missing from BOTH benchmark.runs[] and the standalone evaluation-run store
RunInspectorPage — Re-run button (benchmark mode)
  ✓ disables Re-run button for benchmark-embedded runs

Tests: 20 passed, 20 total
```

**Integration (`tests/integration/server/routes/storage/benchmarkRunRowLink.integration.test.ts`, real Express route handlers):**
```
Benchmark run row link — mid-run linkability (real routes, mocked storage)
  ✓ confirms the precondition: benchmark.runs[] does not contain the still-running run
  ✓ serves the standalone evaluation-run doc via GET .../evaluation-runs/:id — the fallback RunInspectorPage.tsx now uses
  ✓ serves the run via the benchmarkId-scoped list — the union query the runs LIST page already relied on
  ✓ returns 404 (not a hang or 500) for a run id that exists in neither store — the genuine not-found case the inspector page now covers with an explicit error state

Tests: 4 passed, 4 total
```

**E2e (`tests/e2e/benchmark-run-row-link.spec.ts`, real Chromium + real server + real OpenSearch-backed storage):**
```
Running 2 tests using 2 workers
  ✓ run inspector for a genuinely nonexistent run renders an explicit not-found state, not a silent bounce (2.1s)
  ✓ run inspector for a not-yet-linked evaluation-run renders the run (not a silent bounce back to the runs list) when navigated to directly (2.2s)

2 passed (1.1m)
```

**`npm audit --audit-level=high`:** pre-existing findings only (undici via `@earendil-works/pi-coding-agent`, vite/launch-editor Windows-only) — no dependency changes in this PR.

**DCO:** signoff present on the commit.

Test data created during verification was tracked and deleted via `TestDataTracker`/the Playwright `testData`-fixture-equivalent pattern (id-scoped, no name/prefix sweeps against shared storage).

TackAdam (approved):

PR #465, opened 2026-09-01
feat(runs): inline rename + newest-first sorting for evaluation runs

goyamegh:
## Summary

Owner ask (verbatim): "I also want to be able to rename the evaluation run name, and the sorting should be by time, not by name of the benchmark then time."

### Rename
- `PATCH /api/storage/evaluation-runs/:id` already existed but accepted `name` unvalidated. It now validates via `validateRunNameUpdate()` (`lib/runName.ts`): trims, rejects empty/non-string/>200-char names with 400, touches only `name` (no version bump, no stats change).
- New shared `InlineRenameField` component (pencil -> click -> input -> Enter/blur save, Escape cancel, inline error on failure -- there is no toast system in this repo) wired into both `RunInspectorPage`'s header title and `EvalRunsPage`'s run-row name cell, with optimistic update + revert-on-failure.
- Rename only applies to top-level `EvaluationRun` docs (eval-run mode / `kind: eval-run` rows) -- legacy benchmark-embedded runs have no PATCH-able endpoint, same gating as the existing Re-run button.

### Sorting
- The storage-layer default (`OpenSearchEvaluationRunOperations.list()`) already sorts `createdAt desc`, and `EvalRunsPage`'s flat view already defaulted to `timestamp desc`.
- The actual bug was the **grouped** view: `groupedByBenchmark` sorted groups alphabetically by benchmark name -- exactly "benchmark name, then time." Extracted `sortGroupsByRecency()` (`lib/runSort.ts`) so groups order by their most-recent run, descending. Flat/Grouped toggle + its persisted preference are unchanged.

### Tests
- Unit: `tests/unit/lib/runName.test.ts`, `tests/unit/lib/runSort.test.ts`, extended `tests/unit/server/routes/storage/evaluationRuns.test.ts`, `tests/unit/components/evals3/EvalRunsPage.test.ts`, `tests/unit/components/evals3/RunInspectorPage.test.ts`.
- Integration (real storage API): `tests/integration/server/routes/storage/evaluationRunRename.integration.test.ts` -- PATCH persists a trimmed rename without touching stats/status, rejects empty/over-cap names, list endpoint returns newest-first.
- E2E: `tests/e2e/evalrun-rename-and-sort.spec.ts` (imports `./fixtures/test-fixtures`) -- rename via the inspector header and the run-row both persist across a hard reload; the runs list renders newest-first by default even when alphabetical order would put an older run first.

All new/changed code paths are covered at unit, integration, and e2e levels per the repo's testing conventions. `npm run build:all` and the full unit suite (332 suites / 5620 tests) pass; integration suites relevant to this change pass; e2e spec passes under both serial and parallel execution.

### CHANGELOG
Updated under `## [Unreleased]`.


goyamegh:
## Verification

**Typecheck** (`tsc --noEmit`): clean for this PR's files. The repo's shared dev-node_modules is missing an installed `prismjs` runtime for two unrelated files (`ContextValueView.tsx`, `EvalSourceCodeView.tsx`) -- confirmed pre-existing on `origin/main` (reproduced via `git stash`), unrelated to this change.

**Targeted unit suites** (all passing):
```
Test Suites: 6 passed, 6 total
Tests:       105 passed, 105 total
```
Covers `lib/runName.test.ts`, `lib/runSort.test.ts`, the extended `evaluationRuns.test.ts` PATCH validation cases, `EvalRunsPage.test.ts` (rename + grouped-sort-by-recency), `RunInspectorPage.test.ts` (rename), and `evaluationRunsApi.integration.test.ts` (client error-message passthrough regression test).

**Integration** (real OpenSearch-backed server, `AH_PORT=4941`, `BENCHMARK_RUN_RECOVERY_DISABLED=1 EVALUATION_RUN_RECOVERY_DISABLED=1`):
```
PASS tests/integration/server/routes/storage/evaluationRunRename.integration.test.ts
  PATCH /api/storage/evaluation-runs/:id -- rename
    OK persists a trimmed rename and does not touch stats/status/other fields
    OK rejects an empty name with 400 and leaves the stored name unchanged
    OK rejects a name over the 200-character cap with 400
    OK returns 404 for a non-existent run id
  GET /api/storage/evaluation-runs -- default order
    OK returns runs newest-first by default (createdAt desc at the storage layer)
Tests: 5 passed, 5 total
```
All fixtures created via `uniqueTestName()` + `TestDataTracker`; verified 0 leaked docs on the shared cluster after the run.

**E2E** (real browser + real dev server + real backend, `tests/e2e/evalrun-rename-and-sort.spec.ts`, imports `./fixtures/test-fixtures`):
```
OK renaming a run from the inspector header persists across reload
OK renaming a run from the Evaluation Runs list row persists across reload
OK Evaluation Runs list defaults to newest-first, not alphabetical-by-benchmark
3 passed
```
Passed under both `--workers=1` and `--workers=2`.

**Full suite** (unrelated to this change): `npm run build:all` is green; the full unit suite is green (332 suites / 5620 tests, unchanged by this PR). A handful of unrelated integration suites (bulk test-case upsert, CLI SDK dedup, benchmark judge execute) timed out when run in the SAME full-suite pass against a single dev server instance under many concurrent Jest workers -- confirmed pre-existing contention/flakiness by re-running each in isolation (`--runInBand`), where all pass cleanly. None of the failing suites touch this PR's files.

`npm audit --audit-level=high`: pre-existing findings only (undici via `@earendil-works/pi-coding-agent`, vite/launch-editor Windows-only) -- no `package.json`/`package-lock.json` changes in this PR.

## Second-opinion review (codex_review, gpt-5.4)

Ran an adversarial review against the actual diff, then again after a follow-up commit addressing findings. Findings and disposition:

|---|---|---|---|
| 1 | `updateEvaluationRun()` client discarded the PATCH route's `{error}` body, throwing a generic `statusText` -- the new server-side validation message ("name must not be empty", etc.) was invisible in the rename UI | MED | **Applied** -- now parses/surfaces the real message, matching every other mutating call in this client module; added a regression test |
| 2 | `InlineRenameField`'s `commit()` could theoretically double-fire on Enter+blur without a synchronous in-flight guard | HIGH (flagged) | **Already mitigated** in the diff under review -- a `committingRef` (ref, not state) guard checked+set synchronously before the async gap was already present; re-verified it closes the exact race described |
| 3 | Raw HTML `maxLength` on the rename input matched the server's cap on the *trimmed* value, so a name padded with leading/trailing whitespace could be blocked from being typed even though it would validate after trim | LOW | **Applied** -- buffered the raw input's `maxLength` above the trimmed cap |
| 4 | `lib/runSort.ts` exported `sortRunsByRecency()` (+ tests) with no production call site | LOW | **Applied** -- removed the unused export and its tests; kept only `mostRecentTime`/`sortGroupsByRecency`, which do have a real call site |
| 5 | Doc comment on `lib/runName.ts` claimed it was "shared...with any client-side pre-check", but no client-side pre-check actually calls it | LOW | **Applied** -- corrected the comment to describe reality (server-route-only) |
| 6 | Non-strict-TS discriminated-union narrowing gotcha (`!validated.ok` silently fails to narrow under this repo's tsconfig, which has neither `strict` nor `strictNullChecks`; `validated.ok === false` is required) flagged as a sharp edge for future contributors | LOW (informational) | **Documented** -- added an explicit code comment at the one call site explaining why, for future discriminated-union checks in this route tree |
| 7 | Rename/rerun eligibility gating (`mode === 'evalRun'` vs `rr.kind === 'eval-run'`) is duplicated per-file instead of a shared helper | LOW | **Not applied** -- this matches the pre-existing Re-run button's gating convention in the same files (not a new inconsistency this PR introduces); introducing a new shared helper for one boolean check seemed like the over-abstraction the review was otherwise warning against |
| 8 | `sortGroupsByRecency<T>`'s generic signature is more machinery than its single call site needs | LOW | **Not applied** -- extracting it to a plain, non-React, unit-testable function was an explicit requirement of this task's test coverage bar ("sort comparators incl. grouped-by-recency" as an isolated unit) |

No correctness issue was found in: the optimistic-update trim consistency (both save paths send the already-trimmed value, so optimistic state matches persisted state on success), `mostRecentTime()` returning 0 for empty groups (the production grouped-view path never constructs empty groups), or the `RunNameValidation` discriminated-union type itself.

goyamegh:
Converted to draft while a rename-affordance gap found in live testing is fixed (pencil not rendering for benchmark-embedded run entries). Un-drafts after the fix + re-verification; the sort half is confirmed working.

TackAdam (approved):

PR #467, opened 2026-09-02
fix(evals3): test-case detail page — pending runs show as false 'Failed' and never auto-refresh

goyamegh:
## Summary

Fixes the reported bug: after pressing **Run Test** twice on a test-case detail page (agent-trace-judge / post-#461-deploy era), the new runs appeared not to show up.

## Diagnosis

- **Creation-side: not the bug.** Both runs the owner triggered were persisted correctly (`report-1788240114417-c64t09j9t`, `report-1788238638479-r2mygj14z`), and 12 older runs for the same test case (`tc-1788156435231-s7rwc2ud0`) were present too — verified via `POST /api/storage/runs/search {testCaseId}` and `GET /api/storage/runs/by-test-case/:id` against the live cluster.
- **Query-side: not the bug.** `GET /api/storage/runs/by-test-case/:id` (the exact endpoint `TestCaseDetailPage` calls) returns the pending run immediately (`refresh: 'wait_for'` on every write) and reflects the judged verdict right after the update — see the new integration test.
- **Real bug: UI staleness + misclassification.** `POST /api/evaluate` (UI mode) uses `awaitTraces: false` for trace-mode agents, so it sends its SSE `completed` event with `metricsStatus: 'pending'` *before* the background trace-judge finishes (this is an existing, intentional contract — see `tests/integration/services/evaluation/traceBlocking.integration.test.ts`). `TestCaseDetailPage.tsx`:
  1. classified any run that wasn't `passFailStatus === 'passed'` as **"Failed"** (red `XCircle`) in the runs-list row — so a fresh, not-yet-judged run looked exactly like a real failure, and
  2. called `loadData()` exactly once, right after that SSE event, with no mechanism to ever refetch — so once the background judge resolved the run (seconds to ~1 minute later), the page never picked up the real verdict without a manual reload.

## Fix

`components/evals3/TestCaseDetailPage.tsx`:
- Row icon now uses the existing `getStatus()` helper (already correct for the right-hand inspector panel) instead of its own simplified `isErrored`/`isPassed` check, so pending/running rows show a spinner (`title="Judging…"`) instead of "Failed".
- Poll every 5s while any run is `pending`/`running`, stopping once nothing is left pending — mirrors the identical, already-shipped pattern in `components/RunDetailsPage.tsx`.

## Tests

- **Unit** (`tests/unit/components/evals3/TestCaseDetailPage.test.ts`): pending row renders the spinner, not "Failed"; page polls while pending and stops once resolved. Both fail on the pre-fix component (verified red→green locally).
- **Integration** (`tests/integration/services/storage/runsByTestCasePending.integration.test.ts`): pins the query/creation-side contract this fix depends on — `GET /api/storage/runs/by-test-case/:id` returns a `metricsStatus:'pending'` run immediately and reflects its update once judged.
- **E2E** (`tests/e2e/test-case-detail-pending-run-refresh.spec.ts`, imports `./fixtures/test-fixtures`): drives "Run Test" on the real page (intercepts only `POST /api/evaluate` to avoid depending on a real trace-mode agent/OTel round trip), asserts the new run shows "Judging…" not "Failed", then — **without any reload** — flips to "Passed" once the underlying report is updated via the real storage API, proving the poll actually fires.

## Verification

```
npm run build:all   # ✅
npm run test:unit         # 330 suites, 5590 passed, 5 pre-existing skipped
npm run test:integration  # new test passes; 5 unrelated suites (CLI/benchmark/bulk-upsert judge-timeout tests) failed —
                           # pre-existing/environmental (real Bedrock/OpenSearch round trips on a shared cluster under load
                           # during this session, one of which coincided with a mid-session AWS credential refresh); none
                           # touch TestCaseDetailPage, runs/by-test-case, or the polling logic.
npx playwright test tests/e2e/test-case-detail-pending-run-refresh.spec.ts \
  tests/e2e/test-case-detail-runs-list.spec.ts tests/e2e/test-case-detail-live-run.spec.ts
  # 9/9 passed — no regressions in sibling specs.
npm audit --audit-level=high   # pre-existing findings only (react-router/undici/vite transitive deps);
                                # no package.json/package-lock.json changes in this PR.
```

CHANGELOG.md updated under `## [Unreleased]` → `### Fixed`.


goyamegh:
## Verification proof

**Local (against a scratch server on the shared OpenSearch cluster, `AH_PORT=4981`, `BENCHMARK_RUN_RECOVERY_DISABLED=1 EVALUATION_RUN_RECOVERY_DISABLED=1`):**

```
npm run build:all            # ✅ tsc + vite build + esbuild server/cli + tsc-alias — all clean
npm run test:unit            # ✅ 330 suites, 5591 passed, 5 pre-existing skipped
npx jest tests/integration/services/storage/runsByTestCasePending.integration.test.ts
                              # ✅ 1/1 passed
npx playwright test tests/e2e/test-case-detail-pending-run-refresh.spec.ts \
  tests/e2e/test-case-detail-runs-list.spec.ts tests/e2e/test-case-detail-live-run.spec.ts
                              # ✅ 9/9 passed (no regressions in sibling TestCaseDetailPage specs)
npm audit --audit-level=high # pre-existing findings only (react-router/undici/vite transitive
                              # deps) — no package.json/package-lock.json changes in this PR
```

Red→green verified locally for all three new/updated regression tests (each fails against the pre-fix component, per commit history in this branch — the first-pass fix, then a `codex_review` pass that caught a status-misclassification + loading-skeleton-flash + unbounded-polling issue, all fixed in the second commit with dedicated regression tests).

**CI on this PR (commit `aefd5ceb2`):** `build-and-test` (Node 18/20/22), `lint-and-typecheck`, `security-scan`, `DCO`, `check-changelog`, `release-rehearsal`, `codecov/project`, `codecov/patch` — all ✅. `integration-tests` and `link-check` both show a **pre-existing failure unrelated to this PR**:
- `link-check`: dead link in `CHANGELOG.md` → `tests/e2e/resume-run-benchmark-link.spec.ts` (a file that doesn't exist on `origin/main` either — this changelog entry predates this PR and references a since-removed spec).
- `integration-tests`: the only failing test is `evaluationRuns.integration.test.ts › should promote an ad-hoc run to a benchmark`, which doesn't touch `TestCaseDetailPage`, `runs/by-test-case`, or anything in this diff. This PR's own new integration test (`runsByTestCasePending.integration.test.ts`) passed in the same CI run.

## codex_review

Ran an adversarial second-opinion review (OpenAI gpt-5.4) on the first-pass fix. Findings, all applied in the second commit:
- **HIGH** — reusing the page's local `getStatus()` helper for the runs-list row icon would misclassify a genuinely failed run (`status:'failed'`, no verdict yet) as "pending" instead of "Failed". → Replaced with the shared `getResultStatus()`/`StatusIcon`/`getStatusDescription()` primitives from `ResultStatus.tsx` (already used by `RunInspectorPage`/`BenchmarkRunDetailPage`/`RunDetailsPage`), which handle this correctly.
- **HIGH** — the polling effect called the full `loadData()`, which toggles `isLoading` and re-shows the full-page skeleton (hiding the definition hero) on every 5s tick. → Split into `loadData` (initial load) vs `refreshRuns` (background refetch, no `isLoading` toggle); the poll uses `refreshRuns`.
- **MED** — `setInterval` could fire again before a slow refresh resolved, causing overlapping requests. → Switched to a `setTimeout` chain gated on the effect's own re-run, so only one refresh is ever in flight.
- **MED** — the poll had no upper bound (one orphaned pending run polls forever). → Added a 60-attempt (~5 min) cap mirroring the server's own trace-poll give-up convention; a fresh "Run Test" resets the budget.

Added a dedicated regression test proving a genuinely-failed run still renders "Failed" (guards against re-introducing finding #1), and extended the polling test to assert the definition hero stays mounted throughout (guards against #2).

PR #468, opened 2026-09-02
feat(runs): complete run lifecycle actions — cancel/delete/re-run/retry with a prefilled run dialog

goyamegh:
## Summary

Adds a consistent run-lifecycle action matrix — **Delete / Cancel / Re-run / Retry judgement** — across every run surface (runs list, benchmark runs list, run report page, inspector header), and upgrades the Re-run confirmation into a prefilled, editable dialog. Fixes a live "not able to cancel this run" report for runs whose executor process is gone.

**Base:** built on `origin/main`, additively. A separate, larger stacked PR train is landing concurrently and touches the same list/inspector surfaces; this PR intentionally avoids large refactors of those files (small, targeted insertion points only) so it can rebase cleanly onto the train once it merges.

## Action matrix

| Action | Runs list | Benchmark runs list | Run report page | Inspector header |
|---|---|---|---|---|
| Delete | ✅ kebab + confirm | ✅ kebab + confirm | ✅ kebab + confirm | ✅ kebab + confirm |
| Cancel | ✅ only while `running` | ✅ only while `running` | ✅ (existing button, unchanged) | ✅ only while `running` |
| Re-run | ✅ (existing icon, dialog upgraded) | ✅ reuses "Add Run" dialog, prefilled from the clicked run | ✅ (existing button, dialog upgraded) | ✅ (existing button, dialog upgraded) |
| Retry judgement | ✅ EvaluationRun, terminal, judge-failed>0 | disabled (reason shown) | ✅ | ✅ |

Delete/Cancel/Retry-judgement are consolidated into one new shared `RunActionsMenu` kebab, reused verbatim across all four surfaces. Re-run intentionally stays each surface's existing, already-tested button — only its dialog's internals changed, so the working code path wasn't touched.

## Cancel: zombie-run fallback

If the process that started a run restarted/crashed, the in-memory cancellation token is gone and `POST .../cancel` used to 404 forever even though the run doc stayed `running` — this is the reported "not able to cancel this run" bug. Both cancel routes (`evaluation-runs/:id/cancel` and `benchmarks/:id/cancel`) now fall back to a doc-status-only cancel when no token is found but the run doc is still `running`, stamping an audit `cancelNote` so a fallback cancel is distinguishable from a graceful in-process one.

## Retry judgement

Re-invokes only the LLM judge against a test case's already-captured trajectory (no agent re-run) — for terminal `EvaluationRun` docs with at least one case where the agent completed but the judge verdict was `failed`. New route: `POST /api/storage/evaluation-runs/:id/retry-judgement`. The visibility predicate (`lib/runActions.ts`) is shared between the UI's item-enablement and the server route's validation so they can't drift. Scoped to top-level `EvaluationRun` docs in this PR — legacy benchmark-embedded runs show the item disabled with a reason, consistent with this repo's stated direction of converging on the evaluation-run model.

## Re-run dialog upgrade

`RerunConfirmDialog` is now prefilled from the source run and every field is editable before launch: Run Name, Agent, **Evaluator** (now visible + editable — previously carried silently), Judge Model, and — behind an "Advanced" disclosure — Concurrency and a benchmark swap for the test-case source. Uses the same field components (`Select` / `JudgeModelSelect` / `Input`) as the existing "Add Run" Configure-Run dialog rather than introducing a new one. `POST .../rerun` now accepts an optional overrides body; an untouched rerun stays a faithful duplicate, a tweaked one is flagged `modified: true` in addition to the existing `rerunOf` provenance link.

For the legacy benchmark-embedded run model, Re-run reuses that page's own existing Configure-Run ("Add Run") dialog, prefilled from the clicked run instead of always the latest one — no second rerun endpoint for a run model already being phased out.

## Tests

- **Unit:** action-visibility matrix per run status/docType (`lib/runActions.ts`), rerun-override merging (`applyRerunOverrides`), retry-judgement service, the three new/changed routes (rerun overrides, retry-judgement, cancel zombie fallback ×2 — evaluation-runs and benchmarks).
- **Integration** (real server, real HTTP): both cancel zombie fallbacks, rerun-with-tweaks producing a modified config + provenance, retry-judgement end-to-end via the `demo-model` judge provider (no AWS credentials required).
- **E2E:** kebab menu renders per run status/kind, delete confirm flow, cancel visible only while running, rerun dialog prefill + "modified" hint, inspector header parity. Also fixed a latent always-skipped test in an existing rerun e2e spec (benchmark seeding used `PUT` on a nonexistent id) and a strict-mode locator collision in two existing specs (`text=Failed` matching both a stats label and a `status:'failed'` badge) — both discovered while extending this area, unrelated regressions on `main`.

Coverage on changed files (unit): `lib/runActions.ts` 100%, `lib/evaluationRerun.ts` 100%, `services/evaluationRerun.ts` 100%, `services/evaluationRetryJudgement.ts` ~94%, `server/routes/storage/evaluationRuns.ts` ~94% lines (integration-only paths aren't counted by Jest's instrumentation, as with the rest of this file).

## Known limitations / follow-ups (flagged, not silently dropped)

- **`useTraces` and "persona/conditions"** are not per-run fields today (`useTraces` is agent-level config; judge instructions live on the Evaluator, which the dialog does expose). Making them per-run overrides would need a larger server change — out of scope here, noted for a follow-up.
- **Retry judgement** is not implemented for legacy benchmark-embedded runs in this PR (menu item shown disabled with a reason) — would need a parallel endpoint operating on `benchmark.runs[]` rather than the top-level `EvaluationRun` collection.
- Rebases onto the concurrent stack train once it lands (additive design, no conflicting refactors expected).

## Verification

```
npm run build:all   # PASS
npm run test:unit         # 334 suites / 5679+ tests passed
npm run test:integration  # new evaluationRunLifecycleActions suite: 11/11 passed; full suite green except 3 pre-existing,
                           # environment-caused failures unrelated to this change (missing OpenSearch cluster / Bedrock
                           # model access in this sandbox — confirmed via server logs, not a regression)
npm run test:e2e          # new run-actions-menu spec: 5/5 passed; full suite 467+ passed, 2 pre-existing failures
                           # (Dashboard copy-text drift, Skills page) unrelated to this change, confirmed by inspecting
                           # the untouched source files
npm audit --audit-level=high   # no new vulnerabilities (all pre-existing transitive deps, unchanged by this PR)
```

DCO signoff present on the commit. `CHANGELOG.md` updated under `## [Unreleased]`.


goyamegh:
## codex_review (gpt-5.4, different model family) — findings & resolution

Ran against the core production diff (routes, lib, services, dialog/menu components) after CI-equivalent local checks were green.

|---|---|---|---|
| 1 | HIGH | Cancel zombie-fallback could theoretically mark a brand-new run cancelled in the narrow window between it being persisted and its executor registering a cancellation token (also flagged the pre-existing, separate limitation that cancellation tracking is per-process, not multi-process-aware) | **Applied** — both cancel routes now require a run to be ≥5s old (`ZOMBIE_CANCEL_MIN_AGE_MS`) before taking the fallback path, 409ing (retryable) otherwise. The multi-process limitation is a pre-existing, already-documented constraint of this codebase's run-tracking model (see `AGENTS.md`'s orphan-run-recovery notes) — not something this PR could fix without the heartbeat-based ownership that doc already calls out as the eventual replacement; documented explicitly in `lib/runActions.ts` instead of left implicit. |
| 2 | HIGH | Retry judgement re-fetched the test case's CURRENT (possibly-edited) definition instead of what the original run actually used, silently changing what "retry" means | **Applied** — now judges against the run's snapshotted `testCaseSnapshots[].version` via `storage.testCases.getVersion()`, falling back to the current doc only for legacy runs with no recorded version. |
| 3 | MED | The rerun dialog's benchmark-source dropdown could send `benchmarkId: null`, clearing the association field while `applyRerunOverrides` left `sources` untouched — an internally inconsistent result | **Applied** — the dialog's sentinel option now always means "no override" and never sends `null`; the `null`-clears-to-default capability stays supported at the `applyRerunOverrides`/API layer for other callers, just isn't reachable from this dialog. |
| 4 | MED | Two concurrent retry-judgement requests for the same run would race on the same wholesale results/stats overwrite (last-write-wins, doubled judge-call cost) | **Applied** — the route now rejects a second concurrent request for the same run with 409. |
| 5 | LOW | `RunActionsMenu`'s escape hatches (`hideCancel`, `variant`, `align`) are "abstraction tax" vs. a simpler design | **Rejected, with reason** — the four surfaces it's reused on have genuinely different pre-existing constraints (one page already has its own destructive-variant Cancel button predating this PR; sizing differs between a compact list row and a page header). These aren't speculative flexibility, they're accommodating real heterogeneity across pages built over time. |

All findings applied above are covered by new/updated tests (unit: `lib/runActions.test.ts`, `evaluationRuns.test.ts`, `evaluationRetryJudgement.test.ts`; integration: `evaluationRunLifecycleActions.integration.test.ts`).

## Verification (after codex fixes)

```
npx tsc --noEmit -p tsconfig.json         # clean
npm run build:all                          # PASS
npm run test:unit                          # 334 suites / 5691 tests passed
npm run test:integration -- --testPathPatterns=evaluationRunLifecycleActions   # 12/12 passed (incl. new 409 age-guard + concurrent-retry tests)
npm run test:integration -- --testPathPatterns=evaluationRuns.rerun.integration # 8/8 passed
npx playwright test tests/e2e/run-actions-menu.spec.ts tests/e2e/rerun-evaluation-run.spec.ts   # 16/16 passed
npm audit --audit-level=high               # no new vulnerabilities (pre-existing transitive deps only, unchanged by this PR)
```

DCO signoff present on both commits.

goyamegh:
### Scope change (owner feedback): header actions collapsed into the "…" kebab — `0acb24f15`

**What changed**
- **Run inspector header** (`/evaluations/runs/:id/inspect` and the benchmark-scoped `/evaluations/benchmarks/:bid/runs/:id/inspect`): the standalone **Re-run**, **Retry judgement (n)** and **Compare** buttons are removed. The kebab is now the only action surface and carries **Re-run · Cancel (running only) · Retry judgement (n) · Delete**. Compare is gone from this header (the comparison page stays reachable from the runs list / compare nav).
- **Run report page** (`/evaluations/runs/:id`, same run via the other route): same treatment for parity — its standalone Re-run and Cancel buttons fold into the kebab (the `hideCancel` prop is gone, no caller needed it). Non-lifecycle buttons ("Customize before re-running…", "Convert to Benchmark", "View Benchmark") stay.
- `RunActionsMenu`: opt-in `onRerun` item (with `canRerun` + reason for legacy benchmark-embedded runs), count-bearing `Retry judgement (N)` label, `aria-label` on the trigger, menu closes before handing off to a confirm dialog.
- Gating is unchanged from the removed buttons — Re-run: EvaluationRun docs only; Retry judgement: EvaluationRun + terminal + judge-failed count > 0 (inspector keeps the report-derived count that the confirm dialog shows); Cancel: `running`; Delete: confirm.

**Kebab gating matrix (inspector)**

| run state | Re-run | Cancel | Retry judgement | Delete |
|---|---|---|---|---|
| running eval-run | enabled | shown | disabled — "only available once the run has finished" | shown |
| completed, judge failures | enabled | hidden | enabled, `(n)` | shown |
| completed, clean | enabled | hidden | disabled `(0)` — "No judge-failed cases to retry" | shown |
| legacy benchmark-embedded run | disabled (reason) | per status | disabled (reason) | shown |

**Verification**
```
tsc --noEmit                                          clean
jest tests/unit/components/evals3 + lib/runActions    10 suites, 129 passed
  RunInspectorPage.test.ts                            37 passed (new block: standalone buttons absent,
                                                      exactly four item kinds, enabled/disabled+reason per state)
jest evaluationRunLifecycleActions.integration        11 passed
playwright run-actions-menu / retry-judgement /
  rerun-evaluation-run (real server, CI=1)            27 passed
playwright evaluation-runner "Cancellation UI"        3 passed
```
Full unit run: 357/360 suites green; the 3 failing suites (`cli/index`, `loader.mjsReload`, `loader.multiFile`) fail identically on the previous tip in this scratch checkout because `lib/dist`/`cli/dist` weren't built there — unrelated to this change.

codex_review pass run on the diff; applied: removed the now-dead `hideCancel` prop. Other findings (retain persistent Cancel error state, keep Compare in the header, don't touch the report page) were weighed and not applied — the owner's request explicitly collapses everything into the dropdown, the kebab shows its own busy spinner + inline error, and the report page was aligned deliberately so both routes rendering the same run match.

TackAdam (approved):

goyamegh:
### Follow-up: one shared run dialog (Add Run + Re-run) with a Concurrency field — `bd789b2`

Owner feedback addressed: *"Concurrency is also missing in the Run dialog box, and re-run should open the same dialog box with prepopulated box."*

**What changed**
- New `components/evals3/RunConfigDialog.tsx` (`mode: 'create' | 'rerun'`, discriminated props). The benchmark page's **Add Run** renders it in create mode; **Re-run** (runs-list row, report-header kebab, inspector-header kebab) renders the *same* component in rerun mode, prepopulated from the run doc's persisted values (name → `<name> (re-run)`, agent, evaluator, judge model, concurrency, benchmark source). `RerunConfirmDialog.tsx` is deleted.
- **Concurrency** is a first-class field in both modes (`1..20`; empty = runner default `1`, shown as placeholder; tooltip). Create mode forwards it on `POST /api/storage/evaluation-runs`; rerun mode sends it in the `/rerun` overrides only when it differs from the source (faithful duplicates stay `modified`-free).
- If the source run's agent / evaluator no longer exists in config it's shown as a disabled "(not in config)" entry with a hint and submit is blocked — never silently swapped. "Missing evaluator" is only asserted after the evaluator list actually loaded.
- Re-seed is keyed on the source run *id*, so a parent poll/refetch of the same run while the dialog is open doesn't wipe in-progress edits.

**Verification**
```
npm run build:all                                   ✓
npm run test:unit                                   Test Suites: 361 passed | Tests: 6194 passed, 5 skipped
tests/unit/components/evals3/RunConfigDialog.test.ts   18 passed (both modes, prepopulation incl. concurrency + evaluator,
                                                      diff-only override body, missing agent/evaluator, refetch-while-editing)
integration (real server, shared cluster):
  evaluationRuns.rerun + evaluationRunLifecycleActions  19 passed
e2e (Playwright, real server + real cluster):
  run-actions-menu / rerun-evaluation-run / evals3-benchmark-runs   37 passed, 1 flaky (pre-existing: BenchmarkEditor
  loads the full test-case collection, ~15 s on this dataset; fails identically on the unchanged base)
  new: inspector kebab Re-run → shared dialog prepopulated (name/agent/evaluator/concurrency) → submit creates a real
       run with rerunOf, concurrency == source, modified falsy
  new: retired agent → disabled current selection + hint, Re-run blocked, re-enabled after picking a configured agent (real Radix)
  Add-Run dialog renders Concurrency; Start Run posts concurrency: 3 to the unified runner
```

**codex_review** (gpt-5.4, adversarial): 14 findings; applied — re-seed on run id not object identity (HIGH), discriminated-union props (MED), block submit on a *known*-missing evaluator + don't flag on fetch failure (MED), real-Radix e2e for the missing-agent path (HIGH), refetch-while-editing + legacy-no-concurrency tests (MED). Rejected with reason — "clearing concurrency can't reset to default" (it can: empty ⇒ `1`, sent when source ≠ 1; unit-tested), "close-before-onStart loses edits" (create mode closes-then-runs exactly as before and the parent keeps the last values; rerun mode closes only after success), "single dialog is over-engineered" (the single dialog is the explicit owner ask), double-close / `submitting` guard (standard pattern here, three call sites unchanged).

goyamegh:
**Follow-up commit `bb0df089d` — inspector kebab Delete dispatches on the run's kind, not the route** (owner report: *"the run doesn't get deleted when I go inside the run page and try it myself"*)

Repro on a real instance (read-only; the DELETE was intercepted): on `/evaluations/benchmarks/<bid>/runs/<rid>/inspect` for a run that exists as a standalone evaluation-run doc (not embedded in `benchmark.runs[]` — 27 of 48 runs on one live benchmark), the kebab's Delete issued

```
DELETE /api/storage/benchmarks/<bid>/runs/<rid>   → 404 {"error":"Run not found"}
```

`asyncBenchmarkStorage.deleteRun` swallowed the 404 as `false`, no error was shown, and the page navigated to the runs list with the run still there.

**Fix:** `handleDelete` dispatches on `isEvaluationRun(run)` (the run shown on the benchmark route IS the first-class doc whenever one exists) → `DELETE /api/storage/evaluation-runs/:id`; a legacy embedded-only run keeps the benchmark nested-run endpoint; a swallowed `false` now surfaces as the kebab's inline error instead of a silent navigate-away. Confirm copy: reports are kept (non-cascade policy), running runs are flagged. Removing a dual-written run's embedded projection as well is server-side and lands in #492 (independent).

Red → green:

```
# unit, tests/unit/components/evals3/RunInspectorPage.test.ts — before:
✕ benchmark route + first-class evaluation-run doc → DELETE via the evaluation-runs API, then back to the benchmark runs list
✓ benchmark route + legacy embedded-only run (no doc) → the benchmark nested-run DELETE
✕ a swallowed failure (benchmark deleteRun → false) surfaces an error and does NOT navigate away
✓ eval-run route → evaluation-runs API, then back to the runs list
# after: 41 passed, 41 total

# e2e, tests/e2e/run-delete-from-inspector.spec.ts — before (old client + old server):
✘ standalone (non-embedded) run: kebab → Delete → confirm → back on the benchmark runs list, run gone
    Expected: 404   Received: 200        ← navigated away, run still exists
# after:
✓ standalone (non-embedded) run: kebab → Delete → confirm → back on the benchmark runs list, run gone (1.9s)
   observed call: DELETE /api/storage/evaluation-runs/<rid> → 200; GET afterwards → 404
```

Existing `run-actions-menu.spec.ts` (10 tests) still green. `codex_review` on the combined change: no findings specific to this commit beyond the confirm copy — it must not promise "will be cancelled first" (cancellation on delete is best-effort/server-side in #492), so the copy now only flags that the run is still running.

PR #469, opened 2026-09-02
fix(metrics): comparison-page Cost/Tokens/LLM Calls stayed empty for trace-mode runs

goyamegh:
## Summary

Comparison-page **Cost / Tokens / LLM Calls** columns stayed empty for
trace-mode runs even though the Traces tab found the exact same spans for
the same run. Root-caused with a read-only live bug hunt against a real
comparison of two trace-mode runs on the shared observability cluster — a
subprocess (Claude-Code-family) agent and a REST-connector agent — that
turned up **three independent gaps**, each masking a different symptom, in
`server/services/metricsService.ts`'s `computeMetrics`/`computeBatchMetrics`.

### Gap 1 — no Strategy-A (traceId) correlation at all

REST-connector agents never get a native runId (`RESTConnector.execute()`
returns none), so `report.runId` falls back to `report.traceId` upstream;
subprocess agents whose vendor SDK never adopts `agent_health.run.id` /
`gen_ai.conversation.id` are in the same boat. The metrics query only ever
tried Strategy B, so these runs 0-correlated even though `/api/traces
{traceId}` found their spans instantly.

`computeMetrics(runId, config, traceId?)` / `computeBatchMetrics(runIds,
config, traceIdByRunId?)` now OR in a direct `traceId` term/terms clause;
`GET /api/metrics/:runId?traceId=` and `POST /api/metrics/batch {traceIds}`
carry it end-to-end from `report.traceId`.

### Gap 2 — span attributes read from the wrong OpenSearch schema

The live `otel-v1-apm-span-*` index stores attributes in the legacy
`span.attributes.<key with @ for .>` flattened shape, not the nested
`attributes` object this file read directly — every `attrs['gen_ai...']`
lookup silently returned `undefined` on correctly correlated spans
(confirmed live: `toolCalls` counted right from `span.name`, `llmCalls`/
tokens stayed `0`). Token/model reads now go through `transformSpan()`
(already used by the Traces tab / `/api/traces`), reusing the one existing
normalizer instead of adding a second one.

### Gap 3 — vendor SDK token-key naming

A real Claude Code `claude_code.llm_request` span stamps
`gen_ai.request.model` correctly but reports usage under bare
`input_tokens`/`output_tokens`, not the OTel registry names
`gen_ai.usage.input_tokens`/`gen_ai.usage.output_tokens` this file read
exclusively — now falls back to the vendor keys, matching the fallback
`lib/matchers/traces.ts` / `services/traces/traceSummary.ts` already use
elsewhere. Spans sharing a trace with agent-health's own eval/judge spans
(now reachable via the Strategy-A traceId match) are excluded from the
agent's own token/LLM-call count.

### Bonus — batch-only `_source` projection regression

Even with 1–3 fixed, `computeBatchMetrics` (the exact function the
comparison page calls) still returned all-zero tokens while the single-run
`computeMetrics` (no `_source` restriction) returned the real numbers for
the *same* run — its `_source` field projection omitted the
`span.attributes.*` / `resource.attributes.*` wildcards. Caught only by
testing the real batch code path.

### Verification against the real reported runs (read-only, shared cluster)

Before: `POST /api/metrics/batch` → `totalTokens: 0, llmCalls: 0` for both
runs. After the fix: real non-zero tokens/cost/LLM-calls for both — a
subprocess-connector run and a REST-connector run, isolated correctly in
one batched query.

### Also fixed the same bug outside the comparison page

Threaded the new `traceId` correlator through every other
`fetchBatchMetrics`/`fetchRunMetrics` call site that has the identical bug
(`RunDetailsContent`, `ReportsPage`, `BenchmarkSummaryCharts`,
`RunSummaryPanel`, `Dashboard`) — not just `ComparisonPage`.

## Tests

- **Unit**: `tests/unit/server/services/metricsService.test.ts` (vendor-key
  token fallback, eval/judge-span exclusion, traceId should-clauses,
  batch grouping-by-`gen_ai.conversation.id`/traceId fix),
  `tests/unit/server/routes/metrics.test.ts` (route-level `traceId`/
  `traceIds` plumbing + validation), `tests/unit/services/metrics.test.ts`
  (client-side query-param/body plumbing),
  `tests/unit/services/comparisonService.test.ts`
  (`collectTraceIdsFromReports`), `tests/unit/components/RunDetailsContent.test.ts`.
- **Integration** (real query-building + real span-parsing; only the
  OpenSearch client's `.search()` is faked — an in-memory index that ALSO
  applies real `_source` field-projection semantics, so a regression to the
  projection list makes the test fail, not just a query-body string match):
  `tests/integration/services/traces/metricsTraceIdAndVendorTokens.integration.test.ts`.
  Verified this suite goes red without the `_source` fix by reverting it
  locally and re-running.

```
Test Suites: 335 passed, 335 total
Tests:       5 skipped, 5700 passed, 5705 total
```

`npm run build:all` clean. `npm audit --audit-level=high`: pre-existing
transitive vulnerabilities only (react-router-dom / undici via a dev
dependency / vite) — unrelated to this change.

## Notes for reviewers

- Deliberately does **not** touch `services/storage/asyncRunStorage.ts`'s
  separate `runId: stored.traceId || (stored as any).runId` quirk — that's
  a real but independent bug (worth a follow-up) that this fix doesn't
  need: `report.traceId` is always available as its own field regardless of
  that mapping, which is what the new correlator uses.
- Deliberately does **not** add Strategy D (`session.id`) correlation here —
  that already lives on an existing large in-flight PR
  (`goyamegh/comparison-ui-declutter`); this PR only adds what neither
  `main` nor that PR has yet (Strategy A traceId correlation + span-schema/
  vendor-key normalization + the `_source` projection fix).


goyamegh:
## codex_review (gpt-5.4) — findings + disposition

Ran an adversarial review against a self-contained brief of this diff before posting. All findings and what I did with each:

1. **HIGH** — new `traceId` correlator is "fail-open" (OR broadens results on any input). **Rejected.** `traceId` here isn't attacker/user-controlled input — it's always `report.traceId`, the trace id of the eval span the SAME report's own runner emitted for that specific test-case invocation. ORing two server-derived correlators for the same report is the exact pattern this file (and `services/traces/tracesService.ts`) already uses for Strategy B (`agent_health.run.id` OR `gen_ai.conversation.id`).
2. **HIGH** — `traceIdByRunId` assumes one traceId per run; shared traces could collapse the reverse map. **Rejected, but documented.** `startTestCaseSpan` mints a fresh span/traceId per test-case invocation — never reused across runs or shared between concurrent test cases. Added a doc comment on `buildRunIdShouldClauses` making this invariant (and that it's inherited from the existing `/api/traces` Strategy-A consumer, not new) explicit.
3. **HIGH** — "count every span on the trace" is too coarse without a stated invariant; needed the eval/judge exclusion as a tell. **Same rationale as #2** — addressed with the same doc comment; the eval/judge exclusion is defense-in-depth for the case where a judge call *does* run inside the eval span's context (the SDK inline-judge path), not evidence the invariant is missing.
4. **HIGH** — leaving `services/storage/asyncRunStorage.ts`'s `runId: stored.traceId || stored.runId` precedence bug unfixed while touching adjacent identity plumbing. **Applied.** Flipped to `(stored as any).runId || stored.traceId` (prefer the real connector runId; fall back to traceId only when the connector never returned one — REST connectors). Added 2 regression tests.
5. **MED** — "Strategy D lives on another open PR" isn't a technical scoping defense. **Rejected, reasoning tightened in PR description** — this is a repo policy call (avoid duplicate/competing implementations of the same correlation strategy across two open PRs), not a technical limitation; Strategy A alone was verified sufficient for the two reported repro runs.
6. **MED** — `readAttrs` → `transformSpan()` cross-layer coupling / possible reconstruction cost. **Rejected** — reuses the ALREADY-correct, already-tested normalizer instead of re-deriving the same @-raw/plain-raw merge a second time in this file; batch size caps at 10k spans/chunk per request, not a hot loop.
7. **MED** — vendor token-key fallback chain applies to every span, not gated to LLM-only spans. **Rejected for this PR** — pre-existing behavior (the original code already summed `gen_ai.usage.*` unconditionally per span); not a regression this diff introduces, flagged as a separate follow-up.
8. **MED** — `isEvalOrJudgeSpan` name-string matching is fragile. **Rejected** — matches internal span names agent-health's own OTel helper (`startTestCaseSpan`) emits; not agent/tool-controlled, symmetric with existing conventions (`resolveJudgeRunId`'s doc comments reference the same names).
9. **MED** — claimed the `_source`-projection integration test might be "coverage theater". **Rejected with evidence** — manually reverted the `METRICS_SOURCE_FIELDS` fix and re-ran; the new test goes red (`Expected: 36547, Received: 0` and an explicit `_source` field-list assertion failure), confirmed green again after restoring the fix.
10. **LOW** — helper functions add indirection without a single source of truth. **Acknowledged, no action** — general style note, not a specific defect.

Also re-verified against the real reported runs on the shared cluster (read-only) after applying finding #4: `POST /api/metrics/batch` for both runs still returns the real non-zero tokens/cost/LLM-calls.

goyamegh:
## Verification proof

**Full test suite** (after all fixes, including the codex-review follow-up commit):
```
Test Suites: 335 passed, 335 total
Tests:       5 skipped, 5702 passed, 5707 total
```
`npm run build:all` clean.

**Live cluster before/after** (read-only diagnosis, then re-verified after the fix — two real trace-mode runs from the reported comparison URL: one subprocess-connector run, one REST-connector run; agent/run names withheld per internal-naming policy, only opaque runId/traceId shown):

Before (on `main`, single-run endpoint):
```json
{"runId":"3afe42681cd16f6c9e1d3370cbcd7492","totalTokens":0,"costUsd":0,"llmCalls":0,"toolCalls":3,"status":"success"}
```

After (this branch, `POST /api/metrics/batch` — the exact call the comparison page makes):
```json
{
  "metrics": [
    {"runId":"subprocess-1788335139441","traceId":"3afe42681cd16f6c9e1d3370cbcd7492",
     "inputTokens":159764,"outputTokens":876,"totalTokens":160640,
     "costUsd":0.4924319999999999,"llmCalls":5,"toolCalls":3,"status":"success"},
    {"runId":"c7cf50fdecabc421db6dffe004c68266","traceId":"c7cf50fdecabc421db6dffe004c68266",
     "inputTokens":13024,"outputTokens":834,"totalTokens":13858,
     "costUsd":0.051582,"llmCalls":1,"toolCalls":0,"status":"success"}
  ],
  "aggregate": {"totalRuns":2,"totalCostUsd":0.5440139999999999,"totalInputTokens":172788,"totalOutputTokens":1710,"avgLlmCalls":3}
}
```

Reproduced by booting a local server (`BENCHMARK_RUN_RECOVERY_DISABLED=1 EVALUATION_RUN_RECOVERY_DISABLED=1 TRACE_RECOVERY_DISABLED=1`, unique port) pointed at the same shared observability cluster and re-querying the exact runIds/traceIds from the reported comparison URL, both before and after each fix commit.

**Regression-lock proof for the `_source` projection fix** (integration test): reverted `METRICS_SOURCE_FIELDS` locally and re-ran — test goes red:
```
Expected: 36547
Received: 0
Expected: ArrayContaining ["span.attributes.*", "resource.attributes.*"]
Received: ["attributes", "resource", "name", "traceId", "startTime", "endTime", "durationInNanos", "status"]
```
Restored the fix — green again (see `tests/integration/services/traces/metricsTraceIdAndVendorTokens.integration.test.ts`).

goyamegh:
## Scope addition: "Average accuracy" rename + "Avg score" aggregate + primary-rubric chip

Extended this PR with a follow-up to the "Avg Accuracy renders '--' for custom-evaluator runs" fix above. `--` was honest, but a run scored entirely by a custom evaluator still had **no quality aggregate at all** on the scoreboard, and its per-case cells rendered a bare verdict with no number.

1. **Label:** the accuracy row on the live comparison surfaces is now "Average accuracy" (was "Avg Accuracy"), with a tooltip explaining the derivation.
2. **New run-level aggregate "Avg score"** (`RunAggregateMetrics.avgScore`): mean, over test cases with a derivable score, of each case's overall score — `metrics.accuracy` → the report's primary rubric → the mean of every numeric metric on the report (documented three-tier fallback). Renders `--` only when not one case in the run has any score.
3. **Per-case chip fallback:** `MetricCell` now shows the report's primary rubric (a compact `name value%` chip, full name in the tooltip) whenever `accuracy` is absent, instead of a bare verdict with no number.

Primary-rubric selection is the alphabetically-first metric key with a finite numeric value — deliberately matching the evaluator-agnostic convention `RunScore`/`getRunOverallScore()` already use, rather than adding a new evaluator-doc fetch to the comparison page's data flow (documented as a follow-up in `getPrimaryRubricKey`'s doc comment if that dependency is ever added).

### codex_review (gpt-5.4) — adversarial pass on this addition

|---|---|---|---|
| 1 | Alphabetical "primary rubric" isn't the evaluator's true declared/highest-weight rubric | Major | Rejected — documented tradeoff (no evaluator-doc fetch in scope); tooltip clarifies it's the evaluator's own metric, not a declared primary |
| 2 | Score formatting ignores `ScoringMetric.scale` (assumes 0–100) | Major | Rejected — every evaluator template/example in-repo uses `scale: 100`, consistent with every other percent-formatted metric surface already in this codebase (not a regression this change introduces) |
| 3 | Internal contradiction: per-case chip has no delta ("noise"), but the run-level aggregate delta is colored | Major | **Applied** — reworded the per-case comment (the real reason is a missing `baselinePrimaryRubric` prop, not "noise"); added a tooltip caveat to the aggregate delta that it's only directly comparable across runs scored by the same evaluator |
| 4 | Tier-3 "mean of all metrics" fallback in `computeOverallScore` is documented as currently unreachable | Minor | Rejected — this exact three-tier fallback was the specified design; kept as a documented defensive fallback |
| 5 | `primaryRubric` was attached to every result even when `accuracy` was present — a heuristic value looking authoritative where never read | Minor | **Applied** — now only populated when `accuracy` is absent |
| 6 | Renaming labels in `RunSummaryTable.tsx`/`AggregateMetricsTable.tsx` (confirmed dead/unwired, not reachable from `ComparisonPage.tsx`) was scope creep | Minor | **Applied** — reverted both files |
| 7 | Test coverage doesn't exercise cross-evaluator / non-percent-scale failure modes | Major | Rejected — those require the evaluator-fetch / scale-awareness work above, which is out of scope; existing tests lock in the current documented behavior |

### Proof

```
Test Suites: 335 passed, 335 total
Tests:       5 skipped, 5734 passed, 5739 total
```

`npm run build:all` clean. `tsc --noEmit` clean. `npm audit --audit-level=high`: pre-existing transitive vulnerabilities only (dev-tooling / vite), unrelated to this change.

New tests: `tests/unit/services/comparisonService.test.ts` (`getPrimaryRubricKey`/`getPrimaryRubric`, `computeOverallScore`'s three tiers, `avgScore` for accuracy-present / custom-evaluator-only / empty runs, `primaryRubric` attachment for both accuracy-present and accuracy-absent reports), `tests/unit/components/comparison/MetricCell.render.test.ts` (primary-rubric chip renders/omits, accuracy precedence, errored suppression), and two new e2e assertions appended to the existing seeded-fixture `tests/e2e/comparison-accuracy-dash.spec.ts`.

goyamegh:
### Scope addition: scoreboard column tooltips + clickable run names (`a23a819d5`, `fadf6476e`)

Owner ask folded into this PR: *"Each column should be explainable by a hover with a one line description. The run names should be clickable to the run link, complete name should show up on hover."*

**Column tooltips** (`ComparisonScoreboard.tsx`, headers now driven by `SCOREBOARD_COLUMNS`; every `<th>` has a one-line `title`):

| Column | Hover text |
|---|---|
| Run | Run name — click to open the run report |
| Pass Rate | % of test cases whose verdict is pass |
| Average accuracy | Mean of the judge-graded accuracy over test cases that report one |
| Avg score | Mean per-case overall score: accuracy if present, else primary rubric, else mean of all rubric metrics |
| Cost | Total LLM cost across all test cases in the run |
| Avg Duration | Mean wall-clock duration per test case |
| Tokens | Total tokens across all test cases |
| LLM Calls | Total LLM calls across all test cases |
| Tool Calls | Total tool invocations across all test cases |
| Coverage | Test cases this run shares with the comparison set |
| Delta (footer) | A minus B per column; blue/green when A is better on that metric, red when worse |

**Run-name links**: the name on each scoreboard row and each run header in `UseCaseComparisonTable` is now a `<Link>` with the full name as its `title`. Route comes from a new shared `lib/runReportPath.ts` (benchmark-scoped `/evaluations/benchmarks/:benchmarkId/runs/:runId` for runs embedded in a benchmark; bare `/evaluations/runs/:runId` for ad-hoc eval-runs / label-only associations — same rule the existing "Open run" icon already used; the icon now calls the same helper so the two can't disagree). Ids are path-segment encoded. Open/Remove icon actions unchanged.

**Tests**
- Unit (RTL): `tests/unit/components/comparison/ComparisonScoreboard.columnTooltips.render.test.ts` — every header has a non-empty one-line title with exact wording pinned (esp. Average accuracy vs Avg score); name anchors carry correct href + full-name title for benchmark vs ad-hoc runs; agent-name fallback; per-case header anchors; `runReportPath` incl. encoding.
- E2E: `tests/e2e/comparison-open-run-deep-link.spec.ts` extended — name-link hrefs/titles for a benchmark run and an ad-hoc eval-run, 10 header tooltips present, keyboard (focus+Enter) activation lands on the inspector, per-case header click lands on the eval-run route. `comparison-scoreboard-clarity.spec.ts` updated (`div[title]` → `a[title]` + href).

**Verification (local, branch build)**
```
npm run build:all                                       ✓
npm run test:unit                                       Test Suites: 361 passed · Tests: 6176 passed, 5 skipped
playwright: comparison-open-run-deep-link, comparison-scoreboard-clarity,
            comparison-accuracy-dash (CI=1, prod build, port 5019)   12 passed
tsc --noEmit                                            ✓
```
Rendered check against a real two-run comparison: both scoreboard name links and both per-case header links resolve to the benchmark-scoped inspector route; all 10 header `title`s present.

**codex_review (gpt-5.4)** — applied: helper moved out of the component into `lib/runReportPath.ts` (dependency direction), ids URL-encoded, per-case header keeps `title` on the truncating wrapper too (hover target), Delta wording made per-metric, keyboard e2e added. Rejected: replacing native `title` with a focusable Tooltip primitive (the page's existing convention is native `title` on headers/cells; owner asked for a hover, kept consistent), restoring the two-sentence Avg score header text (owner asked for one line; the `--`-semantics caveat lives on the delta cell tooltip), loosening exact-string test assertions (the wording precision was the explicit ask).

TackAdam (approved):

PR #470, opened 2026-09-02
feat(compare): hover preview of test-case prompt on case links

goyamegh:
## Summary

Every test-case hyperlink on the comparison page (`/compare`) now surfaces a hover/focus preview of the test case's input prompt — the case row's name link, and the expanded row's "View full test case" link.

**Stacks on #460** (comparison page overhaul) — this branch is based on #460's current tip so it applies cleanly once that merges. If #460 hasn't merged yet when this is reviewed, please hold merge on this one until it does.

## What it shows

- Test case name, category/difficulty badges (when cheap/already available), and a ~12-line clamped preview of the initial prompt with a bottom fade.
- Opens on hover (250ms intent delay) or keyboard focus.
- Shows the **version the run actually used** (`report.testCaseVersion`), not necessarily today's edited content — falls back to the latest known version when the exact one isn't available.

## Implementation notes

- **Primitive**: built on the existing `@/components/ui/tooltip` (Radix Tooltip — already a dependency, already used this way by `VersionIndicator`) instead of adding `@radix-ui/react-hover-card`. Tooltip already provides the open-intent delay, portal rendering, collision detection, and focus-triggered display.
- **Data source**: a comparison row's report only carries `{id, version, name}` for its test case — no prompt content. `GET /api/storage/test-cases/:id` (`asyncTestCaseStorage.getById`) always comes back with `versions: []` (the client-side conversion deliberately omits history); the only endpoint that returns per-version content is `GET /api/storage/test-cases/:id/versions`. A module-level `Map<testCaseId, Promise<TestCaseVersion[]>>` cache guarantees at most one fetch per id no matter how many hover triggers reference the same case (row-sweep dedup).
- New files: `components/comparison/TestCasePromptHoverCard.tsx`, `services/comparison/testCasePromptCache.ts`, `hooks/useTestCasePromptPreview.ts`.

## Tests

- Unit: `tests/unit/services/comparison/testCasePromptCache.test.ts` (cache dedup — one fetch per id despite N hovers/link sites; `selectPromptForVersion` exact/middle/string/fallback/no-history/no-prompt cases), `tests/unit/hooks/useTestCasePromptPreview.test.ts` (inactive never fetches; re-activation and multi-site hovers for the same id still fetch once; failure surfaces `error` without throwing).
- E2E: `tests/e2e/comparison-hover-prompt-preview.spec.ts` (new, `data-testid="compare-hover-prompt"`) — hovering the case-row link shows the run's captured prompt + version badge (not the test case's current edited content), keyboard focus opens the same preview, and the expanded row's "View full test case" link opens it too. Fully mocked via `page.route()`; imports `./fixtures/test-fixtures`.

## Verification

```
$ npx tsc --noEmit                     # clean
$ npm run test:unit                    # 341 suites, 5855 passed
$ npm run test:integration              # 91 suites, 871 passed
$ npx playwright test tests/e2e/comparison-hover-prompt-preview.spec.ts
  3 passed
$ npm run build:all                    # clean
```

No new dependencies added (`package-lock.json` unchanged).


goyamegh:
## Verification proof

```
$ npx tsc --noEmit
(clean)

$ npm run test:unit
Test Suites: 341 passed, 341 total
Tests:       5 skipped, 5855 passed, 5860 total

$ npm run test:integration
Test Suites: 3 skipped, 91 passed, 91 of 94 total
Tests:       7 skipped, 871 passed, 878 total

$ npx playwright test tests/e2e/comparison-hover-prompt-preview.spec.ts
  ✓ focusing the case row link (keyboard) also opens the preview
  ✓ hovering the case row link shows the RUN's prompt (not today's edited content)
  ✓ expanding a row and hovering "View full test case" also shows the preview
  3 passed

$ npm run build:all
(clean — client + server + cli + lib all build)

(pre-existing transitive vulnerabilities in react-router-dom/undici/vite,
 unrelated to this change — no new dependencies added, package-lock.json
 unchanged)
```

## `codex_review` (adversarial second-opinion, gpt-5) — findings and disposition

| # | Finding | Severity | Applied? | Notes |
|---|---|---|---|---|
| 1 | Version selection in the expanded-row's task strip depends on which run's report loads first | major | No | Self-correcting once all reports load (settles on the first run in stable array order); consistent with the rest of the page's incremental-load UX (cells already fill in as reports stream). Not a new failure mode. |
| 2 | `selectPromptForVersion`'s fallback-to-latest-version was invisible in the UI — a row whose exact captured version wasn't available showed today's content with a plain `vN` badge, indistinguishable from an exact match | major | **Yes** | Badge now turns amber with an explanatory tooltip (`"v3 (not the run's v1)"`) when it's a fallback, not an exact match. |
| 3 | The module-level cache stored a rejected fetch's `[]` result permanently — one transient failure would brick a test case's hover for the rest of the session | major | **Yes** | Cache entry is now evicted on rejection so the next hover retries; added a regression test proving retry-after-failure. |
| 4 | The hover's version-aware preview vs. the link's destination (which navigates to the current, non-version-scoped test case page) can show different content | major | No | No versioned test-case-detail route exists in this codebase today — making the destination version-aware is a materially larger, separate feature outside this PR's scope. |
| 5 | Potential race conditions in the fetch-on-active-toggle hook (id changes mid-flight, unmount before resolve) | medium | Verified, no fix needed | The existing `cancelled`-flag + effect-cleanup pattern already handles this correctly (React's cleanup-before-next-effect ordering). Added an explicit regression test (stale id's result never clobbers a newer id's state) to lock this in rather than just assert intent. |
| 6 | Same root cause as #1 — a row's shown version can change as reports stream in | minor | No | Same reasoning as #1. |
| 7 | Unbounded module-level cache growth across a long-lived SPA session | minor | No | Cached values are small per-test-case prompt text; matches this codebase's existing risk tolerance for similar lightweight module caches. LRU eviction would be speculative complexity for a hover-preview convenience feature. |
| 8 | Threading `version` through 3 component layers (table → expanded row → task section) | minor | No | Necessary prop-drilling for correctness; consistent with how this component tree already threads `windowAgentsByRunId`/`spanDeepLink`. |
| 9 | Radix Tooltip renders content twice (visible + visually-hidden a11y announcer), which surfaced as test-selector ambiguity | minor | No (test-only fix applied) | Pre-existing Radix behavior already present in this codebase's other Tooltip usages (e.g. `VersionIndicator`); not user-visible, only affected test selectors (`.first()`). |
| 10 | Testing gaps: rejection-then-retry, id-change-mid-flight | testing | **Yes** | Added both as explicit regression tests (see above). |

No new dependencies (built on the existing `@radix-ui/react-tooltip`, already a project dependency).

TackAdam (approved):

PR #473, opened 2026-09-03
fix(inspector): auto-select first case + compact re-run provenance chip

goyamegh:
## Summary

Two owner-reported papercuts on the run inspector, both traced to `components/evals3/RunInspectorPage.tsx` (and its sibling `EvalRunDetailPage.tsx`, which duplicates the same provenance-chip markup).

**Stacks on #467** (`goyamegh/testcase-page-runs`) — the in-flight inspector-page train (#462/#464/#465/#466) touches `RunInspectorPage.tsx` heavily, so this branches off the stack tip instead of `origin/main` to avoid churn/conflicts. Base branch should be re-pointed to `main` once the stack merges.

### 1. Dead pane on load
The inspector's left list already renders every case with its verdict on mount, and there's already an initial-selection effect that auto-selects the first row (list order — deep-linked `?reportId=` and the previously-selected row on a run switch still win). This was already correct on this stack tip, but had **no regression test** locking in the "never shows the empty placeholder when the list is populated" contract — added one at both the unit and e2e level so a future edit to this effect can't silently reintroduce the dead pane.

### 2. Re-run provenance chip crowding the title
A long source-run name rendered as `re-run of <full name>` with no width bound or truncation — wrapping the pill onto multiple lines and crowding the (already-truncated) run title next to it. Fix: the chip's label is now a bounded, truncated `<span>` (`max-w-[220px] truncate`), so it always stays single-line; the full source name (and the full run title) now only ever surface via a `title` tooltip. Fixed identically in both places this chip is rendered — `RunInspectorPage.tsx`'s split-panel inspector and `EvalRunDetailPage.tsx`'s report page, which independently duplicated the same chip.

## Tests
- **Unit** (`tests/unit/components/evals3/RunInspectorPage.test.ts`): auto-selects the first case and never renders the dead-pane message when the list is populated; the chip stays single-line/truncated with the full source name only reachable via the tooltip.
- **E2E** (`tests/e2e/run-inspector-papercuts.spec.ts`, real browser via `./fixtures/test-fixtures` + `TestDataTracker` cleanup): a case is selected and its detail renders on load and after a hard reload with no "Select a test case"; the chip's bounding-box height proves it never wraps, on both `RunInspectorPage` (`/inspect`) and `EvalRunDetailPage` (no `/inspect`); the chip still navigates to the source run.

## Verification
- `npm run build:all` ✅
- `npm run test:unit` — 5738 passed ✅
- `npm run test:integration` — 838 passed (against a local ephemeral OpenSearch, matching this repo's CI job) ✅
- `npx playwright test tests/e2e/run-inspector-papercuts.spec.ts` — 5 passed, verified against a real server wired to the shared OpenSearch cluster ✅
- `npm audit --audit-level=high` — pre-existing transitive-dep advisories only (react-router/undici/vite), unrelated to this change; CI's audit step is informational (`|| true`), not a hard gate.


goyamegh:
## codex_review (adversarial second opinion — gpt-5.4)

Findings, severity, and disposition:

| 1 | HIGH | Truncated `<span>` inside a flex `<button>` could fail to shrink (flex items default `min-width: auto`) — pill might still overflow instead of ellipsizing | **Applied.** Added explicit `min-w-0` to the label span. Also empirically verified in a real headless-Chrome run against the shared cluster with a ~180-char source name: chip width capped at 220px, span shrinks to fit and ellipsizes, no horizontal overflow — true both before and after the `min-w-0` addition (Tailwind's `truncate` already sets `overflow:hidden`, which triggers the CSS flexbox spec's automatic-min-size-zero rule). Kept the explicit class anyway since it documents the intent and costs nothing. |
| 2 | HIGH | Papercut #1's "fix" is only a new regression test — production auto-select logic was already correct, so this doesn't diagnose a real failure mode | **Accurate, not applied as a "fix."** Confirmed via code read + 34 passing unit tests that list-order auto-selection, deep-link precedence, and reselection-on-run-switch were already implemented and covered by a pre-existing test. No repro of the dead-pane state was found on this stack tip. Framed honestly in the CHANGELOG/PR body as "verified existing behavior + added a regression test" rather than claiming a bug was fixed — if the owner's screenshot reproduces again, it's a different code path than the one investigated here and needs its own diagnosis. |
| 3 | MED | `title={value}` added to the shared `InlineRenameField` affects every consumer (also used by `EvalRunsPage` row names), not just this PR's pages | **Rejected (kept as a shared default).** The component already unconditionally truncates text for every consumer; a hover tooltip for already-truncated text is strictly additive UX with no identified downside for either current consumer (run inspector header, run-list row name). Not made opt-in. |
| 4 | MED | E2E assertions on a hardcoded Tailwind class (`border-l-blue-500`) and a bounding-box height threshold are brittle / implementation-coupled | **Accepted, not changed.** This matches this repo's existing e2e idiom for CSS/layout-class and bounding-box assertions (e.g. `toHaveClass(/bg-muted/)`, row-ordering via `boundingBox().y` in `evalrun-rename-and-sort.spec.ts`). There's no toast/notification layer or other affordance to assert against for a purely-visual "does it wrap" bug; the height check is the least-brittle proxy available and is comment-documented with its basis. |
| 5 | MED | Same provenance-chip markup (and its `sourceRunName`/`sourceRunMissing` resolution effect) is duplicated across `RunInspectorPage.tsx` and `EvalRunDetailPage.tsx` — extract a shared component | **Deferred, not applied in this PR.** Valid and actionable, but `RunInspectorPage.tsx` is already under heavy concurrent edit by 4 other in-flight PRs in the same stack (#462/#464/#465/#466); a component extraction now would meaningfully raise conflict surface for no functional gain to this fix. Left as a follow-up for once the stack settles. |
| 6 | LOW | `title`-only tooltip disclosure is mouse-centric (no touch/keyboard affordance) | **Accepted, not changed.** Consistent with the rest of the app's existing tooltip conventions (no toast/notification system); not a regression introduced by this PR. |
| 7 | LOW | `max-w-[220px]` is a fixed magic number, not responsive | **Accepted, not changed.** Reviewer's own note: acceptable for a one-off pill fix; no responsive-layout requirement here. |

## Verification
- `npm run build:all` ✅
- `npm run test:unit` — 5738 passed, 5 skipped ✅
- `npm run test:integration` — 838 passed, 7 skipped (against a local ephemeral OpenSearch, matching this repo's CI job exactly) ✅
- `npx playwright test tests/e2e/run-inspector-papercuts.spec.ts` — 5/5 passed, run twice (before/after the `min-w-0` follow-up commit) against a real server wired to the shared OpenSearch cluster; zero leaked fixtures confirmed by post-run query ✅
- Direct DOM measurement (headless Chrome) with a ~180-char source-run name confirms the chip never exceeds its 220px cap and the label ellipsizes rather than overflowing, on both `/evaluations/runs/:id` and `/evaluations/runs/:id/inspect` ✅
- `npm audit --audit-level=high` — pre-existing transitive-dep advisories only (react-router/undici/vite), unrelated to this change; this repo's CI audit step is informational (`|| true`), not a hard gate.

TackAdam (approved):

PR #476, opened 2026-09-03
fix(cli): quick mode runs ad-hoc instead of minting a quick-<timestamp> benchmark per invocation; ambiguity-safe -n name reuse

goyamegh:
### Description

Re-running the bare `agent-health benchmark` command (quick mode) created a fresh `quick-${Date.now()}` **Benchmark doc on every invocation**, growing the benchmarks list unbounded — runs are per-invocation, benchmarks are not. This was the remaining un-landed piece of the benchmark-dedup work: #437 shipped the content-addressed images + doctor, but the quick-mode CLI change itself was inadvertently dropped from that PR during a branch repair (the 0.6.0 changelog listed it in error — corrected here with an erratum note).

**Changes**

1. **Quick mode → ad-hoc evaluation run.** Bare `benchmark` (no `-n`/`-f`, no server running) now routes through the existing unified evaluation-runs API as an ad-hoc run over all stored test cases. Zero Benchmark docs created; the run appears on Evaluations → Runs with the standard "promote to benchmark" affordance. Legacy server lifecycle preserved (starts its own server, stops it afterwards — teardown remains ownership-gated in `createServerCleanup`, it only ever stops a process it spawned).
2. **`findBenchmark` name reuse, ambiguity-safe.** Resolution is now: exact id → exact name → **unique** trimmed/case-insensitive match, so `-n "my bench "` reuses `My Bench` instead of minting a duplicate. A new `findBenchmarkDetailed` distinguishes *not found* from *ambiguous*: when 2+ benchmarks collide case-insensitively (`Foo` + `FOO`), the `-n` flows **error listing the collisions** — the CLI neither guesses nor creates a third near-duplicate under the colliding name.

**Behavior changes (intentional):** quick mode no longer leaves a `quick-*` benchmark behind (use run promotion instead); `-n` with a case/whitespace variant of exactly one existing benchmark now reuses it instead of creating a new one.

### Testing

- Unit: quick-mode action pins the new contract (ad-hoc POST with `test-case-ids` sources, **no** `createBenchmark` call); `findBenchmark`/`findBenchmarkDetailed` matrix (exact-over-fuzzy, unique fuzzy reuse, ambiguous → surfaced collisions). Full unit suite: 5,705 passed.
- Integration (live server): two identical ad-hoc runs leave benchmark entities unchanged — runs carry no `benchmarkId`, no benchmark references the seeded test case, no new `quick-*` benchmark appears.
- Manual e2e: bare `benchmark -a demo` twice → 0 Benchmark docs, 2 ad-hoc runs (server stamps both with the same image digest, confirming the dedup convergence shipped in #437).

An adversarial model review (gpt-5.4) was run on the diff; its findings (ambiguity-created duplicates in the create-if-missing `-n` path, teardown ownership) were addressed in this revision — ambiguity is now a hard error, and teardown was verified ownership-gated.

### Check List
- [x] New functionality has been documented (CHANGELOG + erratum)


goyamegh:
**Re: Code-Diff-Analyzer high finding on `package-lock.json` — the finding is inverted; this change *restores* the authoritative scope.**

`origin/main`'s `package-lock.json` currently carries `"name": "@goyamegh/agent-health"` — a fork-side package rename (used only for fork npm publishing) that leaked upstream via the #437 branch repair. `package.json` on `origin/main` correctly says `@opensearch-project/agent-health`, so the lockfile has been inconsistent with the manifest since #437 merged. Regenerating the lockfile restores the organization-controlled `@opensearch-project` scope; there are no dependency changes (3 insertions / 3 deletions, name fields only — split into its own commit for easy review).

In other words: the analyzer is right that scope changes matter, but the hijack-shaped state is the one currently on `main`, and this PR fixes it.

TackAdam (approved):

PR #479, opened 2026-09-04
feat(evals3): benchmark Runs tab as compact table + pass-rate-over-time chart with click-to-filter pills

goyamegh:
## Summary

The per-benchmark **Runs** tab (`/evaluations/benchmarks/:id/runs`) was a stack of tall cards — one ~140px card per run with the case-verdict heat strip always expanded — which fit ~3 runs per screen and gave no way to see how agents compare over time. It also had a ~400px blank band above the list (see "Also fixed" below).

It is now a **compact table with a pass-rate-over-time chart on top**:

- **Table columns:** Run (link to the inspector) · Agent · Model · Size · Pass % · Judge (evaluator) · J. Model (judge model) · Date. Rows are ~30px. Every column sorts (newest-first default; null pass rates always sink).
- **Chart:** one line per agent, real time X-axis, 0–100% Y-axis. Legend entries are clickable.
- **Click-to-filter with pills:** clicking any Agent / Model / Judge / J. Model cell, a `Running`/`Cancelled` status badge, or a chart legend entry toggles a filter. Active filters render as removable pills (`Agent: …`, `Judge: …`, `J. Model: …`) with `Clear` and an `N of M runs` count. Values on the same field OR together; different fields AND together. Agent filters *dim* the other chart lines (so a second agent can still be toggled back in from the legend); every other filter narrows the chart to exactly the rows the table shows. Filters are session-scoped by design.
- **Compact:** page padding `p-3/p-4` (was `p-4/p-6`), header `text-xl`, description clamped to 2 lines, same-year dates drop the year. Pass % is colour-graded (≥80 / ≥50 / <50) with the `passed/failed[/⚠errored][/pending]` breakdown beside it and a tooltip stating the denominator (judged = passed+failed; errored/pending excluded per #242).
- **Heat strips are still there** — behind a per-row expand chevron instead of always occupying space.

Existing behaviour preserved: Compare checkbox selection (Select All now acts on the visible/filtered rows), Delete/Cancel gating for standalone eval-run docs, version filter + its empty state, `Latest` badge, polling while a run is in flight.

### Also fixed
- **Blank band above the runs list.** Radix hides the inactive `TabsContent` with the `hidden` attribute, but the `flex flex-col` class on the Cases panel (needed for #447's scrolling fix) out-specifies the UA `[hidden]{display:none}` rule — so the inactive Cases panel stayed `display:flex` with `flex-1` and pushed the Runs panel ~400px down. `data-[state=inactive]:hidden` fixes it.
- **`Latest` badge** was `filteredRuns[0]`; the merged list appends standalone eval-run docs *after* embedded runs, so a newer CLI/API-started run never got it. Now `latestRunId()` = max `createdAt`.
- `tests/unit/components/evals3/BenchmarkRunsPage.test.ts` gets the fuller router mock it needs (`useLocation`/`Link`) and drops the coarse `CaseHeatStrip` stub (the heat-strip expand test needs the real one; the ESM `react-markdown` chain is neutralised via the `@/components/ui/markdown` mock instead).

## Tests

| Level | File | Covers |
|---|---|---|
| Unit | `tests/unit/lib/benchmarkRunsTable.test.ts` (22) | pass-rate derivation excludes errored/pending; filter AND/OR + toggle; sort incl. null-sinking; per-agent series ordering is stable under new runs; un-judged runs dropped from chart; `latestRunId` |
| Unit | `tests/unit/components/evals3/BenchmarkRunsTable.test.ts` (11) | cell click filters without navigating; pill labels; action gating; expand; aria-sort |
| Unit | `tests/unit/components/evals3/BenchmarkRunsPage.test.ts` (15) | columns, default sort, `Latest` via createdAt, filter pills, legend toggle, heat-strip expand, inactive-panel regression |
| Integration | `tests/integration/server/routes/storage/benchmarkRunsTableFields.integration.test.ts` (4, real backend) | `GET /api/storage/benchmarks/:id` round-trips `judgeModelId`/`evaluatorId`/`testCaseSnapshots`/per-result verdicts the columns depend on, incl. under `?runsSize=` pagination and `fields=polling`; stale `run.stats` ignored |
| E2E | `tests/e2e/benchmark-runs-table.spec.ts` (5, new) | headers exactly as designed; chart above table with one legend entry per agent; rows ≤40px and no blank band; click→pill→remove; AND/OR + legend toggling + Clear; run link → inspector; heat-strip expand |
| E2E | `benchmarkruns-associated-evalruns.spec.ts`, `benchmark-runs-passed-count.spec.ts` | re-targeted from card selectors to the table's `data-testid`s |

All of the above green locally against a server built from this branch's head, wired to a real OpenSearch cluster (test data tracked + cleaned by id).

## Adversarial review (`codex_review`, gpt-5.4)
Applied: stable chart colours under polling (series were ordered by point count) · `Latest` by `createdAt` · tooltip renders every series at the hovered x (was `payload[0]`) · dropped the untyped `activeDot.onClick → payload.runId` navigation and a dead `allowDuplicatedCategory` prop on a numeric axis · Pass % denominator tooltip.
Rejected: including errored cases in the Pass % denominator (contradicts the repo-wide #242 convention every other surface follows) · making agent pills narrow the chart like other fields (deliberate — keeps the legend usable as a toggle, documented inline) · removing row-click navigation (pre-existing, covered by existing e2e) · folding the `running` back-solve into `computeRunStats` (moved verbatim from the previous renderer; `computeRunStats` doesn't expose `running`).

## Checklist
- [x] DCO signoff on every commit
- [x] `npm run build:all` · `npm audit --audit-level=high` clean
- [x] Unit + integration + e2e added (see table)
- [x] SPDX headers on new files


goyamegh:
## Verification

Built from this branch's head and run against a real OpenSearch storage cluster (`AH_PORT=4361`, run-recovery disabled so it can't touch other servers' in-flight runs). All test data is created with `uniqueTestName()` and deleted by id via the tracker.

**Unit** (`npx jest tests/unit/lib/benchmarkRunsTable.test.ts tests/unit/components/evals3/BenchmarkRunsPage.test.ts tests/unit/components/evals3/BenchmarkRunsTable.test.ts`)
```
Tests:       48 passed, 48 total
```

**Integration** (`AH_PORT=4361 npx jest tests/integration/server/routes/storage/benchmarkRunsTableFields.integration.test.ts`)
```
✓ round-trips judgeModelId / evaluatorId / testCaseSnapshots / results on embedded runs
✓ the table row derived from the real response shows Size=2, Pass %=50 (ignoring stale stats), Judge + J. Model ids
✓ preserves those fields under ?runsSize= pagination (what the page actually requests) and reports totals
✓ fields=polling (the in-flight poll request) still carries the verdict + judge fields the table re-derives on each poll
Tests:       4 passed, 4 total
```

**E2E** (`PLAYWRIGHT_SKIP_WEBSERVER=1 npx playwright test tests/e2e/benchmark-runs-table.spec.ts tests/e2e/benchmarkruns-associated-evalruns.spec.ts tests/e2e/benchmark-runs-passed-count.spec.ts`, then the new spec again with `--repeat-each=2`)
```
7 passed (13.1s)
10 passed (21.1s)
```

**Scripted interaction check against a real 28-run benchmark** (Playwright script, not a test — 3 agents, mixed running/cancelled/completed):
```
rows before filter:            27
click Agent cell on row 1  →   6 rows, pills: ["Agent: <agent A>"]
click Judge cell on row 1  →   6 rows, pills: ["Agent: <agent A>", "Judge: <evaluator>"]
legend entry aria-pressed:     1
expand row 1               →   1 heat strip rendered
click first pill (remove)  →   16 rows
Clear                      →   27 rows
```
Rows measure ~30px tall (were ~140px cards); the blank band between the tab strip and the list is gone (chart top sits <60px below the tabs).

**CI note:** the `e2e-tests` job on this PR fails on the same 5 specs that fail on `main`'s own latest push CI (run 33817338957 @ 48a463af8): `benchmark-run-row-link` ×2, `benchmark-version-link-panel`, `evalruns-no-left-status-icon`, `judge-tab-whyfix`. None of the specs added/changed here are among them; one new-spec test was flaky on its first attempt under worker contention (page loader bounced on a slow backend GET) and is now routed through a one-retry `openRunsTab()` helper.

TackAdam (approved):

goyamegh:
**Heads-up for reviewers / compose seam with #468 (`goyamegh/run-actions`)**

This PR and #468 both touch `components/evals3/BenchmarkRunsPage.tsx`. In isolation both merge cleanly onto `main`; composed together they conflict on two import lines only:

- `Checkbox` (old card list, removed here) vs `JudgeModelSelect` (inline Configure-Run form, replaced by #468's shared `RunConfigDialog`).
- `RunStats` vs `Evaluator` in the `@/types` import.

Resolution used on the integration branch (whichever PR lands second should apply the same):

1. Drop both `Checkbox` and `JudgeModelSelect` imports — the Runs tab is this PR's `BenchmarkRunsTable` + `BenchmarkPassRateChart`; the run form is #468's shared `RunConfigDialog` (with Concurrency, reused by Re-run).
2. Keep the `Evaluator` import and the page-level `evaluators` fetch — the table's Evaluator column uses it to label rows by name (`evaluatorLabel`); the dialog owns its own list for the picker.
3. Row actions stay as this PR's inline Delete/Cancel (`useBenchmarkCancellation` → `POST /benchmarks/:id/cancel`, which #468 hardened with the zombie-cancel fallback). #468 deliberately keeps its kebab (`RunActionsMenu`) off this page, so there is one action surface per row, not two.
4. Cases tab (#447) is untouched.

Verified on the composed tree: `tsc` clean; unit suites for `BenchmarkRunsPage`, `BenchmarkRunsTable`, `RunConfigDialog`, `benchmarkRunsTable`, `runActions` green (88/88); e2e `benchmark-runs-table`, `benchmark-runs`, `benchmark-runs-passed-count`, `run-actions-menu` green.

goyamegh:
### Follow-up pushed: cancelled / failed runs never render as "in progress" (owner bug, 2026-09-08)

**Symptom (live tunnel, benchmark Runs tab):** a run cancelled at 43/62 showed the **Cancelled** badge next to `11/32 /19 ⟳` — a pending spinner that never went away. DOM audit of one real benchmark: **21 of 45 rows** carried the spinner (16 cancelled + 3 failed + the 2 genuinely running ones).

**Root cause (this PR's row builder, `lib/benchmarkRunsTable.ts`):** `pending = total − passed − failed − errored − running` regardless of the run's TERMINAL status, so the planned-but-never-started remainder was re-derived as pending forever. The embedded `benchmark.runs[]` projection and the standalone evaluation-run docs were verified **in sync** on the live cluster (25/25 embedded runs: same status + same result count as their docs) — purely a display bug, no dual-write drift.

**Fix (`819eb4f25`, `f64e3b196`):**
- `buildRunTableRow()` is terminal-aware: once `run.status ∈ {cancelled, failed, completed}` the remainder — plus any per-case entry a dead executor left `running` — is a new `notRun` bucket; `pending`/`running` are 0 by definition. Invariant `total = passed+failed+errored+pending+running+notRun`. Only the **explicit** persisted status gates this (the results-derived legacy fallback is display-only — it reports `completed` for a status-less run between two cases). Mirrors #486's `TERMINAL_RUN_STATUSES`; collapses onto `isTerminalRunStatus` once that lands.
- Table: `· n not run` (muted, no spinner); **Failed** badge (click-to-filter, like Running/Cancelled); Cancelled badge carries a `Ban` icon + tooltip with the not-run count.

**Red → green:** the 6 new/changed unit + RTL assertions fail on `c4e28e070` and pass on this head.

```
tests/unit/lib/benchmarkRunsTable.test.ts            28 passed  (status × partial-results matrix, dead-executor `running` entry,
                                                                 per-case cancelled markers, status-less legacy run NOT terminal,
                                                                 status filter for running/cancelled/failed)
tests/unit/components/evals3/BenchmarkRunsTable.test.ts  12 passed  (cancelled/failed/live rows: badges, "n not run", no spinner)
tests/e2e/benchmark-runs-table.spec.ts                6 passed  (embedded + standalone-doc cancelled runs on the real page:
                                                                 badge, "2 not run", planned size, pass rate over judged, no spinner,
                                                                 badge filters to cancelled)
```

**After, on a local build against the shared cluster (read-only), same benchmark — DOM audit of all 45 rows:**
```
{"n": 45, "spinners": 0, "runningBadges": 0,
 "cancelled rows": 16 → e.g. "11/32 · 19 not run", "2/3 · 57 not run", "15/25/12 · 10 not run"
 "failed rows":    3  → e.g. "0/5/⚠32 · 25 not run", "18/34 · 10 not run"}
```

Composes cleanly with #468 and #486 (verified: origin/main + #468 + this + #484 + #486 → `tsc` clean, seam suites green).

**Adversarial review (`codex_review`, gpt-5.4):** 10 findings. Applied: status-filter coverage for the new `failed` badge value; per-case `cancelled`-marker invariant test; comment pointing the local terminal check at #486's shared constant. Rejected with reason: "status-less legacy terminal runs still show pending" — by design, the legacy fallback can't distinguish in-flight from terminal, so it keeps the pre-existing rendering (live data: 35 status-less embedded runs, **0** with planned > observed); "run.status may flip terminal before per-case writes land" — inherent to main's cancel route, fixed by #486's cancel-is-a-request contract in the same compose; "per-case cancelled markers bucket as failed on main" — no writer on main produces them (only #486's finalize, which also fixes the bucketing); "stranded `running` entries ≠ not run" — matches #486's `bucketRunResults` convention; "two status authorities" — same split #486's `computeRunStats` makes, documented in-code.

Full e2e suite on this branch has the same pre-existing failures as `origin/main` (run-inspector `h2` vs inline-rename, edit-versioning checkbox flow) — being fixed in #487; verified identical against the unmodified `c4e28e070` specs on the same server.

goyamegh:
**Follow-up commit `d92abc170` — Delete on every row of the Runs tab, dispatched per run kind** (owner report: *"Delete button should be present for all runs on the benchmark details page"*)

Before: rows backed by a standalone evaluation-run doc (not embedded in `benchmark.runs[]` — the common case since the dual-write; 27 of 48 runs on one live benchmark) rendered **no** Delete/Cancel, because the row actions only knew the benchmark nested-run API (which 404s for a non-embedded run) and the page hid them via an `actionsDisabledIds` gate.

After: the gate is gone. Every row has Delete; running rows have Cancel. `handleDeleteRun` / `handleCancelRow` dispatch on the run's kind — evaluation-run docs → `DELETE /api/storage/evaluation-runs/:id` / `POST …/cancel`, legacy embedded-only → the benchmark nested-run endpoints. Removing both persisted forms of a dual-written run is server-side (#492), not reconstructed client-side. The confirm names the run, flags a running one, and states that per-test-case reports are kept.

Red → green:

```
# unit — BenchmarkRunsTable.test.ts + BenchmarkRunsPage.test.ts: 30 passed
#   (every row offers Delete; only running rows offer Cancel; kind → API dispatch for
#    standalone / dual-written / legacy rows; Cancel dispatch; confirm copy; decline deletes nothing)

# e2e, tests/e2e/benchmark-runs-delete-all-rows.spec.ts — before (old client):
✘ every row (standalone doc, legacy embedded, dual-written) renders a Delete button
✘ deleting a STANDALONE (non-embedded) row hits the evaluation-runs API, removes the row, and the doc is gone
✓ deleting a LEGACY embedded-only row hits the benchmark nested-run API and removes it from benchmark.runs[]
✘ declining the confirm keeps the row and issues no DELETE
# after: 4 passed; benchmarkruns-associated-evalruns.spec.ts flipped (Delete + Cancel ARE offered on a merged-in running row); benchmark-runs-table.spec.ts (6) still green
```

`codex_review` (gpt-5.4) on an earlier draft of this commit flagged that a client-side "delete the doc, then also clear the projection" fallback would report success on a partial delete and depend on a best-effort listing — applied by removing it: the page now makes exactly one call per row, and the server owns removing both forms (#492).

PR #480, opened 2026-09-04
fix(connectors): per-invocation session capture, evidence-preserving trajectory merge, real tool outputs for subprocess agents

goyamegh:
### Description

Three correctness bugs surfaced by a retrieval audit of real runs. Each one corrupted what the LLM judge was shown, so all three are fixed together (they compound: bug 1 makes the poller fetch the wrong case's spans, bug 2 then throws away the connector's evidence in favour of those spans, bug 3 meant the connector's own evidence wasn't marked as evidence in the first place).

**1. P0 — concurrent runs of a subprocess agent swapped each other's `sessionId` (cross-run data corruption).**
The registry hands out ONE connector instance shared by every in-flight evaluation task. `ClaudeCodeConnector` kept the captured stream-json `session_id` in an instance field (re-stamped on every data event). Under `concurrency > 1`, a sibling child's stdout landing between a run's last chunk and its `close` re-pointed the field, so the finished run's report carried the *sibling's* `sessionId` — and the trace poller (Strategy D, `session.id` correlation) then fetched the sibling case's spans and rebuilt the trajectory from them: **cases were judged on another case's tool calls** (measured: 11/62 reports of one concurrent run cross-wired). The same instance-state pattern covered every streaming buffer in Claude Code / Pi / Kiro / the base class.
Fix (structural): `SubprocessConnector.execute()` creates a per-invocation `SubprocessExecutionState` (subclasses extend via `createExecutionState()`), threaded through `parseStreamingOutput` / `parseStderrChunk` / `onBeforeStreamEnd` / `extraResultMetadata`. No parse hook keeps anything on `this`.
Hardening (from codex review): the pre-existing mutate-`this.config`-then-restore-in-`finally` pattern for per-execution options (base, Claude Code, Pi) was a *second* shared-state race — `data`/`close` callbacks read `this.config` long after the first `await`. Replaced with a pure `resolveExecutionConfig(request)` hook producing an immutable per-call snapshot; `this.config` is never written during execution.

**2. P0 — the trace poller discarded the connector's tool evidence when it replaced the trajectory (judge evidence loss).**
With no `buildTrajectory` hook, the default `spansToTrajectory()` output REPLACED the connector trajectory and only its `response`/`assistant` steps were appended back — the connector / `afterResponse`-hook `tool_result` steps (the ones with `toolOutput`) were dropped. When an agent's spans carry no tool payloads that left the judge with nothing: ~50% of a REST agent's reports ended with **zero tool steps** (the judge literally wrote "no intermediate retrieval steps"), and a subprocess agent's reports ended with 7 × `tool_result: "tool succeeded"` stubs while `rawEvents` held the full 28 KB results.
Fix: replacement is now an evidence-preserving merge (`services/traces/trajectoryMerge.ts`, policy documented in the module header): an explicit hook still wins wholesale (#320); otherwise, if the span trajectory would lose tool evidence — fewer `action`/`tool_result` steps, or fewer bytes of `tool_result` evidence — the connector trajectory is kept (span `latencyMs` folded on only when the tool sequences are provably the same calls); if the spans are at least as complete they win (trace-only agents keep working) and the connector's answer is appended when the spans have no response content. Tool-step count and evidence bytes never decrease.

**3. P1 — subprocess-agent tool results lost their tool name and output (evidence fidelity).**
Stream-json `tool_result` blocks became bare `tool_result` steps (no `toolName`, no `toolOutput`, content-block arrays JSON-encoded verbatim). Results are now paired with their `tool_use` by `tool_use_id`, the text the model actually saw is flattened from string / `text`-block / `tool_use_result` shapes and stored on `content` + `toolOutput`, bounded at 32 KB with an explicit `[tool output truncated: showing N of M chars]` marker (full stream stays in `rawEvents`). The judge compactor now serializes a result once when `content` duplicates `toolOutput`.

### Issues Resolved

Found by a retrieval audit against a 62-case internal retrieval benchmark (no tracking issue; the fixes gate a fairness-critical re-run of that benchmark).

### Testing

Every fix carries red→green regression tests at unit + integration level:

| Bug | Unit | Integration |
|---|---|---|
| 1 sessionId swap | `ClaudeCodeConnector.test.ts`: deterministic two-child interleaving (A data, A result, B init, A close) — pre-fix `Expected "sess-A", Received "sess-B"`; interleaved partial-NDJSON buffering; concurrent executions with different `connectorConfig`s spawn with their own args/env/cwd/input mode and `this.config` stays byte-identical before/during/after | `concurrentSessionIsolation.integration.test.ts`: real `executeEvaluationRun` at concurrency 2 → real `spawn` of a paced fake CLI → real `FileStorageModule`; both persisted reports carry their own `sessionId` and un-mixed trajectories (red on the old connector with the same `sess-A`/`sess-B` swap) |
| 2 evidence loss | `trajectoryMerge.test.ts` (both live shapes + invariants: tool-step count and evidence bytes never decrease, inputs not mutated, name-gated timing fold); `tracePoller.test.ts` through the real `poll()` path (no-tool spans, stub spans, hook still wins) | `restEvidenceMerge.integration.test.ts`: real runner → real REST connector against a local HTTP agent → hook-built steps → real poller with payload-less spans → the persisted AND judged trajectory keeps both `tool_result` steps with their `toolOutput` (old poller persisted zero tool steps) |
| 3 tool outputs | `ClaudeCodeConnector.test.ts`: synthesized stream-json fixture delivered in arbitrary byte chunks — pairing, flattening, truncation marker, event-level fallback, orphan result; `bedrockService.test.ts`: compactor dedupe | covered by the integration test for bug 1 (paired `Read` result carries its output) |

Gates: `npm run build:all` ✅ · `npm run test:unit -- --coverage` ✅ (360 suites / 6152 tests, 84.4% lines) · `npm run test:integration` against a server on the shared cluster ✅ (only failures are environment-specific: a spec that needs port 4361 free, which another local server holds, and CLI specs that refuse to reuse a running server under `CI=1`) · `npm run test:e2e` run against the same server (UI untouched by this PR; see proof comment) · `npm audit --audit-level=high`: pre-existing advisories only, no dependency changes · DCO on all commits · CHANGELOG updated.

### Check List
- [x] New functionality has been documented (CHANGELOG, `AGENTS.md` trace-correlation section, connector skill/docs mention the per-invocation `state`).
- [x] API changes: `SubprocessConnector` protected hooks gain a trailing `state` argument (public SDK surface — legacy subclasses calling `super` without it are tolerated); `parseResponse()` gains an optional resolved-config argument.
- [x] Public documentation issue/PR created: n/a (no user-facing API docs change).


goyamegh:
## Verification proof (text-only)

### Root causes confirmed on `origin/main` @ 48a463af8

| Bug | Where | What we saw on real runs |
|---|---|---|
| 1 sessionId swap | `services/connectors/claude-code/ClaudeCodeConnector.ts:93` (`private sessionId`) set at `:179`, read at `:489` from `extraResultMetadata()` — instance state on the registry singleton | two concurrent runs of a subprocess agent: 11/62 reports carried a sibling case's `sessionId`; Strategy D then fetched the sibling's spans |
| 2 evidence loss | `services/traces/tracePoller.ts:425-439` replaced `report.trajectory` with the span-derived steps + only the original `response`/`assistant` steps | a REST agent whose spans carry only prompt/completion: 30/62 reports reduced to `[thinking, response]` (zero tool steps); a subprocess agent: 7 × `tool_result: "tool succeeded"` (14 chars) while `rawEvents` held the real 32–2005-char results |
| 3 no tool output | `ClaudeCodeConnector.parseJsonEvent` `user`/`tool_result` branch emitted content-only steps (no `toolName`, no `toolOutput`) | `report.rawEvents[].data` stdout chunks (105 chunks, 2 not newline-terminated) re-split into 107 NDJSON lines contain every result; the persisted trajectory had none |

### Red → green

Same test set run with the `services/` files checked out from `origin/main` (RED), then at this PR's HEAD (GREEN):

```
# RED — services/connectors + tracePoller from origin/main, new tests present
✕ two interleaved executions each return their OWN session id (no cross-wiring)
    Expected: "sess-A"   Received: "sess-B"
✕ interleaved partial NDJSON chunks are buffered per execution, not per instance
✕ never writes per-execution options onto the shared this.config (not even transiently)
    Received: {...,"args":[..., "--append-system-prompt","p"],"env":{...,"X":"1"},"inputMode":"arg","timeout":123,"workingDir":"/tmp/x"}
✕ pairs each tool_result with its tool_use and populates toolName + toolOutput
    Expected: "Read"   Received: undefined
✕ bounds oversized outputs with an explicit truncation marker carrying the full length
✕ falls back to the event-level tool_use_result when the block has no content
✕ an unmatched tool_result (unknown tool_use_id) still keeps its output, without a toolName
✕ [poller] keeps hook-built tool_result steps (with toolOutput) when the spans carry NO tool payloads
    Expected length: 1   Received length: 0   Received array: []
✕ [poller] keeps connector outputs over span-derived "tool succeeded" stubs and folds span timing onto them
    Expected: not "tool succeeded"
✕ [integration] two concurrent runs persist distinct, correct sessionIds and un-mixed trajectories   (real spawn, real file storage)
    Expected: "sess-A"   Received: "sess-B"
✕ [integration] persists the judged trajectory WITH the hook-built tool_result steps and their toolOutput   (real runner → REST → poller)
    - Expected ["get_index_profile", "dsl_executor"]   + Received []
Tests: 12 failed, 83 passed, 95 total

# GREEN — this PR's HEAD (a77e6532)
Test Suites: 6 passed, 6 total
Tests:       159 passed, 159 total
```

### Gates

```
npm run build:all                       ✅
npm run test:unit -- --coverage         ✅  Test Suites: 360 passed | Tests: 6152 passed, 5 skipped | lines 84.35% (threshold 70)
npm run test:integration (AH_PORT=5023, server on the shared OpenSearch cluster, recovery disabled)
                                        ✅  884 passed / 7 skipped; the only failures are environment-specific and pass in isolation
                                            (a spec needing port 4361 free — another local server owns it; CLI specs under CI=1 refuse to reuse a running server)
npm audit --audit-level=high            pre-existing advisories only — this PR changes no dependencies (identical result on main)
DCO                                     ✅ all 4 commits signed
CHANGELOG                               ✅ [Unreleased] → Fixed (+ hardening round)
```

### codex_review (adversarial second opinion, different model family) — applied / rejected

|---|---|---|---|
| 1 | high | `execute()` still mutated shared `this.config` (args/env/timeout/cwd) and restored in `finally` — a second shared-state race | **Applied** — pure `resolveExecutionConfig(request)` → immutable per-call snapshot; `this.config` never written; regression test asserts it is byte-identical mid-execution |
| 2 | high | hook returning `[]` should clear the trajectory | **Rejected** — by the hook contract an empty/null result means *trace not ready → keep polling*; handled before the merge, unchanged behaviour |
| 3 | high | merge is still a whole-trajectory chooser (drops connector `thinking`/`assistant` when spans win) | **Rejected** — that branch only fires when the connector has NO tool steps, i.e. the tool-call-less AG-UI narrative #320 replaced on purpose; the answer is still appended |
| 4 | high | by-position `latencyMs`/`toolName` fold can attribute the wrong span | **Applied** — fold only when type pattern AND tool names agree; names are never copied across |
| 5 | high | payload written to both `content` and `toolOutput` doubles the judged JSON | **Applied** (in the judge compactor, so it also fixes REST/generic-span steps that already duplicated) — `content` collapses to `[see toolOutput — N chars]` when byte-identical |
| 6 | med | count mismatch should be a soft signal | **Rejected** — a lower span tool-step count is exactly the measured evidence-loss shape; keeping the connector is the conservative choice, spans stay on `report.spans` |
| 7 | med | evidence-bytes heuristic; `{}`/`[]` discounted | **Applied** the empty-result part (`{}`/`[]` now count as evidence); bytes stay as the tiebreak (14-char stubs vs multi-KB results is the real-world gap) |
| 8 | med | `pendingToolUses` overwrites duplicate ids | **Rejected** — tool-use ids are unique per message in the stream-json format and entries are consumed on pairing |
| 9 | med | protected hook signature change breaks third-party subclasses | **Applied** — base `parseStreamingOutput`/`onBeforeStreamEnd` tolerate a missing `state` (legacy `super.…(chunk, trajectory, onProgress)` callers) |
| 10 | low | dead `resultMetadata` bucket; unnecessary exports | **Applied** — removed; merge helpers trimmed to what callers use |

goyamegh:
### e2e-tests status — pre-existing, identical on `main`

The `e2e-tests` job on this PR fails **5 specs** (`benchmark-run-row-link` ×2, `benchmark-version-link-panel`, `evalruns-no-left-status-icon`, `judge-tab-whyfix`) with 518 passed / 4 flaky. The most recent `main` CI run at this PR's base (`48a463af`, [run 33817338957](https://github.com/opensearch-project/agent-health/actions/runs/33817338957)) fails the **same 5 specs** (519 passed / 3 flaky) — as do the two `main` runs before it. This PR changes no frontend files (`git diff --stat origin/main..HEAD -- components hooks '*.tsx'` is empty); the e2e breakage is upstream and independent of this change.

Full local Playwright pass against a server built from this branch (shared OpenSearch cluster, `PLAYWRIGHT_SKIP_WEBSERVER=1`): 312 passed before the 50-min wall-clock cap while the integration suite was hammering the same server; the failed spec files re-run in isolation reproduce only the same UI-flow failures seen on `main` (dashboard first-run hero text no longer exists in `Dashboard.tsx`, benchmark editor wizard flows, etc.).

TackAdam (approved):

PR #481, opened 2026-09-04
fix(judge): trace judge degrades to trajectory-only for non-instrumented agents; surface run-level judge failures

goyamegh:
## Problem

A real run of a **62-case benchmark** against a **REST agent without tracing** (`useTraces: false` — its trajectory is built by an `afterResponse` hook, not from OTel spans) with `judgeModelId: 'agent-trace-judge'` ended with **all 62 cases `status: 'failed'`, `passFailStatus: null`**. The run-level UI showed only a bare `⚠ 62` errored count. The real cause was buried per-report:

> `Evaluation failed: Bedrock Judge validation error (not retryable): The agent (trace) judge provider needs a runId or at least one trace correlation hint (agents: serviceName+window, or sessionId) to scope its trace tools — this request had neither.`

### Root cause chain
1. `services/evaluation/index.ts` (STANDARD MODE judge call) forwarded `runId=undefined` (REST connectors never mint one) and `agents=[]` — `resolveAgentServiceName` correctly refuses to guess a `service.name` for `rest` (see `UNKNOWN_SERVICE_NAME_PROTOCOLS`). **A non-instrumented agent can never be trace-correlated — by design.**
2. `server/routes/judge.ts` (`provider === 'agent'`) treated "no correlation handle" as a **validation error** → HTTP 400.
3. `services/evaluation/bedrockJudge.ts` marks 4xx `nonRetryable` → thrown.
4. `runEvaluationWithConnector` had a single outer `catch` around **both** agent invocation and judge call → judge failure landed as `status: 'failed'`, no `metricsStatus` — indistinguishable from an agent crash, bucketed as *failed* (not *errored*), and **invisible to #462 Retry judgement** (`isJudgeFailedCase` required `metricsStatus: 'error'`).
5. Nothing aggregated the per-report reason to the run level.

## Fix

**1. Graceful degradation (never a 400 for lack of correlation)** — `server/routes/judge.ts`, `server/services/piAgenticJudgeService.ts`
- `traceToolsAvailable = hasTraceCorrelation(runId, agents)` now selects the judge **mode**, not whether the judge runs.
- `trajectory-only` mode: no `query_spans`/`query_logs` tool pack (`tools: []`, extension not registered) and a new system-prompt addendum (`NO_TRACE_TOOLS_ADDENDUM`) telling the model no trace tools exist for this run and not to claim span/log checks it never performed.
- Resolved mode returned as `judgeMode: 'trace-tools' | 'trajectory-only'` and **persisted as `report.judgeMode`** on every judge call site (standard, trace-poll, retry-judgement, browser recovery) so comparisons can show which verdicts had trace evidence.

**2. Judge-step failures separated from agent failures** — `services/evaluation/index.ts`
- The judge call gets its own `try/catch`; failure returns the canonical `buildEvaluatorErrorPatch('judge_failed', …)` shape (`status: 'completed'`, `metricsStatus: 'error'`, `passFailStatus: null`, trajectory preserved) — buckets as *errored* and is Retry-judgement-eligible. Genuine agent-invocation failures keep the existing outer-catch shape.

**3. Run-level surfacing** — `lib/judgeFailureSummary.ts`, `types/index.ts`, runners, `EvalRunsPage`, `RunInspectorPage`
- `computeJudgeFailureSummary` aggregates the dominant per-case judge-failure reason when ≥50% of a run's cases failed at the judge step → `run.judgeFailureSummary` (e.g. `62/62 cases failed at the judge step: …`). Rendered as an inspector banner and the runs-list errored-badge tooltip. Computed at run completion (evaluation runs, benchmark execute) and on `refresh-stats`, so **existing runs get it without re-running**. Recognizes both the canonical shape and the legacy pre-fix shape.

**4. Retry judgement salvages pre-fix runs** — `services/evaluation/retryJudgement.ts`
- `isJudgeFailedCase` also recognizes the legacy shape (agent completed / trajectory present / `status: 'failed'` / reasoning names the judge), so the existing 62-case run can be re-judged at judge cost only.

## Tests (red → green)
| Level | File | Cases |
|---|---|---|
| unit | `tests/unit/server/routes/judge.test.ts` | 2 rewritten (400 → degrade, `traceToolsAvailable=false`), 3 updated for 3rd arg |
| unit | `tests/unit/server/services/piAgenticJudgeService.test.ts` | +4 (trajectory-only addendum, default back-compat) |
| unit | `tests/unit/lib/judgeFailureSummary.test.ts` | +14 (new) |
| unit | `tests/unit/services/evaluation/bedrockJudge.test.ts` | +2 (`judgeMode` forwarding) |
| unit | `tests/unit/services/evaluation/retryJudgement.test.ts` | +4 (legacy-shape selection) |
| integration | `tests/integration/server/routes/judgeAgentProviderHints.integration.test.ts` | 2 rewritten (real HTTP 400 → 200 + `judgeMode`) |
| integration | `tests/integration/services/evaluationRestJudgeDegradation.integration.test.ts` | +3 (new: verdict + `judgeMode`; judge failure → `metricsStatus:'error'` + `isJudgeFailedCase` true; agent failure unchanged) |
| integration | `tests/integration/server/routes/storage/evaluationRuns.retryJudgement.integration.test.ts` | +1 (legacy-shape report salvaged over real HTTP; pre-fix `total` was 0) |
| e2e | `tests/e2e/judge-failure-summary.spec.ts` | +2 (new: inspector banner, list tooltip; imports `./fixtures/test-fixtures`) |

`tsc --noEmit` clean. Changelog under `## [Unreleased] → Fixed`.

### DIFF
--- a/components/evals3/EvalRunsPage.tsx
+++ b/components/evals3/EvalRunsPage.tsx
@@ -733,7 +733,12 @@ export const EvalRunsPage: React.FC = () => {
             {rr.errored > 0 && (
               <span
                 className="flex items-center gap-0.5 text-amber-500 font-medium ml-0.5"
-                title="Evaluator could not run on these (e.g. judge validation error). Excluded from pass-rate aggregation."
+                data-testid="run-row-errored-badge"
+                title={
+                  rr.run.judgeFailureSummary
+                    ? `Judge failure: ${rr.run.judgeFailureSummary}`
+                    : 'Evaluator could not run on these (e.g. judge validation error). Excluded from pass-rate aggregation.'
+                }
               >
                 <AlertTriangle size={10} />
                 {rr.errored}
--- a/components/evals3/RunInspectorPage.tsx
+++ b/components/evals3/RunInspectorPage.tsx
@@ -627,6 +627,20 @@ export const RunInspectorPage: React.FC = () => {
             </Button>
           </div>
         </div>
+        {/* Run-level judge-failure banner (lib/judgeFailureSummary.ts). Before
+            this existed, a run whose cases all failed AT THE JUDGE STEP
+            (e.g. the agent-trace-judge's pre-fix "needs a runId or trace
+            correlation hint" validation error for a non-instrumented agent)
+            showed only the bare "N ⚠" errored count above -- no reason. */}
+        {run?.judgeFailureSummary && (
+          <div
+            data-testid="run-judge-failure-banner"
+            className="mt-2 flex items-start gap-1.5 rounded border border-amber-200 dark:border-amber-800 bg-amber-50 dark:bg-amber-900/20 px-2.5 py-1.5 text-[11px] text-amber-800 dark:text-amber-300"
+          >
+            <AlertTriangle size={12} className="shrink-0 mt-0.5" />
+            <span>{run.judgeFailureSummary}</span>
+          </div>
+        )}
       </div>
 
       {/* Re-run Confirm Dialog (EvaluationRun only) -- doc concern, see above. */}
--- a/lib/judgeFailureSummary.ts
+++ b/lib/judgeFailureSummary.ts
@@ -77,7 +77,8 @@ export function extractJudgeFailureReason(report: CaseFailureLike | null | undef
  * `undefined` for cases that weren't a judge failure) into one run-level
  * summary line.
  *
- * Only fires when judge failures are the DOMINANT outcome (>=50% of total
+ * Only fires when judge failures are the DOMINANT outcome (a strict majority,
+ * > 50% of total
  * cases) -- a couple of incidental judge failures amid an otherwise-healthy
  * run don't need a run-level banner; the per-case badge already covers
  * that. Ties on the most-frequent distinct message break in first-seen
@@ -95,7 +96,7 @@ export function computeJudgeFailureSummary(
   if (total <= 0) return undefined;
   const judgeFailed = reasons.filter((r): r is string => !!r);
   if (judgeFailed.length === 0) return undefined;
-  if (judgeFailed.length < total / 2) return undefined;
+  if (judgeFailed.length * 2 <= total) return undefined;
 
   const counts = new Map<string, number>();
   for (const r of judgeFailed) counts.set(r, (counts.get(r) ?? 0) + 1);
--- a/lib/judgeFailureSummary.ts
+++ b/lib/judgeFailureSummary.ts
@@ -1,0 +1,113 @@
+/*
+ * Copyright OpenSearch Contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Run-level judge-failure surfacing.
+ *
+ * Background: an individual case whose evaluator (judge) could not produce
+ * a verdict is already visible per-case as the amber "errored" badge (see
+ * `lib/runStats.ts` / `components/evals3/ResultStatus.tsx`) -- but nothing
+ * aggregates WHY across a run. A run whose judge call fails on every case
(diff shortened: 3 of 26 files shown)


goyamegh:
## codex_review (adversarial second opinion, different model family) — applied / rejected

| # | Sev | Finding | Decision |
|---|---|---|---|
| 1 | high | Hints-only mode enables trace tools from caller-supplied `agents` with no corroboration (the 403 guard only runs when `runId` is present). | **Rejected for this PR — pre-existing, not introduced here.** The hints-only path (`hasTraceCorrelation` accepting `agents` alone) landed in the #461 lineage; this PR only changes what happens when correlation is *absent* (400 → trajectory-only, zero tools). Agreed the comment claiming hints are "server-derived" is inaccurate for direct API callers; the provider is documented as single-tenant-only (`AGENTS.md` "Trace correlation"). Filed as a follow-up: bind hints to a persisted report id server-side. |
| 2 | high | `query_logs` in hints-only mode degrades to "all logs in the time window" (`/api/logs` has no `serviceName`/`sessionId` filter). | **Rejected for this PR — pre-existing** (same #461 hints-only path). This PR *narrows* exposure: with no correlation the tool is no longer registered at all. Follow-up: withhold `query_logs` unless `runId`/`sessionId` is present. |
| 3 | med | `judgeFailureSummary` is sticky — writers only set it when truthy and storage merges updates, so a stale banner outlives a successful Retry judgement. | **Applied.** All recompute sites (`refresh-stats`, `refresh-all-stats`, backfill, `refreshBenchmarkRunStats`) now write `judgeFailureSummary: <value> ?? null`; `retryJudgementForRun` re-derives it from the fresh report docs and clears it. |
| 4 | med | Evaluation-run summary misses the deferred trace-poller failure mode (only synchronous judge failures are captured at run completion). | **Partially applied / documented.** The trace-mode (`useTraces: true`) poller path is out of scope for this incident (a `useTraces: false` agent); `refreshBenchmarkRunStats` (the poller's benchmark-run refresh) does compute it. Left as a documented TODO in `evaluationRunner.ts` for the first-class evaluation-run poller path. |
| 5 | med | `asyncBenchmarkStorage` / `asyncRunStorage` client mappers are allow-lists and dropped `judgeFailureSummary` / `judgeMode`, so the inspector banner would never render for benchmark-embedded runs. | **Applied.** Both fields are now mapped in/out in all three benchmark-run mappers and the report mapper. (Eval-run docs were already passthrough — that path was covered by the e2e.) |
| 6 | low | Legacy classifier `/^Evaluation failed:.*\bjudge\b/i` is heuristic; banner fired at exactly 50%. | **Applied threshold; regex kept.** Now a strict majority (`> 50%`, tie → no banner; unit-tested). The regex is scoped to `status: 'failed'` + no verdict + trajectory present, only affects the *legacy* pre-fix shape, and is the single definition shared with `isJudgeFailedCase`; new reports use the unambiguous `kind=judge_failed` tag. |
| 7 | low | "Zero tools" in trajectory-only mode relies on `createAgentSession({ tools: [] })` semantics, not a contract test. | **Rejected (accepted risk).** Both `extensionFactories: []` and `tools: []` are set, so no trace tool exists to enumerate regardless of SDK defaults for `[]`; the pre-existing trace-tools mode already relies on the same `tools:` allow-list semantics. |

Post-review gates: `tsc` clean · `build:all` ✅ · `test:unit` 360 suites / 6143 passed ✅ · targeted integration suites (judge route, degradation, retry-judgement, benchmarks, storage mappers) 438 passed ✅ · `npm audit --audit-level=high` unchanged vs `main` baseline (pre-existing echarts advisories).

TackAdam (approved):

PR #483, opened 2026-09-04
fix(judge): agentic trace judge keeps improvement strategies and custom output fields the evaluator asked for

goyamegh:
## Problem

On the run report's Judge Evaluation tab, a run judged by the agentic trace judge (`judgeModelId: agent-trace-judge`) never showed the "Improvement Strategies" section, even when the model emitted a full `improvement_strategies` array per the evaluator's Output Format instructions. The same evaluator judged by a plain Bedrock model showed them fine.

## Root cause

`evaluateWithPiAgenticTrace` (`server/services/piAgenticJudgeService.ts`) unconditionally forced `improvementStrategies: []` on its return value, discarding whatever the model actually emitted. This was a stale RFC 004 stance ("individual judge verdicts never carry recommendations") that no longer matches the shipped UI, which renders per-verdict strategies for every other judge provider (bedrock, openai-compatible, litellm, claude-code, pi, agentic).

## Fix

- Keep whatever `parseJudgeResponse` extracted from the model's actual output instead of forcing `[]`. Updated the stale comment to reflect the current UI contract.
- Audited the `extraFields` (custom evaluator output keys) plumbing end-to-end (`services/evaluation/index.ts` -> storage adapters -> `judgeAccessor.buildJudgeMatcherEntry` -> `MatcherResultsPanel`) and confirmed it's intact -- no fix needed there.
- While tracing that path, found and fixed a second, broader bug: `services/benchmarkRunner.ts`'s two single-test-case persistence helpers (the placeholder-update path in `runSingleUseCase`, and `saveReportWithModule`) never forwarded `report.matcherResults` at all, so the unified judge surface was silently dropped for **every** judge provider on any run through `/api/evaluate`. Only the trace-mode-polled paths elsewhere in the same file already forwarded it.

## Tests

- **Unit** (`tests/unit/server/services/piAgenticJudgeService.test.ts`): drives the real `evaluateWithPiAgenticTrace` end-to-end with only the optional pi SDK mocked at its module boundary -- model emits strategies (kept + shaped), emits none (`[]`), emits a custom key (kept in `extraFields`).
- **Integration** (`tests/integration/server/routes/agentJudgeImprovementStrategies.integration.test.ts`): a real `/api/evaluate` run against the built-in mock connector agent, judged by `agent-trace-judge` against an evaluator whose prompt requests strategies, asserts the persisted `report.improvementStrategies` and `matcherResults[0].improvementStrategies` -- this integration test is what caught the `matcherResults`-forwarding bug.
- **E2E** (`tests/e2e/agent-judge-improvement-strategies.spec.ts`): the Judge Evaluation tab renders the Improvement Strategies section for such a report.

All red before / green after the respective fixes (verified locally).

## Checklist
- [x] `npm run build:all`
- [x] `npm run test:unit` (359/359 suites pass)
- [x] `npm run test:integration` (pre-existing, unrelated failure in `sigv4CredentialRetry.integration.test.ts` confirmed identical without this change)
- [x] `npm run test:e2e` (new spec passes)
- [x] `npm audit --audit-level=high` (pre-existing vulnerabilities unrelated to this change, unchanged from a clean `origin/main` checkout)
- [x] DCO signoff
- [x] CHANGELOG.md updated under `## [Unreleased]` / `### Fixed`


goyamegh:
## codex_review (adversarial pass, gpt-5.4)

Ran against the diff before requesting approval. Findings and resolution:

| 1 | High | `return { ...parsed }` trusts the model's `improvement_strategies` array items unvalidated; the run-detail UI indexes `strategy.category`/`strategy.priority` unconditionally (`strategy.category.replace(...)`, `priorityColors[strategy.priority]`) and would throw or render `undefined` styling on a malformed entry -- this was previously invisible only because the agentic judge always forced `[]`. | **Applied.** Added `normalizeImprovementStrategies()` to the shared `judgeResponseParser.ts` (used by every judge provider): drops non-object/non-string junk, coerces a legacy bare-string entry into a valid object (preserving content rather than losing it), defaults missing/invalid fields so every entry always matches the typed shape. New unit coverage in `judgeResponseParser.test.ts`. |
| 2 | Medium | The PR narrative overclaimed that missing `matcherResults` left the run's judge panel entirely empty; in reality a legacy synthesized `llm-judge` row still renders from `llmJudgeReasoning` -- what was actually lost is the enriched per-entry data (`improvementStrategies`, `judgeExtraFields`). | Acknowledged in this comment; not a code change, a wording precision. The bug (per-entry data silently dropped) and fix are unaffected. |
| 3 | Medium | `judgeExtraFields` attached by `buildJudgeMatcherEntry()` is not actually rendered anywhere in `MatcherResultsPanel.tsx` today -- the only live extra-fields surface is `report.llmJudgeResponse.extraFields`. The PR text oversold restoring a UI path that doesn't (yet) render. | Acknowledged; not a regression introduced by this PR (pre-existing gap in `MatcherResultsPanel`), out of scope for this fix. Flagging as a follow-up rather than scope-creeping this PR. |
| 4 | Medium | The integration test only exercises the placeholder-update branch of `runSingleUseCase` (always taken by `/api/evaluate`); `saveReportWithModule` (the create path, used when no placeholder is pre-created) was fixed but left unverified. | **Applied.** Added a direct unit test in `benchmarkRunner.test.ts` asserting `matcherResults` is forwarded on BOTH branches (placeholder-update via `mockRunsUpdate`, and create via `mockRunsCreate`). |
| 5 | Medium (design) | The e2e spec seeds already-correct data and only proves the render layer, not the parsing/persistence layers -- low marginal signal given the integration test. | Not applied -- kept as cheap UI-contract coverage per the task's e2e requirement; the integration + unit tests are the layers that actually catch regressions in parsing/persistence. |
| 6 | Medium (design) | The integration test mounts hand-picked routers instead of the real `server/app.ts` to dodge an `import.meta.url`/ts-jest CJS incompatibility -- a workaround rather than a root fix. | Not applied in this PR -- consistent with the existing repo pattern (`judgeAgentProviderHints.integration.test.ts` already does this for the same reason); fixing `server/app.ts`'s ESM/CJS testability is a separate, larger change out of scope here. |
| 7 | Low (design) | No migration/backfill story for already-persisted reports missing `matcherResults`/`improvementStrategies` -- only NEW runs get the fix. | Not applied -- this is forward-looking bug remediation, not a backfill migration; noted here for visibility, no backfill script included. |

**Proof:** unit (`piAgenticJudgeService.test.ts`, `judgeResponseParser.test.ts`, `benchmarkRunner.test.ts`), integration (`agentJudgeImprovementStrategies.integration.test.ts`), and e2e (`agent-judge-improvement-strategies.spec.ts`) all verified red-before/green-after locally; full `npm run test:unit` (359/359 suites) and `npm run test:integration` (only a pre-existing, unrelated `sigv4CredentialRetry` failure, confirmed identical on a clean `origin/main` checkout) pass.

TackAdam (approved):

PR #484, opened 2026-09-04
feat(report): per-test SDK definition view (pretty + evaluate function) instead of the whole eval file

goyamegh:
### Description

For a code-SDK test case, the run report's **Test Case Definition** collapsible (and the Test Case detail page) rendered the *entire* persisted `sourceCode` — the whole eval file. For a suite that registers its tests from a data table in a `for` loop, that file never shows *the* test the report is about. JSON test cases, by contrast, get a reader-friendly definition (prompt, expected outcomes, labels, context).

This PR makes the SDK case render "similar to JSON": under the `<file>.eval.js` header, **only this test** is shown, as **Pretty** (default) or **Evaluate function**, with **Whole file** kept as a secondary segment so nothing is lost.

#### Data model (additive)
- `TestCase.definition?: { registeredAs: 'sdk'; options: Record<string, unknown>; bodySource: string; bodyTruncated?: boolean }` — captured by the loader for every registered test: the resolved `test()` options exactly as passed (JSON round-tripped so functions/`undefined` are dropped; circular → `{}`), and `Function.prototype.toString()` of the evaluate callback bounded at 32KB. Both import paths (CLI `benchmark -f`, server-side `code-import` resolver) persist it.
- **Hash decision:** `definition` is **not** part of `sourceHash` (its inputs are already covered by the hashed option fields + whole-file text). Re-importing an unchanged file classifies every record `unchanged` — same ids, no version bump. Because that path would otherwise never write anything, `bulkUpsert` (file + OpenSearch adapters) **backfills a missing `definition` in place** on the current version doc (partial update of `<id>-v<current>`, no new version) and never overwrites an existing one. Net: existing suites need **one re-import** to gain per-test definitions.
- `definition` is stripped from `fields=summary` list projections (like `sourceCode`) and mapped `enabled: false` in the test-cases index so user-shaped `options` never inflate dynamic mappings.

#### UI
- New `SdkTestDefinitionView`: filename header + segmented control **[Pretty | Evaluate function | Whole file]**. Pretty renders this test's options through the same `TestCaseDefinition` block JSON tests use (plus a `test("name", …)` / timeout line); Evaluate function renders `bodySource` with prism highlighting; Whole file renders the existing `EvalSourceCodeView`. Legacy records (no `definition`) fall back to whole-file + a one-line re-import hint. JSON branch unchanged.
- `HighlightedCodeBlock` extracted from `EvalSourceCodeView` (Prism + line gutter) so both code surfaces render identically.

Wired into `CollapsibleTestCaseDefinition` (run report / run inspector / Test Case Output pane) and `TestCaseDetailPage`.

### Tests
- **Unit:** `loader.definitionCapture.test.ts` (per-test options/body from a loop, two-arg form, truncation, circular options, hash unaffected), `SdkTestDefinitionView.test.ts`, `HighlightedCodeBlock.test.ts` (escaping probe), extended `CollapsibleTestCaseDefinition`, `bulkUpsert` (file) + `opensearch/StorageModule` (equal-hash backfill writes `<id>-v<current>` in place, never bumps, never overwrites), `routes/storage/testCases` (summary strips `definition`).
- **Integration:** `testCasesDefinitionCapture.integration.test.ts` — real loader on the synthetic `tests/fixtures/definition-capture.eval.js` (3 loop-registered + 1 standalone test) → bulk route → each record carries its own definition; summary strips it; unchanged re-import bumps nothing; records seeded without `definition` are backfilled at version 1.
- **E2E:** `sdk-test-definition-view.spec.ts` — run report shows Pretty by default with this test's prompt and **not** the sibling tests; Evaluate function shows the callback; Whole file still reachable; legacy fallback; Test Case page parity.

### Notes
- Evaluate body for `.eval.ts` is the esbuild-transpiled JS that actually executed (type-stripped), since `toString()` reflects the executed function. Loop-registered tests share one callback text by design — per-test identity lives in `options`, which is why Pretty is the default.

### Check List
- [x] New functionality has been documented (CHANGELOG under Unreleased).


goyamegh:
### Verification proof (local, synthetic fixtures only)

Test server: isolated file-storage backend on a private port (`BENCHMARK_RUN_RECOVERY_DISABLED=1 EVALUATION_RUN_RECOVERY_DISABLED=1`), built from this branch after rebasing onto `main` @ `ac025b03d`.

**Typecheck / build**
```
tsc --noEmit -p tsconfig.json   → clean
npm run build:all               → OK (vite + server + cli + lib)
```

**Unit (full)**
```
Test Suites: 370 passed, 370 total
Tests:       5 skipped, 6342 passed, 6347 total
```

**Integration — branch-relevant suites against the live HTTP API**
```
PASS tests/integration/server/routes/storage/testCasesDefinitionCapture.integration.test.ts
  ✓ each imported record carries its OWN definition (options + evaluate body), retrievable via GET /:id
  ✓ list view (fields=summary) strips definition and sourceCode but keeps provenance
  ✓ legacy records (no definition) are backfilled in place on the next unchanged re-import — still version 1
  ✓ re-importing the UNCHANGED file is `unchanged` for every record — same ids, no version bump
PASS tests/integration/server/routes/storage/testCasesBulkUpsert.integration.test.ts   (11 tests)
PASS tests/integration/server/routes/storage/testCases.integration.test.ts             (17 tests)
Tests: 32 passed, 32 total
```

**E2E (Playwright, CI mode against the prod bundle)**
```
✓ sdk-test-definition-view.spec.ts › shows Pretty by default with ONLY this test, Evaluate function on switch, whole file still reachable
✓ sdk-test-definition-view.spec.ts › legacy SDK record without definition falls back to the whole file with a re-import hint
✓ sdk-test-definition-view.spec.ts › Test Case detail page uses the same per-test view (Pretty by default)
✓ eval-source-code-view.spec.ts (4 tests — pre-existing whole-file view unaffected by the HighlightedCodeBlock extraction)
✓ run-report-redesign, run-report-lazy-load, run-report-insights, run-inspector-no-autoselect, evals3-test-cases, evals3-benchmark-runs, benchmark-runs, retry-judgement
80 passed
```
One unrelated pre-existing failure observed in the legacy data-dependent `test-case-runs.spec.ts › should open run modal when clicking Run Test`: `QuickRunModal` calls `testCase.initialPrompt.trim()` and crashes on a prompt-less (deterministic) SDK test case that happened to be first in the list — `QuickRunModal.tsx` is untouched by this PR (last changed in #219); will be filed separately.

**Loader capture on the synthetic fixture** (`tests/fixtures/definition-capture.eval.js`, 3 loop-registered tests + 1 standalone) — a record imported from the loop:
```json
{
  "registeredAs": "sdk",
  "options": {
    "prompt": "Synthetic prompt number two",
    "description": "Synthetic description for TWO",
    "expectedOutcomes": ["Outcome two is satisfied", "Outcome two is fast"],
    "labels": ["category:Synthetic", "difficulty:Easy", "marker:TWO"],
    "timeout": 30000
  },
  "bodySource": "async ({ result }) => {\n      // shared-loop-body-marker\n      if (!result) throw new Error(`no result for ${c.name}`);\n    }"
}
```
`bodySource.length < sourceCode.length`; the standalone test's `bodySource` contains its own marker and not the loop's.

**Version semantics**: second import of the identical file → `{created: 0, updated: 0, unchanged: 4}`, same ids, `currentVersion: 1`, `/versions` total = 1. Records seeded *without* `definition` → next unchanged import backfills it at v1 (`PUT` of `<id>-v1` doc in place; no v2).

**Rendered surface** (run report → Test Case Definition, SDK badge): `evals/<file>.eval.js` header → `JavaScript` badge → `[Pretty | Evaluate function | Whole file]`; Pretty shows `test("<name>", …)`, timeout badge, labels, description, input, expected outcomes for this test only; Evaluate function shows the 3-line callback with prism tokens and a line gutter; Whole file shows the pre-existing 11-line full-file view. Screenshots kept off-repo.

goyamegh:
### Follow-up pushed: the SDK "Test Case Definition" dropdown never renders empty (owner bug, 2026-09-08)

**Symptom (live tunnel, run inspector, a pre-capture SDK suite — every case has `sourceFile` + ~28 KB `sourceCode`, no per-test `definition`):** clicking **Test Case Definition** showed only the amber "Per-test definition not captured at import — re-import…" hint and a one-line `dist/<file>.eval.js · JavaScript · 983 lines` row. DOM: `data-mode="legacy"`, `eval-source-toggle aria-expanded="false"`, no `eval-source-code-body`. The code was a second click away, inside a section the user had just opened — it read as "nothing shows up".

**Root causes:**
1. `SdkTestDefinitionView`'s legacy branch rendered `EvalSourceCodeView` **collapsed** (its default) — a nested second disclosure.
2. `RunInspectorPage` bulk-loads test cases as a **summary** projection (no `sourceCode`/`definition`) and fetches the full record lazily per selected row. Until it lands (or if it fails) the summary is indistinguishable from a legacy record → same false hint, and the whole-file view's "Source not captured at import — re-run `benchmark -f` to backfill" advice for a file that IS captured. Also: on re-select, the previous row's full record stayed in state until the new fetch landed.

**Fix (`b87a33264`, `02f050e74`, `fe36ace65`):**
- Legacy branch renders the whole file **expanded** (still collapsible), keyed by test-case id so stepping to another legacy case re-expands it.
- New discriminated prop `fullRecord?: 'loading' | 'error' | 'missing'` (unset = the record is authoritative) on `SdkTestDefinitionView` → `CollapsibleTestCaseDefinition` → `TestCaseInspectorPanel`. SDK view: eval-file header + language badge + "Loading full definition…" / honest error / "record not found" rows. JSON branch: loading row during the in-flight window only.
- `RunInspectorPage` keeps the lazily fetched record **keyed by id** and derives `testCase` + `fullRecord` synchronously in render — a selection change can never paint the previous case's definition for a frame. A `null` resolution (case deleted since the run) is `'missing'`, not a transient error.

**Red → green:** 11 new/changed unit + RTL assertions fail on `80dc49253`, pass on this head.

```
tests/unit/components/evals3/SdkTestDefinitionView.test.ts        14 passed (legacy expanded; re-expands on case switch; legacy w/o
                                                                             source → placeholder; loading; error; missing; captured wins)
tests/unit/components/evals3/CollapsibleTestCaseDefinition.test.ts 15 passed (SDK + JSON loading rows; SDK error row; JSON keeps summary)
tests/unit/components/evals3/RunInspectorPage.test.ts              41 passed (fullRecord lifecycle; no stale record on re-select incl. a
                                                                             late resolution for the old id; error vs missing)
tests/e2e/sdk-test-definition-view.spec.ts                          4 passed (legacy expanded on /runs/:id; run-inspector route with a
                                                                             benchmark-linked SDK case w/o definition: header + expanded code)
tests/e2e/eval-source-code-view.spec.ts                             4 passed (Test Case page: expanded by default, toggle collapses)
```

**After, local build against the shared cluster (read-only), same run:** expanded dropdown → `data-mode="legacy"`, `aria-expanded="true"`, 983 highlighted lines rendered under the `dist/<file>.eval.js` header; stepping through 4 further cases: each lands `legacy` + expanded, and the "Source not captured at import" text never appears.

Composes cleanly with #468 / #486 (both touch `RunInspectorPage.tsx`): origin/main + #468 + #479 + this + #486 → `tsc` clean, seam suites green.

**Adversarial review (`codex_review`, gpt-5.4):** 8 findings. Applied: effect-time reset still left one stale-record frame → id-keyed render-time derivation; `defaultOpen` is mount-only → `key`; "select again to retry" was not an achievable path (re-clicking the selected row is a no-op) → dropped, `null` → `'missing'`; two booleans → one discriminated prop; JSON branch only gates the in-flight window. Rejected with reason: flattening `EvalSourceCodeView`'s header/toggle out of the legacy view (its collapsible contract is shared with the "Whole file" segment and it remounts expanded whenever the outer section reopens); the auto-expanded legacy code on the Test Case detail page is intentional (matches the captured mode's immediate content; the highlighted block is the same one the "Whole file" segment already renders).

TackAdam (approved):

PR #485, opened 2026-09-05
feat(compare): model selector for the trace-grounded deep-dive, default Claude Fable 5.1; lift the 180s deep-dive deadline

goyamegh:
### Description

Two owner asks for the comparison page's **"What's actually different"** panel (the trace-grounded, agentic deep-dive):

1. *"What model is run for What's actually different? I want it to be Fable 5.1."*
2. *"My comparison times out after 180 seconds, remove this limit."*

**What ran before (evidence):** the client never sent a `modelId`, and the server's default reused the agentic judge's `pickJudgeModel`, whose scoring only knows Claude 4.x (`-4-5`/`-4-6` bonus). With `us.anthropic.claude-fable-5-1` credentialed in the pi registry, the live server still logged `[CompareDeepDive] model: amazon-bedrock/global.anthropic.claude-sonnet-4-5-20250929-v1:0` for every deep-dive.

**Model selector (UI + API)**
- New `GET /api/comparison/deep-dive/models` → `{ models: [{provider,id,name}], defaultId }` — read live from the pi `ModelRegistry`, filtered to Claude models on a usable inference profile (region-matching or `global.`; bare ids fail on-demand on Bedrock, wrong-region profiles are dropped), best-first.
- A small `<select>` in the panel header next to *Regenerate*. Default = the server's `defaultId`. Choice persists in this browser only (`localStorage`, same pattern as the panel's custom system prompt). The result cache key now includes the model (prefix `v3` → `v4`), so switching models invalidates the cached narrative for the pair; the first auto-run waits (bounded, 10s) for the model list so it already carries the right `modelId`. The spinner names the model; the footer keeps reporting the model the server actually used.
- `POST /api/comparison/deep-dive` validates `modelId` (string, 1–200 chars → 400), includes it in the job de-dupe key (switching to Fable while a Sonnet job is running starts a *new* job), and an explicitly requested model the server cannot invoke fails loudly instead of being silently swapped.

**Server default:** an explicit **pin** on Fable 5.1 (`DEEP_DIVE_PREFERRED_MODEL_ID = anthropic.claude-fable-5-1`, base-id match so any credentialed `us.`/`global.` profile satisfies it; per-server override `AH_DEEP_DIVE_MODEL_ID`). Only when no Fable 5.1 profile is credentialed does it fall back to a new Claude-5-aware heuristic (`scoreNewestClaudeModel`: any Claude 5 family beats every 4.x, higher minor wins, `fable` preferred in a tie, inference-profile/region rules inherited). **The agentic judge's silent fallback (`pickJudgeModel`) is deliberately unchanged** — regrading runs with a newer model family when the configured judge model is missing would break verdict comparability across runs. Pinned by a regression test.

**180s deadline lifted:** `DEEP_DIVE_DEADLINE_MS` is no longer a 3-minute budget but a **25-minute** safety backstop derived from the job store TTL (`DEFAULT_JOB_TTL_MS − 5 min`), so a genuinely stuck loop is aborted (`session.abort()`) while its job is still pollable. If it fires, the error states how long the loop ran, with which model, and that Regenerate usually succeeds. The client's separate 200s wall-clock cap is removed: it polls for as long as the job reports `running` (404/`error` remain terminal) and shows `Model: <name>` + a `Nm Ns` elapsed label. Checked the pi session for its own caps: the agent loop has no per-turn limit, provider retries are bounded (3), and the deep-dive's trace-tool budgets (12 cases / 40 calls) bound the loop's *work*, not its time. Fable 5.1 keeps its registry defaults (1M ctx / 128k output; the registry's `thinkingLevelMap` clamps the thinking level) — no `maxTokens`/`thinkingLevel` override.

### Issues Resolved
Owner asks (no tracked issue).

### Testing & Coverage
- [x] Unit tests added/updated — `piAgenticJudgeService.test.ts` (Claude-5 scoring, judge fallback pinned to 4.x), `comparisonDeepDiveService.test.ts` (Fable 5.1 pin incl. newer-5.x-does-not-displace, env override, explicit `modelId` echoed, unavailable model rejected, no session caps, model-list filtering, a 10-minute generation completes while a stuck loop is still aborted at 25 min with the new message), `comparisonDeepDiveJobStore.test.ts` (modelId in de-dupe key), RTL `ComparisonDeepDive.modelSelect.render.test.ts` (default resolution, localStorage round-trip, cache invalidation on switch, first POST carries the model, bounded models fetch, spinner past the old cap)
- [x] Integration tests added/updated — `comparisonDeepDiveJobs.integration.test.ts` (`GET …/models`, `modelId` threaded + echoed in the job result, 400s, per-model de-dupe)
- [x] E2E tests added/updated — `tests/e2e/comparison-deepdive-model-select.spec.ts` (selector visible on the compare page, default Fable 5.1, persisted choice, switch regenerates, no client-side timeout while running)
- [x] `npm run build:all && npm run test:all` passes locally (see proof comment)
- [x] Patch coverage >= 90% (new/changed lines are tested)
- [x] No overall coverage decrease (ratchet enforced by Codecov)

### Compliance
- [x] DCO signoff on all commits (`git commit -s`)
- [x] CHANGELOG.md updated
- [x] SPDX license header on new files


goyamegh:
### Verification proof

**What ran before** — live server log for the deep-dive endpoint, no `modelId` sent by the client (3 generations, all the same model):
```
[CompareDeepDive] model: amazon-bedrock/global.anthropic.claude-sonnet-4-5-20250929-v1:0 runs: A,B cases: 6
```
Re-scoring the same credentialed registry (98 models) with the OLD `pickJudgeModel`: `global.anthropic.claude-sonnet-4-5-…` scores 128, `us.anthropic.claude-fable-5-1` scores 58 — Fable could never win.

**After — models endpoint on a local build of this branch** (shared-cluster server, read-only):
```
GET /api/comparison/deep-dive/models
default us.anthropic.claude-fable-5-1
  us.anthropic.claude-fable-5-1 | Claude Fable 5.1 (US)
  global.anthropic.claude-fable-5 | Claude Fable 5 (Global)
  us.anthropic.claude-fable-5 | Claude Fable 5 (US)
  global.anthropic.claude-sonnet-4-5-20250929-v1:0 | Claude Sonnet 4.5 (Global)
  global.anthropic.claude-sonnet-4-6 | Claude Sonnet 4.6 (Global)
  ... (18 total; bare ids and eu./au./jp. profiles excluded)
```

**Real deep-dive, 62-case comparison of two real runs (two different retrieval agents), same request body, two models:**

| model | durationMs | cases traced | chart | experiments | outcome |
|---|---|---|---|---|---|
| `amazon-bedrock/us.anthropic.claude-fable-5-1` (default, no `modelId`) | **118629** (118.6s) | 7 | yes | 4 | `done` |
| `amazon-bedrock/us.anthropic.claude-sonnet-4-6` (explicit `modelId`) | **130909** (130.9s) | 9 | yes | 4 | `done` |

Both completed via the async job (`POST` → `202 {jobId}` → polls `running` → `done`); `result.modelId` echoes the model actually used. Rendered page footer on the real comparison: `Generated by us.anthropic.claude-fable-5-1 in 109s`, selector preselected to "Claude Fable 5.1 (US)" with 18 options.

**Deadline:** `DEEP_DIVE_DEADLINE_MS` = `DEFAULT_JOB_TTL_MS − 5 min` = 1,500,000 ms (25 min). Unit tests (fake timers) verify a 10-minute generation completes and a never-settling loop is aborted at exactly 25 min with:
```
Comparison deep-dive stopped by the safety deadline after 25m 0s (model mock/mock.claude-fable-5-1) — the agent loop was still running … Click Regenerate to try again …
```

**Gates (local, `AH_PORT` test server on the shared cluster):**
```
npm run build:all                  ✓ (vite built, server/cli/lib esbuild ok)
npm run test:unit                  Test Suites: 361 passed · Tests: 6190 passed, 5 skipped
npm run test:integration           98 suites: 85 passed in the parallel run; the 13 that timed out
                                   (CLI-subprocess / storage suites, 90–230s each) all pass with
                                   --runInBand: 12 suites, 117 tests passed. sigv4CredentialRetry (2)
                                   fails identically on a clean origin/main checkout → pre-existing.
                                   comparisonDeepDive*.integration: PASS
targeted e2e (6 deep-dive/comparison specs incl. the new one): 17 passed
git log origin/main..HEAD | grep Signed-off-by   ✓ (1 commit)
CHANGELOG.md                       ✓ [Unreleased] › Added
npm audit --audit-level=high       34 (15 high) — identical on main; this PR changes no dependencies
```

**codex_review (gpt-5.4) findings → disposition**

| # | sev | finding | applied? |
|---|---|---|---|
| 1 | high | "newest Claude" heuristic ≠ pinning Fable 5.1 — a registry change could silently move the default | **applied** — default is now an explicit pin (`DEEP_DIVE_PREFERRED_MODEL_ID`, env override), heuristic only as fallback; test: a hypothetical `fable-5-2` does not displace 5.1 |
| 2 | high | 25-min jobs on a 3-slot in-memory queue → wedged jobs can starve the feature | rejected — owner asked for the limit removed; a deadline-hit job frees its slot (`error` state) and `session.abort()` stops the loop; 3-slot cap and in-memory store are pre-existing and out of scope |
| 3 | high | the 180s limit was moved, not removed | rejected — owner explicitly asked to keep a long hang backstop; 25 min ≈ 8–13× the longest real generation |
| 4 | med | auto-run now depends on the models fetch, which has no timeout | **applied** — `AbortSignal.timeout(10s)`; on timeout/failure the panel generates with the server default (RTL test) |
| 5 | med | deadline coupled to job TTL | rejected — deliberate: the deadline must fire *before* the TTL sweep or the client gets a silent 404 instead of an honest error; owner offered this coupling |
| 6 | med | sticky localStorage choice bypasses future default changes | rejected — intended (user choice persists, same as the custom prompt); a stored id no longer offered already falls back to the default |
| 7 | med | hand-rolled version parsing will rot | partially — with the pin it is fallback-only; kept small and unit-tested against real registry ids |
| 8 | low | cache key uses requested id, footer shows resolved id | rejected — the selector only offers ids from the server list (already region-resolved), so requested == resolved; footer intentionally shows the server's `modelId` |

TackAdam (approved):

PR #489, opened 2026-09-08
fix(judge): surface improvement strategies on the run report — map llmJudgeResponse in the browser read path, recover from raw judge text, add an idempotent backfill

goyamegh:
### Description

**Symptom:** the run report's Judge Evaluation tab showed no "Improvement Strategies" for many recent reports, even though the judge emitted them.

**Diagnosis (headless browser against a live server):** the tab's rendering path DOES render the section — inspector route, benchmark-scoped route and `?reportId=` deep link all mount the same `RunDetailsContent` — whenever the persisted `improvementStrategies` array is populated. Every report that showed a blank section was judged by the agentic trace judge, which for a window of time returned `improvementStrategies: []` unconditionally while still persisting the model's verbatim text — with the full `improvement_strategies` array — on `llmJudgeResponse.rawResponse`. A read-only dry run against a shared cluster found **1,383 such reports (4,584 strategies) among the most recent 10,000**.

**Three layers fixed:**

1. **Browser read path dropped the judge response entirely** — `services/storage/asyncRunStorage.ts` `toTestCaseRun()` never mapped `llmJudgeResponse`, so `report.llmJudgeResponse` was `undefined` for every browser-side reader. Verified on the live page: the Judge Evaluation tab's "Judge output" block (raw text, judge model, tokens, extra fields — #265) never rendered for any report. Mapped now.
2. **Read-side recovery** — new `lib/judgeStrategies.ts`: `resolveImprovementStrategies(report)` returns the persisted array when present, otherwise re-parses `rawResponse` with the same extractor the judge parser uses (`extractJsonFromResponse` moved here from the server parser and re-exported, so browser + script share one implementation). `RunDetailsContent` renders recovered strategies with a count badge and a dashed "Recovered from the judge's raw output …" notice pointing at the backfill, instead of a blank section. Persisted strategies render exactly as before (no notice).
3. **Idempotent backfill** — `npx tsx scripts/backfill-improvement-strategies.ts`: dry-run by default, `--apply` writes, `--run <evaluationRunId>` (repeatable) targets one run's reports via the batch `?ids=&fields=` path, unknown flags refused. For each report whose stored array is empty but whose raw text yields ≥1 strategy it PATCHes `improvementStrategies`, `llmJudgeResponse.improvementStrategies` and — only when the report has exactly one `[llm-judge]` matcher row whose own array is empty — that row's `improvementStrategies`. Nothing else on the doc is rewritten; an applied report is skipped on the next pass.
   **Shared-cluster guard rails:** provider gate (`--judge`, default `agent-trace-judge`, `any` lifts it); in-flight reports (`metricsStatus` pending/calculating) skipped; `--apply` refuses to run when `/api/storage/health` isn't `ok` (the listing route degrades to sample data on storage errors, which would otherwise look like a clean zero-candidate pass) or when a full scan hit OpenSearch's 10,000-doc listing window — writes happen only after a complete scan. Scans are field-projected (no trajectories / raw events).
4. **Judge matcher row recovers too** — the `[llm-judge]` row's "how to fix it" list is filled from the same resolver (single-judge-row rule), so the section and the row never disagree. Persisted arrays are normalized on read as well.

### Tests

- **Unit:** `tests/unit/lib/judgeStrategies.test.ts` (extract / normalize / recover / resolve, patch builder's three-surface + idempotency contract); `tests/unit/scripts/backfillImprovementStrategies.test.ts` (real script as a child process vs a mock backend: dry-run writes nothing, `--apply` PATCHes exactly the candidates with the exact body, second pass is a no-op, `--run` scoping, unknown-flag refusal); `tests/unit/components/RunDetailsContent.test.ts` (RTL: section renders for PASSED and FAILED reports with persisted strategies; recovered-with-notice for the `[]`+raw shape; nothing when neither surface has strategies); `tests/unit/services/storage/asyncRunStorage.test.ts` (`llmJudgeResponse` mapped).
- **Integration:** `tests/integration/server/routes/storage/backfillImprovementStrategies.integration.test.ts` (real backend: seeded `[]`+raw report → `--apply --run` → `GET /api/storage/runs/:id` carries strategies on all three surfaces, raw text + every other field untouched, pre-populated sibling untouched, second pass no-op; cleanup via `TestDataTracker`).
- **E2E:** `tests/e2e/judge-tab-improvement-strategies.spec.ts` (inspector route via `?reportId=`: persisted strategies render for a passed and a failed report without the notice; the `[]`+raw shape renders recovered strategies with the notice) — imports `./fixtures/test-fixtures`.

### Adversarial review (`codex_review`, gpt-5.4)

Applied: provider gate + in-flight skip (was: any report with an empty array and raw strategies — would have rewritten reports whose empty array was intentional, and raced live pollers) · matcher rows patched only when there is exactly one judge row (was: report-level array stamped onto every empty judge row — wrong for multi-claim SDK reports, since `llmJudgeResponse` holds the LAST judge call only) · storage-health probe + refuse-apply-on-capped-scan (was: a storage outage read as a clean zero-candidate pass) · judge row recovers alongside the section (was: two surfaces disagreeing until backfill) · notice no longer names a historical cause or tells UI viewers to run a bulk-write command · persisted arrays normalized on read.
Not applied: `if_seq_no`/`if_primary_term` optimistic concurrency or a dedicated server endpoint (no conditional-update path exists in the storage layer today; for a one-off operator script the in-flight skip + `--run` targeting bound the race window — a new write API for this is speculative) · `search_after`/PIT enumeration (the listing route has no cursor; `--run` is the precise path and a capped `--apply` is refused) · importing the normalizer into the server parser (that hunk is owned by open #483).

### Notes for reviewers

- `normalizeImprovementStrategies` in `lib/judgeStrategies.ts` intentionally mirrors the one #483 adds inside the server parser; whichever lands second should import from `lib/` (one-line change) — kept out of this PR to avoid a conflicting hunk with #483.
- The backfill is a one-off operator script; it is not wired into startup or CI.

### Check List
- [x] New functionality includes testing (unit + integration + e2e)


goyamegh:
### Verification proof

**Live-page diagnosis (headless Chromium against a deployed server, read-only).** Same run, Judge Evaluation tab, inspector route `/evaluations/runs/<id>/inspect?reportId=<report>`:

| Report | `improvementStrategies` stored | `rawResponse` has `improvement_strategies` | Section in DOM (before) |
|---|---|---|---|
| Bedrock-judged, passed | 2 | yes | ✅ rendered (2 cards) |
| Bedrock-judged, failed | 4 | yes | ✅ rendered (4 cards + judge row "how to fix it") |
| agentic-trace-judged, passed | **0** | **yes (3)** | ❌ nothing |
| agentic-trace-judged, failed | **0** | **yes** | ❌ nothing |

`document.body.innerText` also never contained `Judge model:` on any report → the "Judge output" block was dead (browser mapper dropped `llmJudgeResponse`).

**After (branch build, file backend, the agentic-trace-judged report above copied over verbatim):**
```
{"section":true,"notice":"Recovered from the judge's raw output — …","recs":3,"judgeOutput":true}
```
then `--apply --run <that run>` → `applied 1, failed 0`; dry-run again → `0 candidate(s)`; page re-probe → `{"section":true,"recs":3,"judgeOutput":true}` with the notice gone.

**Dry runs against a shared cluster (nothing written):**
```
DRY-RUN (all reports) — judge gate: agent-trace-judge
scanned 10000 report(s); 1383 candidate(s) with recoverable strategies (4584 strategies total)
  judge agent-trace-judge: 1383
  day 2026-06-08: 4 / 2026-08-31: 23 / 2026-09-01: 98 / 2026-09-02: 319 / 2026-09-03: 310 / 2026-09-04: 629
  WARNING: hit the 10000-doc listing window; older reports were not scanned. (an --apply over a capped scan is refused)

DRY-RUN (runs: 4 evaluation runs from one day) — judge gate: agent-trace-judge
scanned 258 report(s); 257 candidate(s) with recoverable strategies (908 strategies total)
```

**Gates (branch, `CI=1`, isolated ports):**
```
build:all            exit 0
test:unit            369 suites passed · 6350 passed, 5 skipped
test:integration     new suite PASS (dry-run no writes; --apply → GET shows all 3 surfaces; idempotent)
                     7 pre-existing failures identical on a clean origin/main checkout (real kiro-cli, CLI --judge-model vs Bedrock, file-backend storage stats, create-validation) — unrelated
test:e2e (targeted)  8 passed — judge-tab-improvement-strategies.spec.ts ×3, run-inspector-papercuts ×2, judge-failure-summary ×2, +1
npm audit            same 12 pre-existing highs as origin/main (echarts/browserslist/puppeteer-core chain) — none introduced
DCO                  both commits signed off
```

TackAdam (approved):

PR #492, opened 2026-09-08
fix(storage): run DELETE removes both persisted forms (evaluation-run doc + benchmark.runs[] projection) from either endpoint

goyamegh:
### Description

Deleting a run now removes **both** of its persisted forms from **either** delete endpoint, and the benchmark nested-run `DELETE` no longer returns 404 for a run that was never embedded.

**Bug (owner report):** *"the run doesn't get deleted when I go inside the run page and try it myself"* — plus ghost rows reappearing on a benchmark's Runs page after a delete.

**Root cause:** a benchmark-linked run is stored as a first-class `evaluation-run` document **and**, once it finishes, a legacy-shaped projection embedded in `benchmark.runs[]` (dual-write, #399). Each `DELETE` route removed exactly one of the two:

| Route | Removed | Left behind | Symptom |
|---|---|---|---|
| `DELETE /api/storage/evaluation-runs/:id` | doc | `benchmark.runs[]` projection | ghost row on the benchmark Runs page (the UI merges both sources) |
| `DELETE /api/storage/benchmarks/:id/runs/:runId` | projection | doc | ghost row, other direction — **and 404** for any run that was never embedded (every in-flight run; most runs since the dual-write). The run inspector's benchmark-scoped Delete hit exactly this; the client swallowed the 404 and navigated away. |

**Fix:** both routes delegate to one `deleteRunEverywhere()` (`server/services/runDelete.ts`):

- **Projection first, then document** — the two writes aren't atomic; a failure between them leaves a doc-only run that is still listed and still deletable by retry, never an embedded-only ghost that only the nested-run URL can reach.
- **Ownership on the benchmark-scoped route** — the doc is deleted only when it is actually associated with *that* benchmark (`benchmarkId` or a `benchmark` source). A colliding, unlinked, or other-benchmark doc is left alone (`docSkippedNotOwned: true`; 404 if there's no projection either).
- **Sample/demo data stays read-only** on both routes (the evaluation-runs `DELETE` now guards `demo-*` ids like every other storage route).
- **Cancel-before-delete** — a still-`running` run whose executor lives in this process is told to stop first (both per-route token registries fan in via `server/services/runCancellation.ts`; best-effort and process-local, same limitation as the cancel endpoints; reported as `cancelled: true/false`, the delete proceeds either way).
- **Reports are NOT cascaded** — unchanged, per `AGENTS.md` ("Deleting a benchmark or evaluation run does NOT delete its reports").
- **Response shapes are supersets** of the old ones — `{ success: true, cancelled, projectionDeleted, benchmarkId? }` and `{ deleted: true, runId, projectionDeleted, docDeleted, cancelled }` — so clients checking `ok` / `success` keep working.

Companion client-side fixes (dispatching on the run's kind so the UI picks the endpoint that can find the run) are in #468 (run inspector kebab) and #479 (benchmark Runs table — Delete on every row). This PR is independent of both.

### Issues Resolved

Owner papercut (2026-09-08): run Delete from the run page / benchmark page doesn't delete.

### Testing

Red on `origin/main`, green on this branch:

- **Unit** — `tests/unit/server/services/runDelete.test.ts` (ordering, ownership incl. unlinked doc, sample guard, cancel-before-delete, doc passthrough, canceller fan-out incl. a throwing canceller); `tests/unit/server/routes/storage/evaluationRuns.test.ts` and `benchmarks.test.ts` (route-level: ad-hoc / dual-written / source-resolved / running-with-live-token / running-zombie / cross-benchmark / unlinked / demo / 404 / 500).
- **Integration** (real server, file backend) — `tests/integration/server/routes/storage/runDelete.integration.test.ts`: both routes remove both forms; never-embedded doc → 200 (was 404); cross-benchmark + unlinked → 404; demo → 400; reports survive.
- **E2E** — `tests/e2e/run-delete-no-ghost.spec.ts`: on the benchmark Runs page, delete a dual-written run → row gone, **stays gone after reload**, doc 404, projection removed. On `origin/main` this fails with the row reappearing (`Expected: 0, Received: 1`).

### Check List

- [x] New functionality includes testing (unit + integration + e2e).
- [ ] New functionality has been documented — n/a (behavioural fix; CHANGELOG updated).
- [x] API changes companion pull request created — n/a (response shapes are supersets).


goyamegh:
**Verification proof** (local server on an isolated file backend; nothing touched on shared storage)

Red → green, integration suite `tests/integration/server/routes/storage/runDelete.integration.test.ts`:

```
# origin/main server
✕ DELETE /evaluation-runs/:id removes the doc AND the projection embedded in benchmark.runs[]
✕ DELETE /benchmarks/:id/runs/:runId removes the projection AND the evaluation-run doc
✕ DELETE /benchmarks/:id/runs/:runId on a standalone (never embedded) doc of that benchmark → 200, doc gone (was 404)
✕ an UNLINKED (ad-hoc) run doc is not deletable through any benchmark URL
Tests: 4 failed, 3 passed, 7 total

# this branch
✓ DELETE /evaluation-runs/:id removes the doc AND the projection embedded in benchmark.runs[]
✓ DELETE /benchmarks/:id/runs/:runId removes the projection AND the evaluation-run doc
✓ DELETE /benchmarks/:id/runs/:runId on a standalone (never embedded) doc of that benchmark → 200, doc gone (was 404)
✓ a run doc belonging to ANOTHER benchmark is not deletable through this benchmark's URL
✓ an UNLINKED (ad-hoc) run doc is not deletable through any benchmark URL
✓ sample/demo ids are rejected with 400 on the evaluation-runs DELETE too
✓ deleting a run does NOT cascade to its per-test-case report documents
Tests: 7 passed, 7 total
```

E2E `tests/e2e/run-delete-no-ghost.spec.ts` (benchmark Runs page, dual-written run):

```
# origin/main:  ✘  … Delete on the run row → row gone, stays gone after reload …
#               Error: expect(locator).toHaveCount(expected)  Expected: 0  Received: 1   ← the "deleted" run came back on reload
# this branch:  ✓  … Delete on the run row → row gone, stays gone after reload, doc 404, projection removed (2.6s)
```

Unit: `runDelete.test.ts` 19 passed; `evaluationRuns.test.ts` 49 passed; `benchmarks.test.ts` 88 passed. Full unit suite: 6341 passed (the 8 failures in `cli/index.test.ts` / `loader.*.test.ts` fail identically on pristine `origin/main` in this environment — unrelated).

Live repro of the reported bug (read-only against a real instance, DELETE intercepted): from the benchmark-scoped run inspector, the kebab's Delete issued `DELETE /api/storage/benchmarks/<bid>/runs/<rid>` for a run that is a standalone evaluation-run doc (not embedded) → server answers 404 → the client swallowed it and navigated to the runs list with the run intact. On one live benchmark 27 of 48 runs are in exactly that not-embedded state.

**Adversarial review (`codex_review`, gpt-5.4) — applied:** projection-first ordering (a mid-way failure leaves a retryable doc-only run, not an embedded-only ghost); explicit ownership on the benchmark-scoped route (an unlinked or other-benchmark doc is never deleted through another benchmark's URL — previously the caller-supplied id alone would have justified it); sample-data guard on the evaluation-runs DELETE; the pre-flight doc is passed through (no second read / read race); a throwing canceller can't block the others; response shapes documented as supersets. **Rejected:** returning 409 when the doc belongs to another benchmark but this benchmark still holds a projection (would leave the user unable to remove a ghost projection from the UI; the response flags it with `docSkippedNotOwned` instead); refusing to delete a running run whose executor isn't local (zombie runs are exactly what users need to delete — cancellation is best-effort and reported, not promised); collapsing both token registries into one shared map (larger blast radius across two in-flight PRs for no safety gain).

TackAdam (approved):

PR #490, opened 2026-09-08
fix(traces): run-id trace correlation (Strategy B) never matched OSI-ingested spans — query both attribute shapes

goyamegh:
## Summary

Run-id trace correlation (**Strategy B**: `agent_health.run.id` / `gen_ai.conversation.id == runId`) **never matched a single span** on an OpenSearch-Ingestion / Data Prepper `otel-v1-apm-span-*` index — which is this service's *default* index pattern.

That index template stores span attributes **flat**, one top-level field per attribute, dots encoded as `@`:

```
"span.attributes.agent_health@run@id":    "run-…"
"span.attributes.gen_ai@conversation@id": "run-…"
```

Both `fetchTraces` (Traces tab, `/api/traces`, judge + comparison trace tools) and `computeMetrics` / `computeBatchMetrics` (`/api/metrics`, `/api/metrics/batch`) built their Strategy-B `terms` clauses against the **nested** path only (`attributes.agent_health.run.id`, `attributes.gen_ai.conversation.id`) — an unmapped field on that index — so they matched nothing. Spans were only ever found via Strategy A (shared `traceId`) and Strategy C (`service.name` + time window).

The READ side (`transformSpan`) always tolerated both shapes — #469 fixed this exact mismatch for the metrics *reader* — which is why spans that *were* found rendered fine and the *query* gap went unnoticed.

### Measured (read-only, on an OSI-ingested observability cluster)

| | `_count` |
|---|---|
| spans on the index | 292,269 |
| spans carrying a run-id attribute — flat-`@` shape | 5,814 |
| spans carrying a run-id attribute — nested shape | **0** |
| shipped Strategy-B clause, a REST agent's run + the eval span's run | **0** |
| same two run ids via the flat-`@` path | 53 + 1 |
| **the exact clause this PR emits, same two run ids** | **54** |

### Second gap found while verifying on a real node

On a nested-schema index with plain **dynamic** mapping (no template), OpenSearch maps a string attribute as analyzed `text` + `.keyword`. A hyphenated run id (`run-1788…-a4e3j7lsj`) is tokenized on `-`, so `terms` on the base path matches nothing — only `attributes.<attr>.keyword` correlates (the same reason Strategy D already queried `attributes.session.id.keyword`).

## Fix

Centralize every run-id / session-id query path in `tracesService.ts`:

- `attributeFieldPaths(attr)` → `attributes.<attr>`, `attributes.<attr>.keyword`, `span.attributes.<attr with @>`
[...]

PR #488, opened 2026-09-05
chore(deps): resolve all high-severity npm audit findings (15 → 0)

goyamegh:
### Description

A pristine checkout of `main` fails the pre-PR audit gate: `npm audit --audit-level=high` exits 1 with **15 high-severity findings** (34 total), so every contributor has had to bypass it by hand. This PR takes it to **0 high** in both the root and `observio-sample-agent` (the sub-project with its own lockfile; there are no npm workspaces).

| Package | Sev | Path (root) | Fix | Compatibility evidence |
|---|---|---|---|---|
| js-yaml | high | direct (prod) | `npm audit fix` → 4.3.2 | in-range patch |
| postcss | high | direct (dev) | `npm audit fix` → 8.5.28 | in-range patch |
| nanoid | high | transitive (@assistant-ui, postcss) | `npm audit fix` → 5.1.16 / 3.3.18 | in-range |
| vite | high | direct (dev) | `npm audit fix` → 7.3.6 | in-range patch; build:all + e2e green |
| react-router / -dom | high | direct (prod) | `npm audit fix` → 7.18.3 | in-range minor; e2e routing specs green |
| browserslist | high | transitive (autoprefixer, babel) | `npm audit fix` → 4.28.9 | in-range |
| fast-uri | high | transitive (ajv via shadcn) | `npm audit fix` → 4.1.4 | ajv's own range allows 4.x |
| form-data | high | transitive (superagent) | `npm audit fix` → 4.0.6 | in-range |
| ip-address | high | transitive (express-rate-limit, socks) | `npm audit fix` → 10.7.0 | in-range |
| brace-expansion (root copy) | high | transitive (minimatch) | `npm audit fix` → 5.0.9 | in-range |
| puppeteer / puppeteer-core / @puppeteer/browsers / extract-zip | high | **optional** direct | **major bump** puppeteer `^24.37.4` → `^25.10.0` | see below |
| undici + nested brace-expansion | high | **optional** direct (`@earendil-works/pi-coding-agent`'s shrinkwrap) | **bump + exact pin** pi `^0.80.2` → `0.84.4` | see below |

**Why the two majors were unavoidable**

[...]

PR #495, opened 2026-09-09
fix(metrics): service-name+window correlation for comparison metrics (parity with the trace judge) + REST report run-id recovery + telemetry smoke gate

goyamegh:
## The problem, as seen on the comparison page

Two benchmark runs of a REST-connector agent (62 cases each, same benchmark, run this morning) came up on the comparison page with **Cost, Tokens and LLM Calls blank (`--`)** — while two other arms of the same comparison (a subprocess agent and a different REST agent) had every column filled in. The blank arms' spans were on the observability cluster the whole time: **1,033 spans / 63 traces** for one of the runs, every span carrying the OTel `gen_ai.conversation.id` the agent stamps, under its configured `service.name`. The run-report **Traces tab showed them. The trace judge used them.** Only the metrics were blank.

## Why they were blank

Each of those runs' reports was persisted with **no correlation id at all**: `runId: null`, `sessionId: null`, `traceId: null`.

- **No `traceId`** — the server had been booted from a config that had lost its `telemetry` block, so no eval `test_case` span was started and nothing stamped `report.traceId` (Strategy A).
- **No `runId`** — the agent echoes its per-request id in the response body under a `session_id` field. `RESTConnector.execute()` recognizes only `runId`/`id`, and the agent's `afterResponse` hook did not (yet) map the echoed field onto `runId` (Strategy B). The very same id is stamped on every span the agent emits.
- **No `sessionId`** — Strategy D is only captured by subprocess connectors.

So how did the Traces tab and the judge still find the spans? Because `/api/traces` accepts one more hint: `agents: [{serviceName, startedAt, endedAt}]` — **Strategy C, the agent's OTel `service.name` plus the run's wall-clock window** — and unions it with A/B/D. `/api/metrics/batch` never accepted that hint; it took only `runIds` / `sessionIds` / `traceIds`. And the comparison page skipped every report that had no `runId` before it even asked.

Two readers of the same cluster, two vocabularies. Reports with no id could be *shown* but not *measured*.

## The generic fix (commit 1) — metrics parity with the trace judge

No agent-, service-, id-prefix-, run- or date-specific logic anywhere; the same code path serves every agent.

**Server**
- `tracesService.ts`: the per-hint Strategy-C/D clause is extracted out of `fetchTraces` into `buildServiceWindowClause` / `buildAgentHintClause` (byte-identical DSL — a unit test asserts `fetchTraces` emits exactly the helper's output). Both readers now build through it, so they cannot drift again.
- `metricsService.ts`: `computeMetrics(.., agents?)` / `computeBatchMetrics(.., agentsByRunId?)` OR the hints into the existing `bool.should` union. Batch attribution tries the **precise** strategies first (run-id attributes, `session.id`, `traceId`) and falls back to the service window; when two windows for one service overlap (the same agent evaluated concurrently, or slack spilling into a neighbour), the span goes to the **nearest window midpoint**, so every span is counted for exactly one key and totals stay exact. `serviceName` is added to the batch `_source` projection — without it the query *matched* the span but grouping couldn't see which service it came from (caught by the new fake-OpenSearch integration test; same failure class #469 fixed for token attributes).
- `routes/metrics.ts`: `POST /api/metrics/batch` accepts `agents: Record<key, [{serviceName, startedAt, endedAt, sessionId?}]>` (validated, 400 on malformed); `GET /api/metrics/:runId` accepts `?serviceName=&startedAt=&endedAt=`. Keys need not be run ids — results come back under whatever key the caller used.

**Comparison page**
- `metricsKeyForReport`: `runId`, else the report's own id — reports without a runId are keyed, not dropped.
- `collectAgentHintsFromReports` derives the hint through **the judge's own `buildJudgeAgentsHints`** (`connectorProtocol` / `agentKey` / timestamp / `durationMs` / `sessionId`, plus the agent's configured `traceServiceName` looked up from the agent config — the same lookup the Traces tab performs). A generic transport protocol with no `traceServiceName` gets **no hint** — never a fabricated service name.
- Stable `data-testid`s on the Cost / Duration / Tokens / LLM Calls / Tool Calls cells.

## Recovery of the already-persisted reports (commit 2) — no re-run

`scripts/backfill-report-run-ids.ts` (pure rule in `lib/reportRunIdRecovery.ts`): for REST-connector reports with an empty `runId`, resolve the echoed id with the **same rule the live path uses** — the agent's own `afterResponse` hook replayed from a loadable config (`--config-dir`) is authoritative; otherwise the first non-empty string under a configurable candidate-field list (`--id-fields`, default `runId,run_id,session_id,sessionId,conversation_id,conversationId,id`). Dry-run by default; `--apply` PATCHes through the normal storage route; already-populated reports are skipped (idempotent); unknown flags are refused; nothing is ever deleted.

**Before → after on the two affected runs (124 reports):**

| | before | after |
|---|---|---|
| reports with `runId` | 0 / 124 | **119 / 124** (5 are failed runs with no agent response — correctly left alone) |
| `/api/metrics/batch` `hasSpans: true` | 0 | **119 / 119 (100%)** |
| reports with tokens > 0 and LLM calls > 0 | 0 | **119 / 119 (100%)** |
| run A on the comparison page | `--` / `--` / `--` | **$20.19 · 5.9M tokens · 312 LLM calls** |
| run B on the comparison page | `--` / `--` / `--` | **$26.80 · 7.6M tokens · 373 LLM calls** |

Re-running the script afterwards: `resolved=0 applied=0 alreadyHadRunId=119` — a no-op.

## Never again: the telemetry smoke gate (commit 3)

`scripts/smoke-telemetry.ts` (`npm run smoke:telemetry -- --base <url> --agent-key <key>`) runs ONE ad-hoc case and asserts the report correlates with its traces: `report.traceId` set (eval span on) · `report.runId` set for REST connectors · spans found via `/api/traces` by A or B · `/api/metrics/batch` `hasSpans` (SKIP without an observability cluster). Exit 1 on any FAIL; deletes the one report it created. Verified: **FAIL** against a server with telemetry disabled (exactly this incident's state), **PASS** against the same server with the eval span exporting to its own OTLP receiver. Documented in `docs/CONFIGURATION.md` as the mandatory step after every server restart / agent relaunch.

## Tests (synthetic fixtures only)

- **Unit** — exact Strategy-C clause DSL and its union with A/B/D; batch attribution incl. the overlap tie-break and precise-beats-window; `_source` keeps `serviceName`; route validation + threading of `agents`; comparison collectors keyed by report id; hint derivation via the judge's helper; recovery rule (hook authoritative, field list configurable, throwing hook degrades); both scripts as real subprocesses against mock backends (dry-run writes nothing, `--apply` patches exactly the candidates, second pass no-op, every smoke check's PASS/FAIL/SKIP/WARN branch, exit codes).
- **Integration** — fake OpenSearch holding flat-@ and nested span docs that carry **no** ids: found by the service window alone, attributed to the right key across both schemas, no decoy bleed, and `fetchTraces` returns the same spans for the same hint (parity).
- **E2E** — the compare page requests a no-runId report under its report id with a `serviceName` + symmetric window hint identical to the judge's derivation, renders the returned Cost / Tokens / LLM Calls, and stays an honest `--` when no spans come back.

Stacked on #490 (`attributeFieldPaths` / `buildRunIdShouldClauses` helpers); merge #490 first.

## Adversarial review (`codex_review`, gpt-5.4) — findings and disposition

|---|-----|---------|-------------|
| 1 | HIGH | Strategy C widens the query; the `size` caps (500 single / 10,000 batch) can silently truncate and undercount while the UI shows exact numbers. | **Applied.** Results whose query hit the cap are flagged `partial`; propagated to `RunAggregateMetrics.traceMetricsPartial`; the scoreboard renders those cells as `≥…` with a "lower bound" tooltip. |
| 2 | HIGH | The report-id surrogate key was also fed into the Strategy-B run-id `terms` clause — conflating "request key" with "correlation id". | **Applied.** Window-only keys are never sent as run-id correlators (`runIdCorrelators`); same rule on the single-run path; a test proves a span carrying the surrogate string as its run id is not attributed. |
| 3 | MED | `runIds` elements weren't validated (null/number → 500); inverted windows accepted. | **Applied.** 400 on non-string/empty `runIds` elements and on `endedAt < startedAt`. |
| 4 | MED | Window-correlated numbers rendered with no confidence signal. | **Applied (tooltip), not a badge.** `MetricsResult.correlatedBy` (`ids`/`window`/`mixed`) is returned; window-correlated runs get a tooltip note on the trace cells. A per-cell visual badge is a separate UX decision — left out deliberately. |

Verdict before hardening was "not review-ready for quantitative metrics until keying is separated from correlation, partial results are surfaced, and `runIds` is validated" — all three are now done.

## Verification

- `npm run build:all` ✅ · `npm run test:unit` ✅ 370 suites / 6,380 tests · `npm run test:integration` — 952 passed; the 6 failures (benchmark cancel, storage stats, CLI judge-model flag, Kiro connector) fail identically on the base branch against the same zero-config test server (file storage, no `kiro-cli`) — pre-existing/environmental · `npm run test:e2e` ✅ 523 passed, 4 flaky-then-passed, 10 skipped · `npm audit --audit-level=high` — lockfile byte-identical to `main`; the reported highs are pre-existing and addressed by #488 · DCO on every commit · CHANGELOG under `[Unreleased]` (Fixed + Added).
- Live proof (internal data, kept out of this PR): after recovery, the batch metrics API returned `hasSpans: true` with tokens and LLM calls for 119/119 recovered reports, and the comparison page rendered Cost / Tokens / LLM Calls for both previously-blank runs.


goyamegh:
## Verification proof (text only; synthetic / generic where it touches live data)

### 1. Unit / integration / e2e
```
npm run build:all                     exit 0
npm run test:unit                     Test Suites: 370 passed, 370 total
                                      Tests:       5 skipped, 6380 passed, 6385 total
npm run test:e2e (full suite)         523 passed, 4 flaky (passed on retry), 10 skipped
npm run test:integration              952 passed, 6 failed, 7 skipped
  -> the 6 failures (benchmark cancel, storage stats, CLI --judge-model x2, Kiro connector x2)
     fail identically on the base branch against the same zero-config test server
     (file storage, no kiro-cli): pre-existing / environmental, unrelated to this change.
```

### 2. The fake-OpenSearch integration test that caught a real bug during development
The first version of the batch path MATCHED window-only spans but attributed none of them — the `_source` projection stripped `serviceName`, so grouping couldn't see which service a span came from (the same projection-strips-what-we-need class #469 fixed for token attributes). The test failed, the projection was fixed, the test now pins it:
```
✓ pre-fix reproduction: with no ids, the correlation union is empty and NOTHING matches
✓ computeMetrics finds an OSI flat-@ run by service window alone and reads its tokens / model / tools
✓ computeBatchMetrics (the /api/metrics/batch path, with _source projection) attributes each run's spans
  to its own key across BOTH index schemas, without decoy bleed
✓ a key with a hint whose window contains no spans stays an honest pending placeholder (hasSpans:false)
✓ parity: fetchTraces (Traces tab / judge) returns the SAME spans for the SAME hint that metrics counted
✓ the union is additive: a run that DOES have a run id is still found by Strategy B when a hint is also present
```

### 3. Route validation, live against a local server (after the review hardening)
```
POST /api/metrics/batch {"runIds":["a",null]}
  -> 400 {"error":"runIds must contain only non-empty strings"}
POST /api/metrics/batch {"runIds":["a"],"agents":{"a":[{"serviceName":"x","startedAt":5,"endedAt":1}]}}
  -> 400 {"error":"agents[].endedAt must not be before startedAt"}
```

### 4. Recovery script — dry-run, apply, idempotent re-run (two real 62-case runs of a REST-connector agent; ids redacted)
```
DRY-RUN : runs=2 scanned=124 alreadyHadRunId=0   notRest=0 resolved=119 unresolved=5 applied=0   failed=0 bySource={"hook":119}
APPLY   : runs=2 scanned=124 alreadyHadRunId=0   notRest=0 resolved=119 unresolved=5 applied=119 failed=0 bySource={"hook":119}
RE-RUN  : runs=2 scanned=124 alreadyHadRunId=119 notRest=0 resolved=0   unresolved=5 applied=0   failed=0 bySource={}
  (the 5 unresolved are failed reports with no agent response body — correctly left untouched)

POST /api/metrics/batch with the 119 recovered run ids:
  run A: reports=62 hasSpans=62 (100%) tokens&llmCalls>0=62 (100%)  avgLlmCalls=5.0
  run B: reports=57 hasSpans=57 (100%) tokens&llmCalls>0=57 (100%)  avgLlmCalls=6.5
Comparison page: Cost / Tokens / LLM Calls populated for both runs (were "--" before).
```

### 5. Telemetry smoke gate — proves the gate itself discriminates
Same local server, same mock agent; the only difference is whether the eval span is on.
```
# telemetry DISABLED (the state that caused this incident)
  [PASS] evaluation ran
  [FAIL] report.traceId set (eval span on) — missing — is the telemetry block present in the config ...
  [SKIP] report.runId set (REST agent echoed an id) — connector is mock
  [FAIL] spans found via /api/traces (A or B) — 0 spans after 5000ms
  [SKIP] /api/metrics/batch hasSpans — no observability cluster configured
TELEMETRY SMOKE: FAIL        (exit 1)

# telemetry ENABLED, eval span exported to the server's own OTLP receiver
  [PASS] evaluation ran
  [PASS] report.traceId set (eval span on) — 5da3ff13602f129fd07a9bdbfd851a9b
  [SKIP] report.runId set (REST agent echoed an id) — connector is mock
  [PASS] spans found via /api/traces (A or B) — 1 span(s)
  [SKIP] /api/metrics/batch hasSpans — no observability cluster configured
TELEMETRY SMOKE: PASS        (exit 0)
```

TackAdam (approved):

PR #494, opened 2026-09-09
feat(judge): record and display the underlying LLM behind the agent trace judge; optional pin

goyamegh:
## Description

`judgeModelId: 'agent-trace-judge'` names a **provider**, not a model. The LLM behind it is picked at run time by `pickJudgeModel()` over the credentialed pi model registry — today that resolves to `amazon-bedrock/global.anthropic.claude-sonnet-4-5-20250929-v1:0` (Claude Sonnet 4.5; the 4.x scorer ties Sonnet 4.5/4.6 and prefers the global inference profile). That choice was recorded only on `llmJudgeResponse.judgeDebug.modelId`, which is `undefined` unless `AH_JUDGE_DEBUG=1` — so every agent-trace-judge report persisted `judgeModelId` **and** `llmJudgeResponse.modelId` as `agent-trace-judge`, and nothing on the run said which LLM actually judged it.

### Data (the part that must be right now)
- `JudgeResponse` gains `judgeModel` (resolved underlying LLM) + `judgeProvider` (judge kind), set by **every** provider: the agent judge reads the model the pi transcript reports (`assistant.model` on the verdict turn, `responseModel` fallback); the pi CLI judge reads it off its NDJSON; Bedrock / OpenAI-compatible record their effective model id (gateway-echoed `model` when present).
- Persisted as **`TestCaseRun.judgeModel`** (new, `keyword`-mapped) next to `judgeModelId` (unchanged = configured judge kind). `llmJudgeResponse.modelId` now carries the **real** model with `llmJudgeResponse.judgeProvider` keeping the kind. Run-level `BenchmarkRun.judgeModel` / `EvaluationRun.judgeModel` from the first report that resolved one.
- `lib/judgeIdentity.ts` is the single derivation: a provider pseudo-id (`agent-trace-judge`, `pi-judge`, `agentic-*`, …) is never written as a model; a plain Bedrock judge's `judgeModel` trivially equals its configured id. Old reports have no `judgeModel` (never persisted, not recoverable) — the UI says "model not recorded — auto-picked at run time" rather than guessing.
- Telemetry: `agent_health.judge.model_id` keeps the configured id (stable cardinality); the resolved LLM goes on a new `agent_health.judge.model`.

### Selectable + deterministic
[...]

PR #486, opened 2026-09-05
fix(runs): terminal-aware run stats (no phantom pending on cancelled/failed runs) + race-free run-doc updates

goyamegh:
## Summary

Two live symptoms on the runs list, one display bug and one real data-integrity bug behind it:

- **(S1)** A run cancelled at 34/62 rendered a **Cancelled** badge *and* "28 pending" with an in-progress spinner — forever.
- **(S2)** A concurrency=3 re-run ended `failed` at 37/62; five per-case results read `Evaluation error: version_conflict_engine_exception … [eval-run-…]`, and the row showed `0% 0/5 / ⚠32 / 25 ⟳`.

## Root causes (confirmed in code)

| | Where | What |
|---|---|---|
| R1 | `lib/runStats.ts` `bucketRunResults()` | folded `plannedTotal − results.length` into `pending` **regardless of the run's terminal status**. |
| R2a | `server/adapters/opensearch/StorageModule.ts` `OpenSearchEvaluationRunOperations.update()` | read-modify-write **full overwrite** (`getById` → spread → `index()`), no optimistic concurrency — every cancel/rename/terminal write re-indexed a stale snapshot of the whole doc (incl. `results`) over concurrently scripted per-case entries. |
| R2b | same file, `updateResult()` | scripted `_update` with **no `retry_on_conflict`** → this is the exact `version_conflict_engine_exception` source: two of N concurrent cases finish in the same refresh window, the loser throws. |
| R2c | `services/evaluationRunner.ts` | awaited that bookkeeping write **inside the evaluation try/catch**, so the storage race was misfiled as an evaluator failure — the already-judged report was rewritten to `metricsStatus:'error'`, and when the retry inside the catch conflicted again the whole run went `failed`. |
| R2d | `server/routes/storage/evaluationRuns.ts` (both execution routes) | terminal `update()` re-sent the in-memory `results` map wholesale + stats computed from it. |

## Fix

- **Terminal-aware stats.** `bucketRunResults(results, plannedTotal, runStatus)` / `computeRunStats(run)`: on a terminal run the never-started remainder, dead-executor `pending`/`running` entries, and explicit `status:'cancelled'` entries bucket into a new `RunStats.notRun` (invariant `total = passed+failed+errored+pending+notRun`). Legacy denormalized `run.stats` with stale `pending` on a terminal run is re-homed the same way. New `passRateOverJudged()` (passed / judged; errored + notRun excluded per the #242 convention, footnoted on the page) for the detail page.
[...]

PR #487, opened 2026-09-05
test(e2e): restore a green e2e suite on main — fixture rot + assertions updated for merged UI changes

goyamegh:
## Why

The `e2e-tests` CI job has been **red on `main` itself** — [latest main run](https://github.com/opensearch-project/agent-health/actions/runs/33817338957/job/100852285777): 5 failed / 3 flaky; on current HEAD (`ac025b03d`) 7 failed / 5 flaky — so every open PR inherits the same failures regardless of what it touches, and `e2e-gate` can't be trusted as a merge gate.

Each failing spec was reproduced locally against a fresh **file-storage** server (exactly CI's setup: `CI=1`, no `OPENSEARCH_*`), root-caused, and repaired on the right seam. **No test was deleted or skipped; no assertion was loosened without the reason in the commit.** Two of the failures turned out to be real product regressions.

## Per-spec root cause

| Spec | Class | Cause → fix |
|---|---|---|
| `judge-tab-whyfix` | **Product regression** | #442's merge conflict resolution took the pre-#454 `MatcherResultsPanel.tsx` wholesale, silently reverting #454's Why/Fix judge-row redesign on `main` (duplicated reasoning + fabricated "score 0%" were back; #454's unit suite replaced; `judgeReasoningParse.ts` orphaned). Restored the #454 panel and layered #442's `notReached` rows on top. Unit suite = #454 + #442 + 1 new case. |
| `skills` (2 tests) | **Product bug** (surfaced locally; CI's runner has no user-scope skills so it never saw it) | `GET /api/skills/discover` lists `~/.claude/skills/*` with a `~/…` path, but `resolveSkillPath()` resolved it against cwd → "Directory does not exist" and Run Evaluation stayed disabled. `~`, `~/`, `~\` now expand to home (`~backup/` deliberately not). Unit + integration tests (red against the unfixed server). |
| `benchmark-run-row-link` (2) | Stale assertion | #460 replaced the run-name `<h2>` with `InlineRenameField` → assert `run-inspector-rename-text`. Fixture's nonexistent agent is intentional (deterministic "failed before linking" shape) and unchanged. |
| `evalruns-no-left-status-icon` | Stale assertion | #460 added the rename pencil `<svg>` in the name cell → `svg:not(.lucide-pencil)` count 0 (the #430 status-icon guard still bites). |
[...]

PR #497, opened 2026-09-09
feat(runs): record the agent configuration fingerprint on every run and report (prompt provenance)

goyamegh:
## Problem

A run document records `agentKey`, but not **which version of that agent's configuration** produced the numbers. The agent's behavior lives in `agent-health.config.ts` (system prompt, allowed tools, model, MCP servers, hooks…) and changes over time. When a system-prompt edit lands between two runs of the same agent, "which prompt produced this pass rate?" becomes a git-archaeology question — and the comparison page happily shows two "same agent" runs side by side with no hint that they measured different configurations.

## What this does

Every run-creation path — `POST /api/storage/evaluation-runs`, its `/rerun`, and the legacy `POST /api/storage/benchmarks/:id/execute` — stamps, **at creation time**, the provenance of the agent config it resolved:

| Field | What |
|---|---|
| `agentFingerprint` | sha256 over the connector-relevant, secret-redacted agent config (see below) |
| `agentFingerprintShort` | first 12 hex — for chips/labels |
| `agentPromptHash` | sha256 over the system/append prompt **alone** (when derivable) — so a prompt-only change is distinguishable from "other config changed" |
| `agentConfigSource` | `realpath` of the authored config file (it is commonly a symlink) + `git rev-parse HEAD` of its repo + dirty flag; best-effort, informational |

The three hash fields are mirrored onto **every per-test-case report** (and the `benchmark.runs[]` projection) so case-level comparisons work too. Runs and reports created before this carry none of the fields; every reader tolerates that.

### Hashed field set

`connectorType`, `endpoint` (incl. the run-level `agentEndpoint` override — it changes what the connector talks to), resolved model (`connectorConfig.model` / `env.ANTHROPIC_MODEL` / `args --model`, same resolution the runner uses), `useTraces`, `traceServiceName`, `tracePolling`, the whole `connectorConfig` (`systemPrompt`, `appendSystemPrompt`, `allowedTools`, `disallowedTools`, `additionalArgs`, `args`, `mcpServers`, `traceContext`, `workingDir`, …), and each hook's **source text** via `Function.prototype.toString` (whitespace-normalized — a reformat is not a change).

### Redacted **before** hashing (by design)

- Header **values** and `auth` **secrets** — never participate. Header **key names** and the auth **shape** (`type`, region, service) do, so an added `X-Api-Version` header or a bearer→sigv4 switch is visible while rotating a credential is not. Run-level header overrides never participate.
- `connectorConfig.env` and MCP-server `env` — **key names only**; values are replaced by a marker, except non-secret `*MODEL*` selectors (e.g. `ANTHROPIC_MODEL`) whose values *are* behavior.
- Credential-carrier keys anywhere in `connectorConfig` (`isSecretKey`: exact names like `token` / `apiKey` / `password` / `secret` / `authorization` / `cookie` / `jwt` in any separator style, plus `*Token` / `*Secret` / `*Key`-style suffixes) — value replaced by a marker. Deliberately **not** a substring match: `maxTokens`, `tokenBudget`, `authorizationMode`, `privateMode` are behavior and their drift stays visible.
- Hook source text is hashed **verbatim** (no whitespace normalization — it would collapse whitespace inside string literals). Stated limit: source text cannot see closed-over variables, so the fingerprint proves "the hook's code changed", never "behavior is identical".
- `name` / `description` / `enabled` / `isCustom` — not hashed (they don't change agent behavior).

Serialization is deterministic for Map/Set/Date/URL/RegExp/BigInt/NaN/class instances; a circular config throws → the route logs a warning and stamps nothing (provenance is metadata, never a reason to refuse a run).

Net effect: rotating a credential never changes the fingerprint; adding/removing an env var or header, swapping the model, changing the auth type, editing the prompt, or changing a hook body does.

## UI

- **Runs list** (agent cell), **run inspector** header and **run detail** header: a mono 12-hex chip; tooltip = full fingerprint, prompt hash, config path `@ sha` (+ "uncommitted edits" when dirty). The detail page's *Run Configuration* panel gets an "Agent config" row.
- **Comparison scoreboard**: when both selected runs are the **same `agentKey`** but carry **different fingerprints**, an amber badge **"config changed between runs · prompt | other fields"** appears on row A (and on the condensed band). Tooltip says whether the system prompt changed or only other connector fields did, with both full hashes. Different agents never trigger it (differing configs *are* the comparison); legacy runs never trigger it (unknown is not a warning).
- **Re-run**: the new run naturally gets the *current* fingerprint; if it differs from the source run's, a **"config changed since source run"** badge sits next to the existing re-run provenance chip.
- Run dialog: nothing to add — computed server-side.

## API

`GET /api/agents/:key/fingerprint` (optional `?agentEndpoint=` mirroring the run-level override) returns the provenance a run started right now would carry, so external tooling can record it up front and compare with `run.agentFingerprint` afterwards. Returns hashes only — never config contents. The evaluation-runs list and the report summary batch include the fields. The config-source git lookup is async (`execFile`, argv array, 2 s timeout — never blocks the event loop) and uncached.

## Tests

- **Unit** — `tests/unit/lib/agentFingerprint.test.ts`: same config → same hash (key-order independent); prompt change → fingerprint + promptHash change; allowedTools/additionalArgs/model/mcpServers/traceContext → fingerprint changes, promptHash stable; header-value/auth-secret/env-value/MCP-env-value/secret-key change → **unchanged**; header key added / auth type or region changed / env key added → changes; hook body change → changes, whitespace inside a string literal → changes; `isSecretKey` exact/suffix/negative cases (`maxTokens`, `authorizationMode` not redacted); Map/Set/Date/BigInt/class-instance determinism; cycle → clear error; pre-hash payload never contains a raw secret; symlink → realpath + git sha + dirty. Route/service: `evaluationRunsAgentFingerprint.test.ts` (create/rerun stamp with the real hasher, two runs across a prompt edit differ, header-only edit doesn't, projection onto `benchmark.runs[]`, unknown agent → run still created, API equals the stamp and leaks nothing), `benchmarksExecuteAgentFingerprint.test.ts`, extended `evaluationRunner.test.ts` / `benchmarkRunner.test.ts` (report mirror; config source never on reports; legacy runs stamp nothing). RTL: `AgentFingerprintChip.test.ts`, `ComparisonScoreboard.agentFingerprint.render.test.ts`, extended `RunInspectorPage.test.ts` and `EvalRunsPage.test.ts`.
- **Integration** — `evaluationRunsAgentFingerprint.integration.test.ts` against a real backend: run + every report carry the fingerprint and equal `GET /api/agents/demo/fingerprint`; same config twice → same; list projection includes it; endpoint override differs.
- **E2E** — `agent-fingerprint-provenance.spec.ts`: runs-list chip + legacy no-chip, detail + inspector header chip, re-run "config changed since source run", scoreboard badge for the same agent / none for different agents.

## Composability

Compose-checked against open #468 (run actions), #486 (run-status integrity) and #494 (judge model identity): only adjacent-line take-both-sides conflicts (import lines, spread additions, interface field appends); the composed trees type-check and this PR's tests pass on each.


goyamegh:
## Verification proof (local, against a real OpenSearch-backed server)

**Type-check / build**
```
npx tsc --noEmit -p tsconfig.json   → exit 0
npm run build:all                   → exit 0 (vite + server + cli + lib)
```

**Unit** (full suite, `npm run test:unit`)
```
Test Suites: 372 passed, 372 total
Tests:       5 skipped, 6400 passed, 6405 total
```
Post-review re-run of the touched suites (fingerprint lib, routes, runners, RTL, server services):
```
Test Suites: 70 passed, 70 total
Tests:       4 skipped, 1273 passed, 1277 total
```

**Integration** — `tests/integration/server/routes/storage/evaluationRunsAgentFingerprint.integration.test.ts`, real backend (OpenSearch storage, built-in mock `demo` agent):
```
✓ GET /api/agents/:key/fingerprint returns a sha256 fingerprint (+ 404 for an unknown agent)
✓ create run → run.agentFingerprint AND every report.agentFingerprint are set and equal the current agent fingerprint
✓ two runs of the same agent under the same config carry the SAME fingerprint; the list projection includes it
✓ a run-level agentEndpoint override yields a DIFFERENT fingerprint than the plain agent
Tests: 4 passed, 4 total
```
Legacy path, `POST /api/storage/benchmarks/:id/execute` with the `demo` agent — persisted `benchmark.runs[]` entry and its report doc:
```
runs[0]: {"id":"run-…","agentFingerprintShort":"8ca6ec042376"}
report:  {"agentId":"demo","agentFingerprint":"8ca6ec04237660288345ba980fd4d458dbc5c5c049cf72139ee770aa44cf86c5","agentFingerprintShort":"8ca6ec042376","agentPromptHash":null}
```
(`agentPromptHash` is legitimately absent — the built-in demo agent has no system prompt.)

`GET /api/agents/demo/fingerprint`:
```json
{"agentKey":"demo","agentFingerprint":"8ca6ec04237660288345ba980fd4d458dbc5c5c049cf72139ee770aa44cf86c5","agentFingerprintShort":"8ca6ec042376","computedAt":"…"}
```
`GET /api/agents/<unknown>/fingerprint` → `404`.

Full `npm run test:integration` (parallel, shared cluster): 900 passed; the 44 failures were all `Exceeded timeout` / `CLI subprocess timed out` under parallel load plus the `kiro-cli` connector suite (needs a live kiro session) — every timed-out suite passes when re-run serially (`--runInBand`), and the one `CI=1`-only failure (`judgeModelIdFlag` refusing to reuse an already-running server in CI mode) passes without `CI=1`. None asserts anything this PR touches.

**E2E** — `tests/e2e/agent-fingerprint-provenance.spec.ts` (Playwright, real backend):
```
✓ runs list shows the mono fingerprint chip for a fingerprinted run and none for a legacy run
✓ run detail page header shows the fingerprint chip (+ the Run Configuration panel row)
✓ run inspector header shows the fingerprint chip
✓ a re-run whose fingerprint differs from its source shows "config changed since source run"
✓ comparison scoreboard warns "config changed between runs · incl. prompt" for the SAME agent with different fingerprints
✓ comparison scoreboard shows NO config-changed badge for two DIFFERENT agents
6 passed
```
Remaining full-suite e2e failures on this branch are the pre-existing fixture-rot cases already fixed by #487 (`benchmark-run-row-link`, `comparison-benchmark-free`, `comparison-hover-prompt-preview`, `comparison-search-popover-anchor`, `dashboard`, `evalruns-no-left-status-icon`) plus `comparison-insights-band`, whose seed PUT 500s on a cluster-specific `expectedOutcomes` mapping conflict — none involve this change.

**Compose check** against open #468 / #486 / #494: merging this branch into each yields only adjacent-line take-both-sides conflicts (imports, spread additions, interface field appends); the composed trees type-check and this PR's tests pass on each.

**Mapping rollout** observed on an existing cluster: `putMapping` applied the new keyword fields to the reports index and to the nested `runs.*` on the benchmarks index cleanly (`[storageInit] All indexes initialized successfully`, no mapping warning).

TackAdam (approved):

PR #500, opened 2026-09-10
fix(evals3): Add Run button tracks the run document, not the SSE connection (no more stuck "Running…")

goyamegh:
## Owner clarification (2026-09-13): Add Run must never be blocked by a running run

Multiple concurrent runs on one benchmark are the normal flow (2–4 agent/model arms launched back-to-back). The first cut of this PR fixed the stuck-after-completion spinner but still turned the button into a disabled "Running…" for the whole run, forcing a refresh before the next launch. Now:

- **Add Run is never disabled by running runs.** It pauses only for the ~1 s window between the POST and the `started` event (bounded by a 30 s `started` timeout so a stalled intermediary can't turn it into a forever-disable), then re-enables. The "A run is already in progress" alert guard is removed.
- The header shows a non-blocking **`● N running` pill** (`runs-in-flight-pill`) counting every run whose effective status is `running` — launched here or anywhere else — from the polled docs; clicking it narrows the Runs table to `status: running` (exactly N rows).
- **One progress block per launched run** (`launchedRuns: {runId, name, launchedAt}[]`): collapsible, name + `k/N` + per-case list; launching a second run never resets/hides the first; each block drops out on its own when its polled doc is terminal (30 s grace for a doc that never appears). The SSE-drop → polling fallback stays.
- Edit keeps its pre-existing gate (disabled while runs launched from this page are in flight); a POST that fails before `started` shows a dismissible inline error.

New e2e: `benchmark-add-run-header-reset.spec.ts` → "THE REGRESSION: with a run in flight the header stays an ENABLED "Add Run" with a "● 1 running" pill; launching a second run → "● 2 running" and two progress blocks, the first untouched".

---

## Problem (user-visible)

On the benchmark details page → Runs tab, clicking **Add Run** and launching a run left the header button spinning **"Running…"** indefinitely. The run itself completed fine on the server (rows on the same page showed it finished), but the header never recovered — the only way out was a page refresh. Reproduced with 62-case runs (30–60+ min).

## Root cause

The header button was bound to a client-local `isRunning` boolean that was reset **only when the launching `POST /api/storage/evaluation-runs` SSE stream ended** (`finally` in `handleStartRun`). Two things conspired:

1. **No heartbeat on that stream.** Between per-case events a long run's stream is idle for minutes. Idle proxies / tunnels / the Vite dev proxy / browsers close (or stop relaying) a silent connection long before `completed` is emitted — the server deliberately treats the stream as an observer (`sendSSE` swallows write failures; the run continues).
2. **Connection-bound UI state.** When the stream went silent, `reader.read()` never resolved and nothing else ever reset `isRunning`; when it closed, the page flipped the other way (button back to "Add Run" while the run was still running server-side, and the progress panel marked remaining cases *failed*).

The per-row status on the same page was already correct because it comes from the polled run documents. The header was the one thing still bound to the fragile connection.

## Fix

**Client — the header derives from the run DOCUMENT, not the connection**
- New `lib/runLaunchState.ts` → `deriveAddRunButtonState({launching, launchedRunId, launchedAt, runs})` → `'idle' | 'launching' | 'running'`. `launching` until the `started` event yields a runId, then `running` exactly while the polled doc for that runId is non-terminal (`isTerminalRunStatus`: completed/failed/cancelled). A doc that never appears is `running` for a bounded 30 s grace, then `idle` (no second way to spin forever).
- `BenchmarkRunsPage.tsx`: `isRunning`/`setIsRunning` replaced by `isLaunching` + `launchedRun` state; the button (now `data-testid="add-run-button" data-run-state=…`) and the `alert('A run is already in progress.')` guard use the derived state. When the SSE stream ends/errors **after** `started`, we `debug()`-log and fall back to polling — no failed flip. When the stream drops, the progress panel follows the polled doc's per-case `results`.
- Nothing else — no progress-panel redesign, no new endpoints.

**Server — keep the stream alive in the first place**
- `POST /api/storage/evaluation-runs` writes an SSE comment heartbeat `: ping\n\n` every 15 s while the run executes (`EVALUATION_RUN_SSE_HEARTBEAT_MS`; `0` disables), cleared in `finally`, guarded with the same `res.destroyed || res.writableEnded` check as `sendSSE`. Verified the client parser skips data-less frames.

## Test matrix (all red on the old code where they assert the bug, green now)

| Level | File | What it locks |
|---|---|---|
| Unit | `tests/unit/lib/runLaunchState.test.ts` | idle / launching / running / terminal→idle / missing-doc grace |
| Unit | `tests/unit/services/client/evaluationRunsApi.sse.test.ts` | `: ping` frames (incl. split across chunks) ignored → `completed` still returned; stream ending w/o `completed` still throws (page hands over to polling); `started` fires before the throw |
| Unit | `tests/unit/server/routes/storage/evaluationRuns.test.ts` | heartbeat written at interval, none after completion, cleared on error path, disabled at `0`; `startSseHeartbeat` fake-timer + dead-socket/`EPIPE` guards; env parsing |
| Unit (render) | `tests/unit/components/evals3/BenchmarkRunsPage.test.ts` | stream dropped → "Running…" while doc runs → "Add Run" once terminal (no reload, no failed flip); doc cancelled elsewhere resets header while the stream is still open; pre-`started` failure resets immediately; progress panel follows polled results; in-progress guard |
| Integration | `tests/integration/server/routes/storage/evaluationRunsSseHeartbeat.integration.test.ts` | real server, `demo` agent: raw SSE body has `: ping` frames interleaved, every frame is a well-formed event or a comment, `event: completed` is last |
| E2E (Playwright) | `tests/e2e/benchmark-add-run-header-reset.spec.ts` | (1) POST SSE sends `started` then **closes** without `completed`, polled list says `running` → header "Running…"/disabled for several poll cycles → list says `completed` → enabled "Add Run" within a poll interval, **no reload**; (2) POST SSE stays **open but silent** after `started` (the live-tunnel manifestation) → same reset while the stale stream is still open; (3) happy path |

### Gates
```
npm run build:all      ✓
npx tsc --noEmit       ✓
npm run test:unit      ✓  371 suites, 6420 passed
integration (touched)  ✓  7 suites / 48 tests incl. the new heartbeat test (file storage, port 4720)
e2e (benchmark page)   ✓  59 passed — new spec + benchmark-runs-table + evals3-benchmark-runs + benchmark-runs + delete-all-rows + associated-evalruns + passed-count + version-filter + stats-refresh + agent-compatibility
coverage               ✓  lib/runLaunchState.ts 100%; every changed line in BenchmarkRunsPage.tsx / evaluationRuns.ts / evaluationRunsApi.ts covered
```
Raw proof — a `demo` run's SSE body with `EVALUATION_RUN_SSE_HEARTBEAT_MS=50`:
```
      1 event: started
      3 event: progress
      1 event: testCaseComplete
     50 : ping
      1 event: completed      ← last frame
```

## Stacking

Rebased onto main after #479/#468 merged; standalone.


goyamegh:
## Verification proof

**Gates (local, file storage on a private port, `EVALUATION_RUN_SSE_HEARTBEAT_MS=50` on the test server)**
```
npm run build:all                          ✓
npx tsc --noEmit                           ✓
npm run test:unit                          ✓  Test Suites: 371 passed · Tests: 6425 passed, 5 skipped
integration (evaluationRun* suites)        ✓  7 suites / 48 tests, incl. the new evaluationRunsSseHeartbeat test
e2e (new spec + benchmark-page specs)      ✓  benchmark-add-run-header-reset ×3, benchmark-runs-table, evals3-benchmark-runs,
                                              benchmark-runs, benchmark-runs-delete-all-rows, benchmarkruns-associated-evalruns — 40 passed
coverage                                   ✓  lib/runLaunchState.ts 100% · every changed line in BenchmarkRunsPage.tsx /
                                              evaluationRuns.ts / evaluationRunsApi.ts executed by the new unit tests
```

**Red → green** (same tests run against the pre-fix `BenchmarkRunsPage.tsx`, with only the `data-testid` added so the locator resolves):
```
e2e  SSE stays open but silent after `started`      ✘ old: Expected "idle", Received "running"   ← the reported bug (spins forever)
e2e  SSE closes without `completed`                 ✘ old: Expected "running", Received "idle"   ← old code lied the other way
unit THE BUG: dropped stream → Running… → Add Run   ✘ old
unit doc cancelled elsewhere resets header          ✘ old
                                                    ✓ all green on this branch
```

**Raw SSE body** of a `demo` run on the patched server (frame types, `\n\n`-split):
```
      1 event: started
      3 event: progress
      1 event: testCaseComplete
     50 : ping
      1 event: completed        ← last frame
```

## Adversarial review (codex, gpt-5.4) — findings & disposition

|---|---|---|---|
| 1 | high | Happy path sets `launchedRun=null` before `finally` flips `isStreamLive`, so the progress panel could show SSE-derived rows with the blanket "mark completed" | **Rejected** — the `try` tail and `finally` run in one synchronous block after the `await`, so React 18 batches them into a single render; and the panel is not rendered at all once the header is idle (pre-existing: it disappears, the rows table is the truth). |
| 2 | high | Button keyed off a capped (`size:100`) newest-first list + 30 s grace could flip to idle for a live run, and polling could stop | **Rejected (documented)** — the launched run is the newest doc for this benchmark and both lists are newest-first, so it is in the first page for its lifetime; it would need 100+ newer runs on the same benchmark during its life. The grace is the deliberate "no second way to spin forever". Assumption now stated on the input type. |
| 3 | high | `handleStartRun` not re-entrancy-safe (fast double submit → two POSTs, second `started` overwrites the tracked run) | **Applied** — synchronous `launchInFlightRef` guard; unit test posts twice in one tick and asserts exactly one execute. |
| 4 | med | Heartbeat interval keeps ticking (no-op) after client disconnect for the rest of the run | **Applied** — `res.once('close', stop)` and self-clear on the first tick that finds `destroyed/writableEnded`; unit tests assert `getTimerCount() === 0`. |
| 5 | med | Comment frames might be coalesced by buffering layers; check compression/anti-buffering | **Rejected** — no compression middleware in the app; the route already sends `X-Accel-Buffering: no` + `Cache-Control: no-cache`; SSE comment frames are the spec's keep-alive mechanism. And the client fix is correct even if a proxy still drops the stream. |
| 6 | med | No unmount/abort guard around the post-`await` setters | **Rejected (pre-existing, out of scope)** — same shape as before this PR; React 18 no longer warns on setState-after-unmount, and the setters are idempotent. |
| 7 | med | `launchedAt ?? now` default hands a bad caller 30 s of "Running…" | **Applied** — `launchedAt` is required with a runId; id-without-timestamp gets no grace (fails closed to idle). |
| 8 | low | Over-engineering: `isStreamLive`, `now` seam, env parser, `data-run-state` | **Rejected** — each has one concrete job: `isStreamLive` selects the progress source; `now` makes the grace testable; the env var is what the integration test uses to set a 50 ms heartbeat; `data-run-state` is what the e2e asserts on. |
| 9 | low | `isTerminalRunStatus` duplicates the private `isTerminalStatus` in `lib/benchmarkRunsTable.ts` | **Rejected for now** — that file's own comment defers to `isTerminalRunStatus` from `lib/runStats` in the in-flight #486; adding it to `runStats` here would conflict with #486. Collapse onto one once #486 lands. |
| 10 | low | "CLI uses a similar parser" is not a guarantee | **Applied** — verified both CLI consumers are line/frame-based and skip data-less frames; added a `cli/utils/apiClient.createEvaluationRun` contract test with `: ping` frames (incl. split across chunks). |

goyamegh:
**Verification proof — owner-clarification rework (head `4b574f349`)**

Gates (scratch worktree rebased on `origin/main` @ `e730c8e6e`, file storage, isolated server on port 4840):

```
npm run build:all         → exit 0
npx tsc --noEmit          → clean
npm run test:unit         → Test Suites: 384 passed, 384 total / Tests: 5 skipped, 6714 passed, 6719 total
CI=1 playwright (tests/e2e/benchmark-add-run-header-reset.spec.ts + benchmark-runs-table.spec.ts)
                          → 9 passed, 1 flaky (pre-existing runs-table openRunsTab bounce under 2 workers; passes on retry and 6/6 alone)
DCO                       → Signed-off-by on all 5 commits
```

New/updated e2e (`tests/e2e/benchmark-add-run-header-reset.spec.ts`, imports `./fixtures/test-fixtures`):
- THE REGRESSION: with a run in flight the header stays an ENABLED "Add Run" with a "● 1 running" pill; launching a second run → "● 2 running" and two progress blocks, the first untouched
- SSE stream closes without `completed` → the progress block stays while the polled doc is running (not flipped to failed), then leaves once it is terminal (no reload)
- SSE connection stays open but goes silent after `started` (tunnel stops relaying) → block still leaves when the polled doc is terminal
- happy path: SSE delivers `completed` → the progress block goes away; Add Run was never disabled

New unit coverage (`tests/unit/components/evals3/BenchmarkRunsPage.test.ts`, `tests/unit/lib/runLaunchState.test.ts`): button enabled while a run runs; no alert; two launches → two blocks, each leaving on its own terminal doc; pill follows polled docs + filters the table; launching-only disable; `started` timeout release + late `started`; stale `started` never releases a newer launch's window; Edit gate; run numbering; `pruneLaunchedRuns` / `countRunsInFlight` / `progressFromPolledDoc`.

Headless screenshot of the header (enabled "Add Run", `● 2 running` pill) with two live progress blocks was captured on the built branch; available on request.

TackAdam (approved):

PR #504, opened 2026-09-10
fix(runs): stamp benchmarkVersion on CLI-created evaluation runs so they appear on the benchmark page

ashwin-pc:
Evaluation runs created through the unified CLI `POST /api/storage/evaluation-runs` route were linked into the benchmark but never stamped with `benchmarkVersion`. The benchmark page filters `runs[]` by version, so those runs were invisible there while still appearing on the Evaluation Runs list — a run exists, is judged, and completes, but its benchmark shows no trace of it.

### Fix

The route already loads the benchmark for linkage; it now copies `currentVersion` onto both the first-class evaluation run and the embedded benchmark-run projection, before finalization (`server/routes/storage/evaluationRuns.ts`). Runs created via the UI path were unaffected — this closes the gap between the two entry points.

### Observed

Reproduced on a live file-mode instance: two completed CLI runs with `benchmarkVersion: null` absent from `benchmark.runs[]`; after the fix (and re-linking via the API) both list with `benchmarkVersion: 1`. Regression test covers the CLI route stamping both copies.


goyamegh:
@ashwin-pc heads-up: after today's merges (#468 #486 #489 #490 #492 landed on `main`) this PR now conflicts in `server/routes/storage/evaluationRuns.ts`. A rebase onto current `main` is all that's needed — once it's clean we'll stage it for goyamegh to test and merge. Thanks!

PR #509, opened 2026-09-13
feat(runs): retry judgement is always available, with an evaluator + judge-model picker (defaults = last selected); only the latest judgement is kept

goyamegh:
Follow-up to #468.

## Owner requirement (plain words)

> Retry judgement should be a retryable step all the time. We only preserve the last one, but the judgement should allow for evaluator type and prompt evaluator when retrying; defaults will be the last selected ones.

Before this PR, "Retry judgement" in the run kebab was only enabled when a run had **judge-failed** cases, always re-used the run's own evaluator + judge model, and silently overwrote the verdict with no trace that a re-judge had happened.

## What changes

**1. Always retryable.** The kebab item is enabled for ANY terminal evaluation run (completed / failed / cancelled) that has at least one completed test case (i.e. stored agent output to re-judge) — not only judge-failed ones. Running runs stay disabled (existing 409 gate). The label's `(N)` is the number of cases the dialog's *default* scope will re-judge: the judge-failed count when there are any, otherwise all cases. (`lib/runActions.ts`: new `countRejudgeable`, `rejudgeableCount`, `retryJudgementCount`.)

**2. Picker dialog.** Clicking it opens the (existing) `RetryJudgementConfirmDialog`, now with three pickers:
- **Evaluator** — from `GET /api/storage/evaluators` (system + custom)
- **Judge model** — the shared `JudgeModelSelect` ("Use evaluator default" allowed)
- **Scope** — "Only judge-failed cases (N)" (default when N > 0) / "All cases (M)"

Every run surface (runs-list row, run report page, inspector header) now opens this dialog — the runs list and report page used to fire the POST straight from the menu. **Defaults = last selected**: the run's `lastJudgementRetry` when a retry has happened, else the run's own `evaluatorId` / `judgeModelId` (`seedRetryJudgementDefaults`).

**3. Only the latest judgement is kept — truthfully.** Per the owner's choice there is *no history array*. Each re-judged report has its judgement fields overwritten in place (`passFailStatus`, `metrics`, `llmJudgeReasoning`, `llmJudgeResponse`, `matcherResults`, `judgeModelId`, `evaluatorId`, `judgeMode` — the last cleared to `null` when the new judge didn't set it, so a stale trace-judge mode can't outlive its verdict) and is stamped with `judgementRetriedAt` (ISO) + `judgementRetryCount` (incremented). The agent output (trajectory / rawEvents / spans) is never rewritten. Run stats are recomputed through the existing `lib/runStats` path.

**4. Last-selected persistence.** The run doc records `lastJudgementRetry: { evaluatorId, judgeModelId, scope, at }` when a retry is launched (written up front, before any judge call, so it survives an interrupted job). Typed on `EvaluationRun`; new report/run fields are mapped explicitly in `server/constants/indexMappings.ts`.

**5. Run report UI.** The Judge Evaluation tab shows a muted one-liner when `judgementRetriedAt` is set: *"Re-judged 5m ago with Factuality · Claude Sonnet (retry #2)"*.

**6. Docs.** `docs/ARCHITECTURE.md` gets a "Retry judgement" section + the two endpoints in the API table; CHANGELOG `## [Unreleased]` → `### Added`.

## Design notes

- **Why no history:** the owner explicitly chose "we only preserve the last one". Keeping a judgement history would mean a second verdict surface in every consumer (comparison, stats, matchers) for a feature whose purpose is *correcting* a judgement, not auditing it. The stamp (`judgementRetriedAt` / `judgementRetryCount`) plus `lastJudgementRetry` on the run gives the "this was re-judged, with what" signal without a second data model.
- **Extend the existing dialog vs. a new one / a third `RunConfigDialog` mode:** extending `RetryJudgementConfirmDialog` was the least code — it already owned the 202 + status-poll + progress + summary flow. `RunConfigDialog` is about *launching* runs (agent, concurrency, sources) and would have needed a mode that hides most of itself.
- **Scope default is contextual, not remembered:** evaluator + judge model default to the last selection; scope defaults to "only judge-failed" whenever there are judge-failed cases (that is what a retry is almost always for) and "all" otherwise. Remembering `scope=all` would make the next retry after a judge outage re-grade the whole run by default.

## API change

`POST /api/storage/evaluation-runs/:id/retry-judgement` — JSON body (all optional; `?scope=` on the query string still honoured for older callers, body wins):

```json
{ "scope": "errored" | "all", "evaluatorId": "system-factuality", "judgeModelId": "<model id>" | null }
```

- `evaluatorId` must name an existing system (`system-*`) or stored evaluator → **400** `Evaluator not found: <id>` otherwise
- `judgeModelId` must be a non-empty string, or `null` = "use the evaluator default" → **400** otherwise
- `scope` must be `errored` | `all` → **400** otherwise
- absent keys inherit the run's own evaluator / judge model (pre-existing behaviour)
- running run → **409** (unchanged); overrides win over the run's values (`resolveRetryJudgeConfig`: overrides > run > report > `BEDROCK_MODEL_ID`)

The client `retryJudgement(id, request | scope, onProgress)` accepts the new request object (old string form still works).

## Test matrix

| Level | File | Covers |
|---|---|---|
| Unit | `tests/unit/services/evaluation/retryJudgement.test.ts` | `resolveRetryJudgeConfig` precedence (overrides > run > report > env, `null` = default); overridden judge call; in-place overwrite of judgement fields + stamp; agent output untouched; second retry → count 2, one judgement; `lastJudgementRetry` written before any judge call; mirrors run values when no override |
| Unit | `tests/unit/lib/runActions.test.ts` | `countRejudgeable`; enablement matrix — terminal+completed → enabled (incl. cancelled), running → disabled, no completed case → disabled with reason; label count for both scopes |
| Unit | `tests/unit/components/evals3/RetryJudgementConfirmDialog.test.ts` | default seeding (lastJudgementRetry > run > built-in; `null` → default, not the run's pinned value); scope default; N=0 disables the judge-failed radio; submit body (`''` → `null`); progress → summary → `onComplete`; 400 surfaced inline |
| Unit | `tests/unit/server/routes/storage/evaluationRuns.test.ts` | body forwarded as `overrides`; system ids resolve in-process, stored ids checked via storage; unknown / malformed → 400 with no job; legacy `?scope=all` |
| Unit | `tests/unit/components/RunDetailsContent.test.ts`, `RunInspectorPage.test.ts` | "Re-judged … with …" line (absent / present / fallback wording); inspector kebab: clean run enabled with all-cases count, dialog handed both counts, no-completed-case disabled |
| Integration | `tests/integration/server/routes/storage/evaluationRuns.retryJudgementPicker.integration.test.ts` (file storage, `demo` agent, demo judge) | (a) `scope=all` + overrides on a fully-judged run → every report re-judged, `evaluatorId`/`judgeModelId` = overrides, count 1, `judgementRetriedAt` set, `lastJudgementRetry` recorded, stats recomputed, agent output untouched; (b) second retry → count 2, one judgement; (c) unknown evaluator → 400, nothing re-judged; (d) running → 409; (e) legacy `?scope=all` + inherited run values |
| E2E | `tests/e2e/retry-judgement-picker.spec.ts` (real server, no judge mocks) | clean run → item ENABLED → dialog preselects the run's evaluator/model, "All cases" checked → pick another evaluator → submit → progress → summary → report Judge tab shows "Re-judged … with Factuality · …" → reopen → defaults are the last selection; runs-list row opens the same picker (no POST until confirm), running row disabled |
| E2E | `tests/e2e/retry-judgement.spec.ts`, `tests/e2e/run-actions-menu.spec.ts` | updated for the new gating (clean run enabled; report-page kebab goes through the dialog) |

Gates run locally against a file-storage server (CI=1 mode for e2e): `npm run build:all` ✅ · `npx tsc --noEmit` ✅ · `npm run test:unit` ✅ (3 pre-existing, unrelated suite failures on `origin/main` too: `tests/unit/cli/index.test.ts`, `tests/unit/lib/testCases/loader.*.test.ts`) · new integration file 5/5 + existing retry-judgement integration 8/8 ✅ · e2e 18/18 across the three specs ✅ · new-line coverage: `retryJudgement.ts` 97% lines, `runActions.ts` 100%, dialog 88%, route 92%.


goyamegh:
## Verification proof (local, file-storage server on port 4800, `demo` agent + demo judge; e2e in `CI=1` production-build mode)

```
$ npm run build:all                       → ✓ built · server/cli/lib bundles Done
$ npx tsc --noEmit                        → exit 0
$ npm run test:unit                       → Test Suites: 373 passed, 373 total · Tests: 6532 passed, 5 skipped
$ AH_PORT=4800 npx jest tests/integration/server/routes/storage/evaluationRuns.retryJudgement
    evaluationRuns.retryJudgement.integration.test.ts        8 passed   (existing)
    evaluationRuns.retryJudgementPicker.integration.test.ts  5 passed   (new)
      ✓ (a) scope=all with overrides re-judges EVERY fully-judged case with the picked evaluator/model, stamps the retry,
            records lastJudgementRetry, recomputes stats, keeps the agent output; (b) second retry → count 2, one judgement
      ✓ (c) 400s an unknown evaluatorId and re-judges nothing
      ✓ 400s a malformed judgeModelId / scope
      ✓ (d) 409s while the run is still running, even with a valid picker body
      ✓ (e) absent body keys inherit the run's evaluator / judge model; legacy ?scope=all still honoured
$ CI=1 PLAYWRIGHT_SKIP_WEBSERVER=1 AH_PORT=4800 npx playwright test \
    tests/e2e/retry-judgement.spec.ts tests/e2e/retry-judgement-picker.spec.ts tests/e2e/run-actions-menu.spec.ts
    18 passed (27.2s)
$ npm audit --audit-level=high            → same lockfile as origin/main (no dependency changes in this PR)
```

New-line coverage (jest `--coverage` over the touched files): `services/evaluation/retryJudgement.ts` 97% lines · `lib/runActions.ts` 100% · `RetryJudgementConfirmDialog.tsx` 88% · `server/routes/storage/evaluationRuns.ts` 92%.

### End-to-end transcript (headless Chrome against the real server, seeded 2-case fully-judged run)

```
kebab → "Retry judgement (2)" ENABLED on a run with 0 judge-failed cases
dialog → Evaluator: RCA Default (System) · Judge model: Demo Model · ○ Only judge-failed cases (0) [disabled] · ● All cases (2)
pick Evaluator = Factuality → Retry judgement → "Retrying judgement... 0/2" → summary: Retried 2 · 2 succeeded → Done
report Judge tab → "Re-judged Just now with Factuality · Demo Model"
GET /api/storage/evaluation-runs/<id>
  lastJudgementRetry: {"scope":"all","evaluatorId":"system-factuality","judgeModelId":"demo-model","at":"2026-09-13T01:29:10.318Z"}
  stats: {"passed":2,"failed":0,"errored":0,"pending":0,"total":2}          (was passed 1 / failed 1)
  evaluatorId: "system-rca-default"                                        (run's launch config untouched)
GET /api/storage/runs/<reportB>
  {"evaluatorId":"system-factuality","judgeModelId":"demo-model","passFailStatus":"passed","judgementRetryCount":1,"judgementRetriedAt":"2026-09-13T01:29:10.324Z"}
  trajectory / rawEvents byte-identical to the seed
reopen dialog → Evaluator preselected: Factuality (System)  ← defaults = last selected
```

## codex_review (gpt-5.4, adversarial) — findings and disposition

|---|---|---|---|
| 1 | high | Dialog seeds `''` → sends `judgeModelId: null` → `serverDefault`, skipping `report.judgeModelId` on runs with no run-level judge model — behaviour change | **Rejected.** The runner stamps `report.judgeModelId = run.judgeModelId` (services/evaluationRunner.ts), so the two are identical in practice; and `null` resolves to exactly what an unpinned original run used (`BEDROCK_MODEL_ID`, then `/api/judge` applies the evaluator's `inferenceConfig`). `null` is needed so a user can explicitly choose "evaluator default" on a run that *did* pin a model. |
| 2 | high | Dialog always sends an evaluator; invents `system-rca-default` when the run has none | **Rejected (equivalent).** `/api/judge` with no `evaluatorId` uses `getDefaultEvaluator()` = `system-rca-default`. Making it explicit is the point: the report now records the evaluator that actually judged it. |
| 3 | high | Success patch only spreads `evaluatorId` when truthy → a stale `report.evaluatorId` could survive a retry that used the default | **Applied** (cb43b5c). `resolveRetryJudgeConfig` now always resolves a concrete evaluator id (`getDefaultEvaluator().id` when neither override nor run pins one) and the report is always stamped with it. Unit test added. |
| 4 | med | `lastJudgementRetry` is written even when the retry fails; failed reports aren't stamped | **Rejected by design.** `lastJudgementRetry` = "what the last retry was *launched* with" (documented on the type) — it seeds the next dialog, and re-trying a failed retry with the same selection is the expected default. The report's `judgementRetriedAt` means "this verdict is a re-judgement"; a failed retry produced no verdict, so stamping it would make the "Re-judged … with" line lie. |
| 5 | med | UI gates on `results[*].status === 'completed'` but the server also requires a stored trajectory → item enabled, job total 0 | **Rejected (known, pre-existing).** #468 had the same gap for the judge-failed count; the run doc can't see report trajectories. The dialog already says "cases whose agent execution never completed are skipped — the retried count may be lower", and `lib/runActions.ts` documents it. |
| 6 | med | `(N)` label = default scope's count, dialog can switch to all → misleading | **Rejected (owner spec).** The requirement asked for `(N)` = cases that will be re-judged for the chosen scope; the label shows the default scope's count and the dialog shows the live count for each scope before submit. |
| 7 | med | Re-seed effect keyed on `lastJudgementRetry.at` can wipe an in-progress selection on parent refetch | **Applied** (cb43b5c). Deps are now `[open, runId]` only; regression test added (fresh run object with a new `lastJudgementRetry` does not reset the selection). |
| 8 | med | `resolveRetryJudgeConfig` falls back to `report.modelId` (agent model) for the judge | **Rejected (pre-existing, unchanged).** That last-resort fallback existed before this PR and mirrors the runner's own resolution; it only applies when `BEDROCK_MODEL_ID` is unset. |
| 9 | med | Run-level `evaluatorId`/`judgeModelId` not updated → two sources of truth | **Rejected by design.** The run's fields describe its launch config (and feed `imageDigest`/re-run seeding); a `scope=errored` retry judges only *some* reports, so a run-level field would be ambiguous. Per-report `evaluatorId`/`judgeModelId` are the truth for "what judged this report", plus `lastJudgementRetry` for the selection. Called out in the PR description. |
| 10 | low | Over-engineering: `string | object` client overload, `?scope=` fallback, `null` semantics, explicit OS mappings, pre-job write | **Partially applied** (cb43b5c): dropped the string overload and the dead `retryJudgementEvaluationRun` alias. Kept: `?scope=` (older callers / existing integration tests), `null` (needed, see #1), explicit mappings (`at` as `date`, ids as `keyword` — cheap and consistent with the index), pre-job write (so the selection survives an interrupted job). |

TackAdam (approved):

PR #510, opened 2026-09-13
feat(evals3): show each run's concurrency on the benchmark Runs tab, the runs list and the run header

goyamegh:
## Problem

Owner feedback on #486 (2026-09-13):

> We should show concurrency on the runs page of each run, for both benchmark details page and Evaluation runs page.

Runs already persist `concurrency?: number` ("Parallel test case execution limit, 1 = sequential") on both `BenchmarkRun` and `EvaluationRun`, but it was never surfaced anywhere in the UI.

## Change

- Added a compact **Conc.** column (header tooltip "Concurrency — parallel test cases") right after **Size** on the benchmark Runs tab (`components/evals3/BenchmarkRunsTable.tsx`), sortable via the existing sort-field plumbing in `lib/benchmarkRunsTable.ts` — numeric column, defaults to descending, legacy runs with no `concurrency` field always sink to the bottom regardless of direction.
- Added the same value/em-dash rendering to the Evaluation Runs list (`components/evals3/EvalRunsPage.tsx`), between Model and Judge.
- Added a tiny "· conc N" chip to the run inspector header metadata (`components/evals3/RunInspectorPage.tsx`), omitted entirely for legacy runs.
- Shows the literal value (including `1` for sequential runs) and an em dash (`—`) only when the field is genuinely absent (pre-existing runs persisted before this field existed).
- **Bug found while e2e-testing this**: `services/storage/asyncBenchmarkStorage.ts` has three separate hand-written allow-list mappers (`toBenchmarkRun`, `toStorageFormat`, `addRun`'s inline mapper) that convert between the app's `BenchmarkRun` shape and the storage layer's shape. None of the three listed `concurrency`, so a benchmark-embedded run's configured concurrency was silently dropped on every read/write round trip — the column always showed "—" even for runs that WERE configured with a concurrency. Standalone `EvaluationRun` docs (fetched via `listEvaluationRuns`/`getEvaluationRun`, which do a plain `response.json()` with no mapper) were unaffected. Fixed all three mappers.

## Test matrix

- **Unit**: `tests/unit/lib/benchmarkRunsTable.test.ts` (row carries `concurrency` through as-is; numeric sort with legacy rows sinking to the bottom in either direction), `tests/unit/components/evals3/BenchmarkRunsTable.test.ts` (column position right after Size, value/em-dash rendering, click-to-sort), `tests/unit/components/evals3/EvalRunsPage.test.ts`, `tests/unit/components/evals3/RunInspectorPage.test.ts`, `tests/unit/services/storage/asyncBenchmarkStorage.test.ts` (regression coverage for all three mappers dropping `concurrency`).
- **E2E**: `tests/e2e/runs-concurrency-column.spec.ts` (seeds sequential/parallel/legacy benchmark-embedded runs; asserts both pages show the right values, the benchmark-tab column sorts ascending/descending with legacy rows sinking, and the inspector header chip is present/absent correctly). Updated the pre-existing column-header assertion in `tests/e2e/benchmark-runs-table.spec.ts` (a neighbouring spec on the same table) for the new column.

Gates run locally: `npm run build:all`, `npm run test:unit` (6558 tests, 375 suites, all green), `npx tsc --noEmit`, targeted e2e (both the new spec and the neighbouring `benchmark-runs-table.spec.ts` green under `CI=1`, file storage, 2 workers), DCO signoff on both commits.


goyamegh:
## codex_review (gpt-5.4) — findings applied / rejected

Ran two passes (the first pass's diff paste accidentally omitted 3 test files, causing 2 false-positive "missing coverage" findings I caught and dismissed after re-verifying the actual PR already had that coverage).

**Applied:**
- **[MED]** `services/storage/asyncBenchmarkStorage.ts`'s `toBenchmarkRun` forwarded `stored.concurrency` verbatim, but every render site only checks `=== undefined` — a schemaless stored `null` would render blank instead of the intended `—`/omitted-chip fallback. Normalized with `stored.concurrency ?? undefined` on read, plus a regression test.
- **[LOW/MED]** The new Conc. column hand-rolled a duplicate `<th>` instead of reusing the existing `SortHeader` component used by every other sortable column — drift risk if `SortHeader`'s a11y/sort-icon behavior changes later. `SortHeader` now takes an optional `title` prop; the Conc. column reuses it.
- Also proactively fixed (found while re-verifying finding #2 from an earlier informal pass, before the final diff): `services/storage/opensearchClient.ts`'s `StorageBenchmarkRunConfig` didn't declare `concurrency` at all (read via an `as any` cast); added the type. Added a regression test locking in that the SEPARATE, pre-existing `server/routes/storage/evaluationRuns.ts` code path (linking a completed run into `benchmark.runs[]`) already correctly forwards `concurrency` — no source change needed there, just coverage.

**Rejected (with reason):**
- "The `StorageBenchmarkRunConfig.concurrency` type addition is only half a contract cleanup — sibling fields (`benchmarkVersion`, `testCaseSnapshots`) still use `as any`." — True, but fixing every under-typed sibling field is a separate, larger typing-debt refactor out of scope for a concurrency-display PR; noted for a future pass.
- "No direct test for `EvalRunsPage`/`RunInspectorPage` rendering, no `null` runtime test" — this was based on an incomplete diff I pasted into the review tool (I omitted `EvalRunsPage.test.ts` and `RunInspectorPage.test.ts`'s diffs by mistake). Both already had this exact coverage before the review; the `null`-runtime gap it also flagged WAS real and is fixed above.
- Speculative "there may be a separate allow-list mapper for standalone `EvaluationRun` docs" — verified directly: `listEvaluationRuns`/`getEvaluationRun` do a plain `response.json()` with no mapper, so standalone docs were never affected by this bug class.

## Verification

- `npm run build:all`, `npx tsc --noEmit`: clean.
- `npm run test:unit`: **375 suites / 6560 tests, all green** (final run after all review-driven fixes, post-rebase onto current `main`).
- E2E (`CI=1`, file storage, ports in the 4820-4829 range, `BENCHMARK_RUN_RECOVERY_DISABLED=1 EVALUATION_RUN_RECOVERY_DISABLED=1`): new `runs-concurrency-column.spec.ts` (3/3) + neighbouring `benchmark-runs-table.spec.ts` (6/6) — 9/9 green under `--workers=2`.
- Visual proof (bottom-right HUD... n/a for this PR; screenshots of the actual Conc. column and chip):
  - Benchmark Runs tab: sequential run shows `1`, parallel shows `5`, legacy run shows `—`.
  - Evaluation Runs list: same three values in the Conc. column between Model and Judge.
  - Run inspector header: `· conc 5` chip on the configured run.

(Screenshots captured against a local build seeded with exactly this scenario; not attached here per the no-committed-binaries convention — happy to share in chat if useful.)

TackAdam (approved):

PR #511, opened 2026-09-13
feat(debug): per-page latency HUD (render / data-ready / API calls) when debug or dev mode is enabled

goyamegh:
## Problem

Owner ask: "publish latencies of each page when debugging or dev mode is enabled."

Debug mode already exists (`lib/debug.ts`: server-side state + a browser `localStorage` cache kept in sync with the Settings page's toggle), but there was no way to see how long a page actually took to render or load its data.

## Change

- **`lib/pageLatency.ts`** (new): per-route latency instrumentation, active ONLY when debug mode is on (`isDebugEnabled()`) or this is a dev build (`import.meta.env.DEV`, isolated in a tiny `lib/viteEnv.ts` module so Jest's CommonJS transform never has to parse `import.meta` — same convention as the existing `lib/config.ts` mock).
  - `startNavigation(pathname)`: called on every route change. Maps the pathname to a stable route key (`routeKeyFromPath`) so dynamic segments (a benchmark id, a run id) don't fragment history into one entry per id, stamps a start time, and schedules a render timestamp two animation frames later.
  - `markPageReady(routeKey)`: a one-line call each instrumented page makes once its primary data load resolves. Guarded against a slow page finalizing a LATER navigation's record after the user has already moved on (`routeKey` must match the currently active navigation).
  - While a navigation is active, `window.fetch` is wrapped to count `/api/*` calls (count + total ms) into that navigation's record; unwrapped again the moment debug/dev mode is off.
  - Finalized records go into a capped (10) history and are also logged via the existing `debug()` logger.
- **`components/DebugLatencyHud.tsx`** (new): a small, unobtrusive bottom-right HUD (`data-testid="debug-latency-hud"`), e.g. `benchmark-runs · render 120 ms · ready 840 ms · 6 api / 610 ms`. Click/hover reveals the last 10 navigations. Polls the debug-mode flag every second (same pattern as the existing `PerformanceOverlay`) so a toggle from the Settings page or another tab is picked up without a hard refresh.
- **`components/Layout.tsx`**: one new `useEffect` calling `startNavigation(location.pathname)` on route change, and renders `<DebugLatencyHud />`.
- Six pages call the one-line `markPageReady(routeKey)` once their primary load resolves: `BenchmarksPage`, `BenchmarkRunsPage`, `EvalRunsPage`, `RunInspectorPage`, `ComparisonPage`, `AgentTracesPage`.
- **Zero behaviour change when inactive**: every exported function short-circuits to a no-op — no timers, no `fetch` wrapping, nothing rendered.
- Docs: 5-line blurb added to `docs/CONFIGURATION.md`'s existing "Debug Logging" section.

## Test matrix

- **Unit**: `tests/unit/lib/pageLatency.test.ts` (route-key mapping for all 6 pages' routes, off-mode no-op including `fetch` left completely untouched, render/ready timing via fake timers, the stale-routeKey guard, `/api/*`-only fetch aggregation, 10-entry history cap, subscribe/unsubscribe), `tests/unit/components/DebugLatencyHud.test.ts` (hidden when inactive or no record yet, summary line rendering incl. em-dash-before-measured, hover history, debug-toggle poll).
- **E2E**: `tests/e2e/debug-latency-hud.spec.ts` — enables debug mode the same way the Settings page does (`POST /api/debug` for the server-side truth + a `page.addInitScript` pre-seed of the same `localStorage.agenteval_debug` key the Settings page writes, avoiding a real race against `App.tsx`'s per-route-change server sync on the very first page load): HUD absent when debug is off; shows real route + non-zero timing once enabled, updates across a client-side navigation, and the previous page's entry shows up in history on hover; disappears again once debug is turned back off. Ran alongside the neighbouring `tests/e2e/evals3-benchmarks.spec.ts` (21/21 green).

Gates run locally: `npm run build:all`, `npm run test:unit` (6569 tests, 377 suites, all green), `npx tsc --noEmit`, targeted e2e (`CI=1`, file storage, production build) green, DCO signoff on both commits.

**Note on debugging this locally**: this PR's e2e run surfaced (and worked around, not "fixed", since it's correct/intended behavior) an environment quirk — if `NODE_ENV=development` is set in the shell that runs `vite build`, the resulting bundle's `import.meta.env.DEV` compiles to `true` even for a "production" build, which would make the HUD show unconditionally. A real CI runner with `NODE_ENV` unset builds correctly with `DEV=false`.


goyamegh:
## codex_review (gpt-5.4) — findings applied / rejected

**Applied (3 real correctness bugs in the `/api/*` fetch-aggregation window, `lib/pageLatency.ts`):**
1. **[HIGH]** A slow `/api/*` call started under page A but resolving *after* the user had already navigated to page B was silently counted against B — the wrapped `fetch`'s `finally` re-read the mutable module-level `current` instead of the record that was active when the call started. Fixed by capturing the record at call-start time (`recordAtCallStart`) and mutating that, never the live `current`.
2. **[HIGH]** Counting never stopped once `markPageReady()` had already finalized a record — a background poll/refresh firing after "ready" kept inflating a number the HUD had already reported as final. Fixed by gating the increment on `recordAtCallStart.readyMs === null`.
3. **[MED]** Turning debug mode off mid-navigation left `window.fetch` wrapped and still mutating the (now-hidden) record until the *next* route change actually got a chance to restore `fetch`. Fixed by also gating the increment on `isPageLatencyActive()` at resolution time — the wrapper stops mutating state instantly, even before the physical unwrap happens on the next navigation.
4. **[MED]** Related fix bundled in: `markPageReady()` now finalizes `renderMs` synchronously if it hasn't fired yet, so a very fast page (ready before the 2-animation-frame render measurement lands) doesn't get a permanent `—` baked into its history entry.
5. **[MED]** Docs wording softened: "shows the current page's load timing" overstated the feature — only the six pages that call `markPageReady` ever populate "ready"/API numbers; other routes show render time only. `docs/CONFIGURATION.md` now says so explicitly.

Added 4 new regression tests in `tests/unit/lib/pageLatency.test.ts` locking in each of the above (call-start attribution, post-ready no-op, mid-flight-toggle no-op, eager renderMs finalization) — 19/19 passing in that file.

**Rejected (with reason):**
- "The route-key regex for `/compare` only allows one extra path segment; a hypothetical deeper compare URL would silently fall back to the raw pathname while `ComparisonPage` still calls `markPageReady('comparison')`, causing a no-op." — No such deeper route exists today (`/compare` and `/compare/:benchmarkId` are the only two). Speculative; not fixing ahead of an actual route that needs it.
- "The `Listener`/`subscribe` pub-sub, the route-key table, and the two-`requestAnimationFrame` render-timing technique might be over-engineered / could duplicate `lib/performance.ts`'s `PerformanceOverlay` pattern." — `PerformanceOverlay` measures ad-hoc named spans devs wrap manually (`performance.mark`-style); this feature measures navigation-to-render/ready/API-window automatically per route with zero call-site instrumentation beyond the one-line `markPageReady`. Different problem, not a reinventable overlap — kept separate.
- "The `lib/viteEnv.ts` isolation + Jest-mock-wholesale approach for `import.meta` — is there a cleaner alternative?" — This mirrors the codebase's own existing, established convention for the exact same problem (`lib/config.ts`, `lib/packagePaths.ts` in `jest.config.cjs`'s `moduleNameMapper`); introducing a different mechanism (e.g. a Babel transform) for one new file would be the actual inconsistency.
- An earlier review pass (before the diff I gave codex included the real fetch-aggregation logic) raised a false-positive about missing `EvalRunsPage`/`RunInspectorPage`-equivalent coverage — not applicable here (that finding was from the OTHER PR in this pair, #510; verified it doesn't apply to this diff).

## Verification

- `npm run build:all`, `npx tsc --noEmit`: clean.
- `npm run test:unit`: **380 suites / 6623 tests, all green** (final run after all review-driven fixes, post-rebase onto current `main`).
- E2E (`CI=1`, file storage, production build with `NODE_ENV` explicitly set — a sandbox quirk where an inherited `NODE_ENV=development` env var makes `vite build` compile `import.meta.env.DEV` to `true` even for a "production" build was found and worked around for local testing; a real CI runner with `NODE_ENV` unset is unaffected): `tests/e2e/debug-latency-hud.spec.ts` (3/3) + neighbouring `tests/e2e/evals3-benchmarks.spec.ts` (21/21) — 21/21 green together under `--workers=1`.
- Leak scan (internal agent/project names) on the full diff: clean, no matches.
- Visual proof: HUD screenshot showing `benchmark-runs · render … · ready … · N api / … ms` in the bottom-right corner, captured against a local build with debug mode enabled via the same `POST /api/debug` + `localStorage.agenteval_debug` mechanism the Settings page uses (not attached here per the no-committed-binaries convention; happy to share in chat).

TackAdam (approved):

PR #528, opened 2026-09-22
fix(traces): semconv-1.37 provider key compliance, normalised span kind, self-time distribution, leaf-only token usage

goyamegh:
### Description

Four small, generic correctness fixes in trace rendering / trace-derived metrics for spans produced by current OTel GenAI instrumentation (semconv ≥ 1.37, and agent frameworks that emit wrapper spans and stamp roll-up usage on them). All are read-side fixes; no data is rewritten.

**1. Compliance false positive on `gen_ai.system`** — `services/traces/spanCategorization.ts`
semconv 1.37 deprecated `gen_ai.system` in favour of `gen_ai.provider.name` (`@opentelemetry/semantic-conventions@1.38` marks `ATTR_GEN_AI_SYSTEM` `@deprecated`), but `EXPECTED_ATTRIBUTES.LLM` required `gen_ai.system`, so every LLM span carrying `gen_ai.provider.name` + `gen_ai.request.model` was flagged "1 missing" (span details panel + Agent-map node triangle). The expectations model now supports any-of groups (`string | string[]`, preferred key first); the missing text names the preferred key with the deprecated alias: `gen_ai.provider.name (or deprecated gen_ai.system)`. A shared `readGenAiProvider()` applies the alias everywhere the provider was read (LLM display names, trace-list service-name fallback, `SpanInputOutput` category/model derivation). AGENTS.md / CLAUDE.md "Required Attributes" now say `gen_ai.provider.name` (formerly `gen_ai.system`).

**2. `/api/traces` spans had no usable top-level `kind`** — `lib/spanKind.ts`, `server/services/tracesService.ts`, `server/services/otlpTransform.ts`
Spans read from the OpenSearch pipeline exposed the kind only as the raw `attributes.spanKind` (`SPAN_KIND_SERVER`), and spans ingested through the OTLP receiver dropped it entirely. Every reader now emits a normalised `kind` (`INTERNAL|SERVER|CLIENT|PRODUCER|CONSUMER`) from the document's `kind` field, falling back to `attributes['span.kind']` / `attributes.spanKind`, accepting the OTLP numeric enum (0–5; 0 = unspecified → no kind), the proto enum name (`SPAN_KIND_SERVER`) and bare/mixed-case names (`Server`). `Span.kind` is typed on the shared `Span`; `attributes.spanKind` carries the same canonical name on every ingest path (raw value only when it could not be normalised); the span-details "Kind" row prefers `kind`.

**3. Time Distribution double-counted nested time** — `services/traces/traceStats.ts` + the three Time Distribution panels
`calculateCategoryStats` summed each span's *inclusive* duration, so wrapper spans (an agent loop span containing LLM + tool children) were credited with their children's time again. On a real 45-span agent trace the bar read OTHER 50% / AGENT 25% / LLM 24% where the LLM calls actually held ~96% of the time. Shares are now **self time**: span duration minus the interval *union* of its direct children (clamped ≥ 0; children matched via `parentSpanId` within the same trace; overlapping siblings handled by union, not naive sum; orphans keep their full self time; parent and children are measured on one length basis — `duration`, placed at `startTime`). `CategoryStats` gains `selfDuration` and keeps the inclusive `totalDuration`; every panel is labelled "(self time)" with the inclusive figure in the tooltip ("… incl. children"). `spanPreprocessing` now delegates to the same function so all trace views agree. The Info view's headline "Duration" is the trace wall-clock instead of a sum of inclusive span durations.

**4. Token metrics double-counted aggregate usage** — `lib/usageAggregates.ts`, `server/services/metricsService.ts`, `services/traces/traceSummary.ts`
Every span carrying `gen_ai.usage.*` (or the `input_tokens`/`output_tokens` fallbacks) was summed, so frameworks that stamp usage on an aggregate parent (an `invoke_agent` span whose usage equals the sum of its `chat` children) were counted twice — on a live run Tokens/Cost read exactly 2× and LLM Calls was off by one. New **leaf-usage invariant**: each span contributes only the usage its descendants (same trace, via the `parentSpanId` chain) do not already account for — `counted = max(0, own − Σ descendants)`, input and output independently. An exact roll-up contributes 0 (and is not counted as an LLM call); a parent carrying *more* than its children still contributes the remainder, so real usage is never dropped (a parent with usage and no usage-carrying descendants is counted in full). Applied to `computeMetricsFromSpans` (single-run + batch — the batch `_source` projection now carries `spanId`/`parentSpanId`, otherwise the batch path alone would keep double-counting), the demo-span path, and the Traces-tab summary strip. `MetricsResult`/`TraceMetrics` gain `usageAggregatesSkipped` (number of roll-up spans skipped) for the UI/debugging. Vendor totals such as `aws.genai.token_count_total` / `gen_ai.usage.total_tokens` are never added on top of input + output (a total-only span contributes 0).

### Review
Adversarial `codex_review` pass applied: subtractive attribution instead of a boolean skip (so a parent with more usage than its children is never dropped), a single length basis for self-time maths, and a canonical `attributes.spanKind` on every ingest path. Rejected (with reasons in the review thread): treating the deprecated-but-accepted `gen_ai.system` as non-compliant (aliases are accepted by the semconv deprecation itself; the missing text still steers to the preferred key), keeping `missingAttributes` as bare keys (the UI badge is the only consumer and now tells users what would satisfy the check), and normalising the Time Distribution against wall-clock (overlapping/concurrent spans make wall-clock the wrong denominator; the bar is explicitly labelled self time and the wall-clock is shown separately).

### Issues Resolved
n/a (found while validating traces from a current-semconv agent framework)

### Tests
- **Unit:** `tests/unit/lib/spanKind.test.ts`, `tests/unit/lib/usageAggregates.test.ts`, `tests/unit/services/traces/traceStats.test.ts` (nested chain, overlapping siblings, orphan child, zero-duration, cross-trace, percentages sum to 100), `spanCategorization.test.ts`, `tracesService.test.ts`, `otlpTransform.test.ts`, `metricsService.test.ts` (parent+children ⇒ once; children-only; parent-only; parent > children ⇒ remainder counted; mixed tree; vendor-total-only; flat @-encoded schema; batch path), `traceSummary.test.ts`, `traceGrouping.test.ts`; rendered-output tests `otelCompliance.render.test.ts`, `timeDistribution.render.test.ts`, `spanInputOutput.provider.test.ts` (real exports, not copies).
- **Integration:** `tests/integration/server/traces-span-kind.integration.test.ts` (OTLP ingest with numeric + string kind → file store → `/api/traces`; OpenSearch reader path with string / numeric / attribute-only kind), `metricsTraceIdAndVendorTokens.integration.test.ts` (aggregate-usage trace through the batch path with `_source` projection; single-run agrees).
- **E2E:** `tests/e2e/traces-semconv-compat.spec.ts` (Agent map: no compliance warning for a `gen_ai.provider.name` span, warning naming preferred key + alias when neither is present; Info view: "(self time)" label, self-time shares, inclusive figure in tooltip).

### Check List
- [x] New functionality has been documented (CHANGELOG, AGENTS.md).


goyamegh:
### Verification proof (head `30b206f6f`)

**Gates (local, this branch)**
```
npm run build:all                          → exit 0
npm run test:unit                          → Test Suites: 387 passed, 387 total
                                              Tests: 5 skipped, 6745 passed, 6750 total
targeted unit + integration (13 suites)    → Tests: 263 passed, 263 total
playwright: traces-semconv-compat.spec.ts + traces-plain-raw / traces-error-span-red /
            traces-eval-capsule / traces.spec.ts  → 29 passed (24.5s)
npm audit --audit-level=high               → exit 0 (no high/critical)
DCO                                        → 7/7 commits Signed-off-by
```

**(2) `/api/traces` now carries a normalised `kind`** — same 45-span agent trace (an HTTP root → `invoke_agent` → loop-cycle → `chat` / `execute_tool` → search spans) read from a Data Prepper `otel-v1-apm-span-*` index, before vs after (span-kind histogram over the response):
```
before (main):      kind: —  (absent on every span)        attributes.spanKind: SPAN_KIND_INTERNAL ×33, SPAN_KIND_CLIENT ×11, SPAN_KIND_SERVER ×1
after (this PR):    kind: INTERNAL ×33, CLIENT ×11, SERVER ×1   attributes.spanKind: INTERNAL ×33, CLIENT ×11, SERVER ×1
```
OTLP receiver path (integration test): `kind: 2` → `SERVER`, `kind: "SPAN_KIND_CLIENT"` → `CLIENT`, `kind: 1` → `INTERNAL`, `kind: 0` → no `kind`.

**(4) `/api/metrics/:runId` on the same run** — the framework stamps `gen_ai.usage.*` on the `invoke_agent` span (= sum of its 10 `chat` children):
```
before (main):   inputTokens 441458  outputTokens 8662  totalTokens 450120  costUsd 1.454304  llmCalls 11
after (this PR): inputTokens 220729  outputTokens 4331  totalTokens 225060  costUsd 0.727152  llmCalls 10  usageAggregatesSkipped 1
```
225060 equals the `invoke_agent` span's own `gen_ai.usage.total_tokens` and the sum of the 10 `chat` spans — i.e. the old number was exactly 2×.

**(3) Time Distribution on the same trace (fullscreen Info view)**
```
before:  OTHER 50% (1m 18.0s) · AGENT 25% (38.36s) · LLM 24% (37.45s) · TOOL 0.76% (1.18s) · ERROR ~0%   [inclusive sums; wall-clock is 38.56s]
after:   LLM 96% (37.45s) · OTHER 3.5% (1.36s) · TOOL 0.17% (68ms) · ERROR 0.01% (6ms) · AGENT 0.01% (3ms)   labelled "(self time)", header = 38.56s wall-clock
```

**(1) Compliance** — Agent-map nodes on that trace: before, all 10 `chat` spans (stamped `gen_ai.provider.name` + `gen_ai.request.model`, no `gen_ai.system`) showed the amber "missing OTel attributes" triangle; after, 0 warnings. A span with neither key now reads `gen_ai.provider.name (or deprecated gen_ai.system)` (e2e asserts both states on a synthetic trace).

PR #532, opened 2026-09-22
fix(traces): precise-first correlation — exact traceId/runId/sessionId before the service-name window (one root span per run)

goyamegh:
### Description

The run report's **Traces tab showed several root spans — other runs' whole trees — for a single agent invocation** whenever the benchmark had run with concurrency > 1.

Reproduced read-only against a shared observability cluster for one report:

| `POST /api/traces` body | spans | traces | roots |
|---|---|---|---|
| `{ traceId }` (Strategy A) | 25 | 1 | 1 |
| `{ runIds: [runId] }` (Strategy B) | 19 | 1 | 1 |
| **what the tab actually sent:** `{ runIds, agents: [service-name + window] }` | **72** | **3** | **3** — roots carried three different `gen_ai.conversation.id` values |

Two causes:
1. `RunDetailsContent.tsx` never passed `report.traceId` / `report.sessionId` — only `runId` + the Strategy-C window hint.
2. The server OR'd every correlation clause into one `bool.should`, so neighbouring cases of the same agent inside the wall-clock window were unioned with the exact match.

### Fix — precise-first correlation (server-side, so every caller benefits)

New `server/services/traceCorrelation.ts` (backend-agnostic; `routes/traces.ts` routes both the OpenSearch and the file observability backend through it):

1. **Exact clauses first** — `traceId` ∪ `runIds` ∪ `sessionId` are queried on their own. If they return any of the **agent's** spans, that is the answer; the window is never consulted. Agent Health's own eval/judge spans don't count as a hit (the eval `test_case` span is on the requested trace and carries the run id, so for a Strategy-C-only agent the exact query returns exactly that span — it must not suppress the fallback).
2. **Window only otherwise** — the Strategy-C `agents[]` clause runs only then. Strategy C stays **always-on as the fallback** (an empty Traces tab by default was judged worse than noise — never empty when the window has data). Any eval spans the exact query found are kept in front of the window result.
3. **Post-filter by run identity, per trace** — a window trace carrying Agent Health's own `agent_health.run.id` for a *different* run is dropped whole (root carries the id; its HTTP/DB children follow it). Only our own attribute is negative evidence: `gen_ai.conversation.id` / `session.id` are OTEL-standard ids a third-party agent may fill with its own thread/session id, so a mismatch there proves nothing and those traces are kept — as are traces with no identity at all (the population Strategy C exists for). `traceId` equality is not required (fallback agents don't propagate W3C context).
4. **Response says what happened** — `correlation: { strategy: 'traceId'|'runIds'|'sessionId'|'window', windowFiltered: n }`, all existing fields (`spans, total, nextCursor, hasMore, backend, warning…`) unchanged. Pagination cursors are phase-tagged so page 2 continues the same query.

Client:
- `RunDetailsContent` passes `traceId` + `sessionId` alongside `runId` + the window hint and captions the Traces header: **"Matched by trace id"** / "Matched by run id" / "Matched by session id" / **"Matched by service-name window — N spans from other runs filtered"**.
- `fetchSpansForRun` (code SDK / runners / retry-judgement) forwards the exact ids too.
- `FileObservabilityModule`: `sessionId` is a unioned correlation clause (as the OpenSearch query already treated it), no longer a must-filter — a stale report session id can't zero out a traceId/runId match.
- `isEvalOrJudgeSpan` moved from `metricsService` to `tracesService` so both readers share it.

Docs: `AGENTS.md` → "Trace correlation conventions" and `docs/CONNECTORS.md` describe the precise-first order and the post-filter.

### Tests

- **Unit** — `tests/unit/server/services/traceCorrelation.test.ts` (direct-first ordering; eval-only exact result still falls back; per-trace filter on `agent_health.run.id` only; third-party `gen_ai.conversation.id`/`session.id` kept; phase cursors), `routes/traces.test.ts` (two-phase through the route), `components/RunDetailsContent.test.ts` (traceId/sessionId forwarded; caption), `FileObservabilityModule.test.ts` (sessionId union), `fetchSpansForRun.test.ts`, `describeTraceCorrelation`.
- **Integration** — `tests/integration/server/traces-precise-first.integration.test.ts`: real `/v1/traces` ingest → file store → `/api/traces`; two overlapping runs of one service → one root; an eval-only exact result still reaches the window; the fallback drops the other run's whole trace (3 spans incl. an identity-less child) and keeps the untagged + third-party-id traces; pagination stays in phase.
- **E2E** — `tests/e2e/traces-precise-first-correlation.spec.ts`: real OTLP ingest + seeded reports, opens `/runs/:id` → Traces. **Red on the previous build (9 spans / 3 roots), green now (3 / 1 + caption).**

### Not changed

`metricsService` on `main` correlates by A/B/D only (no window clause), so no run-id post-filter is needed there. The open Strategy-C-for-metrics work (midpoint attribution) should apply the same per-trace identity check when it lands — `filterWindowSpans` in `traceCorrelation.ts` is reusable.

### Review notes

An adversarial second-opinion review of the first cut flagged two real issues, both fixed in the last commit: (1) the direct-hit gate accepted an eval-span-only result and would have suppressed the window for Strategy-C-only agents; (2) using `gen_ai.conversation.id` / `session.id` mismatches as *negative* evidence would have blanked the tab for agents that legitimately fill those OTEL attributes with their own ids. Trade-off accepted: an agent that relies on the window and stamps only `gen_ai.conversation.id` gets no neighbour-filtering in the (rare) fallback case — stamping `agent_health.run.id = AGENT_EVAL_RUN_ID` (or propagating `TRACEPARENT`) is the documented way to opt in.

### Issues Resolved

Traces tab renders neighbouring runs' spans at concurrency > 1 (reported by the team; no issue filed).

### Check List
- [x] New functionality includes testing (unit + integration + e2e).
- [x] New functionality has been documented (AGENTS.md, docs/CONNECTORS.md).
- [x] CHANGELOG.md updated under `## [Unreleased]`.


goyamegh:
### Verification proof (text only)

**Gates** (worktree on `origin/main` + this branch, head `1ff4fdf43`):
```
npm run build:all                      → exit 0
npm run test:unit                      → Test Suites: 383 passed, 383 total / Tests: 6706 passed, 5 skipped
npm audit --audit-level=high           → exit 0 (no high/critical)
DCO                                    → 4/4 commits signed off
jest tests/integration/server/traces-precise-first.integration.test.ts
                                       → 9 passed
jest tests/integration/server/otlp-file-traces.integration.test.ts
                                       → 4 passed
playwright tests/e2e/traces-precise-first-correlation.spec.ts + 10 existing trace/comparison/run-report specs
                                       → 46 passed
```

**Regression test is red on the previous build.** Same e2e spec, server built from `origin/main` (before this PR):
```
✘ a run that correlates by trace id shows exactly ONE root and none of the neighbouring run's spans
    Expected length: 3
    Received length: 9      ← 3 runs' trees (3 roots) unioned into one report's Traces tab
✘ a Strategy-C-only run falls back to the window, drops other runs' traces and says so
  2 failed
```
Rebuilt with this branch:
```
✓ a run that correlates by trace id shows exactly ONE root and none of the neighbouring run's spans (2.8s)
✓ a Strategy-C-only run falls back to the window, drops other runs' traces and says so (3.7s)
  2 passed
```

**Read-only API check against a shared observability cluster** (the reporting team's case — a run from a concurrency-5 benchmark; ids redacted). Window = `[timestamp − durationMs − 60s, timestamp + 60s]` exactly as the Traces tab derives it:

| server | body | spans | traces | roots | `correlation` |
|---|---|---|---|---|---|
| previous build | `{ runIds, agents }` (what the tab sent) | 72 | 3 | 3 — roots carry 3 different run ids | — |
| this branch | `{ runIds, agents }` | 19 | 1 | 1 | `{ strategy: 'runIds', windowFiltered: 0 }` |
| this branch | `{ traceId, runIds, agents }` (what the tab sends now) | 25 | 1 | 1 | `{ strategy: 'traceId', windowFiltered: 0 }` |
| this branch | `{ agents }` only (no exact id at all) | 72 | 3 | 3 | `{ strategy: 'window', windowFiltered: 0 }` — nothing to filter against, as designed |

Rendered Traces tab on this branch for that report: one root span, badge `25`, caption **"Matched by trace id"** (previous build: badge `72`, three root rows).

